PR #132, opened 2021-01-28
Slightly reduce TLS file security checks

will:
* Allow files to additionally be owned by root
* Allow group read permissions on private key file

I don't think these changes make the certs all that much less secure, and they would allow me to reuse the certs the postgres server is using without having to copy them and change the owner away from root and the permissions.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,3 +2,4 @@ pgagroal was created by the following authors:
 
 Jesper Pedersen <jesper.pedersen@redhat.com>
 David Fetter <david@fetter.org>
+Will Leinweber <will@bitfission.com>
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -45,9 +45,9 @@ See a [sample](./etc/pgagroal/pgagroal.conf) configuration for running `pgagroal
 | failover | `off` | Bool | No | Enable failover support |
 | failover_script | | String | No | The failover script to execute |
 | tls | `off` | Bool | No | Enable Transport Layer Security (TLS) |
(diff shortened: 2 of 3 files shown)


jesperpedersen:
Hi Will,

Maybe we should divide the checks into user and root - just within an overall `if` probably - and document the checks more. What do you think ?

Thanks for contributing !

will:
Hi, I've updated this with more documentation. I wasn't able to figure out a way to do the checks differently. It's not that I'm running pgagroal as root, just that the files may or may not be owned by root, so I cant see a good place to do an overall `if`. I could just be missing it though.

jesperpedersen:
I'm thinking about expanding the checks, like

https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/security.c#L3187

into

```
if (st.st_uid == geteuid())
{
  /* Check for 0600 and log */
}
else if (st.st_uid == 0)
{
  /* Check for 0640 and log */
}
else
{
  /* Log requirements */
  goto error;
}
```
or similar checks within that function such as that the logging is more specific to each case.

Please, squash your commits and remember to update the `AUTHORS` file as well :)

will:
@jesperpedersen sorry for the delay, the last week for me was very busy. I think I've done what you were looking for, let me know. It's been rebased on top of the current master. 

Also it seems like my editor did a bunch of whitespace trimming and I committed them before noticing. It looks like it was just on a few things, so I think maybe you would want them in. If not, I can work through and drop those from the patch.

jesperpedersen:
Thanks Will !

Looking forward to your next contribution :)

Issue #159, opened 2021-07-12
Add more pool metrics


An-DJ:
**Is your feature request related to a problem? Please describe.**
There is a lack of metrics to characterize the pool state.

**Describe the solution you'd like**

- [X] Add `pgagroal_query_count` which characterizes the number of queries in the most recent time interval.
- [X] Add `pgagroal_tx_count` which characterizes the number of transactions in the most recent time interval.
- [X] Add `pgagroal_wait_time` which characterizes the previous client waiting time.

jesperpedersen:
Lets start with a simple guard of `if (config->metrics > 0)`

PR #161, opened 2021-07-25
[#159] Add new pool metric pgagroal_query_count and `Query rate` panel

An-DJ:
# Issue
#159 

# Task
* Add `pgagroal_query_count` which characterizes the number of queries.
* Add new panel `Query rate` in Grafana dashboard,

### DIFF
--- a/src/libpgagroal/pipeline_session.c
+++ b/src/libpgagroal/pipeline_session.c
@@ -220,6 +220,11 @@ session_client(struct ev_loop *loop, struct ev_io *watcher, int revents)
    {
       if (likely(msg->kind != 'X'))
       {
+         if (msg->kind == 'Q' || msg->kind == 'E')
+         {
+            pgagroal_prometheus_query_count_add();
+         }
+
          status = pgagroal_write_socket_message(wi->server_fd, msg);
          if (unlikely(status == MESSAGE_STATUS_ERROR))
          {
--- a/src/libpgagroal/pipeline_transaction.c
+++ b/src/libpgagroal/pipeline_transaction.c
@@ -236,6 +236,11 @@ transaction_client(struct ev_loop* loop, struct ev_io* watcher, int revents)
    {
       if (likely(msg->kind != 'X'))
       {
+         if (msg->kind == 'Q' || msg->kind == 'E')
+         {
+            pgagroal_prometheus_query_count_add();
+         }
+
          if (config->track_prepared_statements)
          {
             int offset = 0;
--- a/contrib/grafana/dashboard.json
+++ b/contrib/grafana/dashboard.json
@@ -577,11 +577,89 @@
     },
     {
       "datasource": null,
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "barAlignment": 0,
+            "drawStyle": "line",
+            "fillOpacity": 0,
+            "gradientMode": "none",
+            "hideFrom": {
+              "legend": false,
+              "tooltip": false,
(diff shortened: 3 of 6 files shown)


jesperpedersen:
This would count all client requests as a query - it would be better to just count the actual SQL queries.

See https://www.postgresql.org/docs/devel/protocol-message-formats.html for the message protocol - you will need support for the simple query and the prepared query format. You can use https://github.com/jesperpedersen/pgprtdbg for protocol debugging.

You can likely reuse code from `pipeline_transaction.c` to track message boundaries, but we need guard against this extra overhead -  maybe a `statistics` configuration variable with a default of `on`.

Only `pipeline_session.c` and `pipeline_transaction.c` needs to support this -- `pipeline_perf.c` won't support extended statistics.

An-DJ:
> This would count all client requests as a query - it would be better to just count the actual SQL queries.
> 
> See https://www.postgresql.org/docs/devel/protocol-message-formats.html for the message protocol - you will need support for the simple query and the prepared query format. You can use https://github.com/jesperpedersen/pgprtdbg for protocol debugging.
> 
> You can likely reuse code from `pipeline_transaction.c` to track message boundaries, but we need guard against this extra overhead - maybe a `statistics` configuration variable with a default of `on`.
> 
> Only `pipeline_session.c` and `pipeline_transaction.c` needs to support this -- `pipeline_perf.c` won't support extended statistics.

T

An-DJ:
@jesperpedersen I have check the message kind in `Simple Query` and `Extended Query(prepare-statement)`. 
* For the SimpleQuery, a query can be created by a `Q` message.
* For the PrepareQuery, a prepare query can be created by a `P` message which is used to parse the unsettled sql. Then a execute query can be created by a `E` message which is used to execute the settled sql.

So the query_count should be increased by 'Q' 'P' and 'E'. The newer commit includes that implementation.

src/libpgagroal/pipeline_perf.c
```diff
@@ -97,6 +97,7 @@ performance_client(struct ev_loop *loop, struct ev_io *watcher, int revents)
    wi = (struct worker_io*)watcher;
 
    status = pgagroal_read_socket_message(wi->client_fd, &msg);
+
```
jesperpedersen, line 100: White space

src/libpgagroal/pipeline_session.c
```diff
@@ -220,6 +220,11 @@ session_client(struct ev_loop *loop, struct ev_io *watcher, int revents)
    {
       if (likely(msg->kind != 'X'))
       {
+         if (msg->kind == 'Q' || msg->kind == 'P' || msg->kind == 'E')
```
jesperpedersen, line 223: `P` is Parse which we shouldn't count as a query, as `B` (Bind) won't need it after 5 statements.

Lets start with `Q` and `E` in this category.

src/libpgagroal/pipeline_transaction.c
```diff
@@ -236,6 +236,11 @@ transaction_client(struct ev_loop* loop, struct ev_io* watcher, int revents)
    {
       if (likely(msg->kind != 'X'))
       {
+         if (msg->kind == 'Q' || msg->kind == 'P' || msg->kind == 'E')
```
jesperpedersen, line 239: Same here.

jesperpedersen (changes requested):

jesperpedersen:
You are missing the reset part for the new metric

An-DJ:
> You are missing the reset part for the new metric

Done. I found that a 'P' message(parse) can be sent by `prepare` statement, and 'B','E'... messages can be sent by  `execute` statement. Maybe the parsing phase should not be considered as query. I have removed it just now.

jesperpedersen:

PR #162, opened 2021-08-05
[#159] Add new pool metric pgagroal_tx_count and `Transaction rate` panel

An-DJ:
# Issue
#159 

# Task
* Add `pgagroal_tx_count ` which characterizes the number of transactions.
* Add new panel `Transaction rate` in Grafana dashboard,

### DIFF
--- a/contrib/grafana/dashboard.json
+++ b/contrib/grafana/dashboard.json
@@ -655,6 +655,85 @@
     },
     {
       "datasource": null,
+      "description": "",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisLabel": "",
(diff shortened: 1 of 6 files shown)


jesperpedersen:
Looks good, but you have to only count the `I` to `T` transition - as the `T` to `E` transition is the same transaction

jesperpedersen:
Looks good, but you have to only count the `I` to `T` transition - as the `T` to `E` transition is the same transaction

An-DJ:
> Looks good, but you have to only count the `I` to `T` transition - as the `T` to `E` transition is the same transaction

@jesperpedersen I have checked my code and found that it only count the `I` to `T` transaction. : )

Perhaps you misunderstood something?

jesperpedersen:
Yeah, I read it wrong, sorry.

PR #166, opened 2021-08-11
Change `Query rate` and `Transaction rate` panels' unit

An-DJ:
Change `%` to `rps`


jesperpedersen:

PR #167, opened 2021-08-15
Add `connections per database` panel

An-DJ:
Add `connections per database` panel

We do not need to add metrics to complete the statistics on the number of connections per database.


jesperpedersen:

PR #168, opened 2021-08-15
Add new metric pgagroal_pipeline_mode

An-DJ:
The metric `pgagroal_pipeline_mode` can show which pipeline is used now.

### DIFF
--- a/contrib/grafana/dashboard.json
+++ b/contrib/grafana/dashboard.json
@@ -32,7 +32,6 @@
     },
     {
       "datasource": null,
-      "description": "pgagroal state",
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Maybe do a panel between "Active connections" and "Session time" that maps the value to its name - `0` -> "Performance" and so on

An-DJ:
> Maybe do a panel between "Active connections" and "Session time" that maps the value to its name - `0` -> "Performance" and so on

Done.

jesperpedersen:
Nice !

PR #169, opened 2021-08-17
Add new metric pgagroal_connection_query_count and new panel "Query rate per database"

An-DJ:
The metric `pgagroal_connection_query_count` shows the number of queries per connection.

The new panel "Query rate per database" shows query rate per db based on `pgagroal_connection_query_count`.

### DIFF
--- a/contrib/grafana/dashboard.json
+++ b/contrib/grafana/dashboard.json
@@ -719,7 +719,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "pgagroal_query_count - (pgagroal_query_count offset 1s)",
+          "expr": "irate(pgagroal_query_count[1m])",
           "interval": "",
           "legendFormat": "",
           "refId": "query rate"
@@ -800,7 +800,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "pgagroal_tx_count - (pgagroal_tx_count offset 1s)",
+          "expr": "irate(pgagroal_tx_count[1m])",
           "interval": "",
           "legendFormat": "",
           "refId": "A"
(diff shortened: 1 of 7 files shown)


jesperpedersen:
Hmm, I don't see any updates to the panel during a pgbench run

An-DJ:
> Hmm, I don't see any updates to the panel during a pgbench run

@jesperpedersen 
Well, I start the pgagroal in session mode, do several simple queries like "select * from t", and there is a clear reaction shown on the dashboard:
image

Maybe the interval is unsuitable?
image

jesperpedersen:
1 second resolution should be ok.

Could you check
```
pgbench -i -s 100 -p 2345 -U <username>
pgbench -M prepared -T 60 -c 4 -j 4 -p 2345 -U <username>
```
on your end ?

An-DJ:
> 1 second resolution should be ok.
> 
> Could you check
> 
> ```
> pgbench -i -s 100 -p 2345 -U <username>
> pgbench -M prepared -T 60 -c 4 -j 4 -p 2345 -U <username>
> ```
> 
> on your end ?

It's my mistake... The promql has been wrong and I have fixed it just now.

Besides, I noticed that some messages may be missed in query statistics. I have fixed it, too.

@jesperpedersen PTAL : )

An-DJ:
I have totally checked the code by pgbench. Actually the prometheus builtin function `irate` is more suitable for the QPS case. `irate` can accurately compute the instantaneous rate for rapidly changing metrics, and  it's not negative. 

We can run the following test, and get normal charts in `Query rate`, `Transaction rate` and `Query rate per database` panels:

```
pgbench -i -s 100 -p 2345 -U <username>
pgbench -M prepared -T 60 -c 4 -j 4 -p 2345 -U <username>
```
image

If rapid change is not recorded by prometheus, we set `scrape_interval = 1s` in `prometheus.yml`. (e.g. I noticed that `pgbench -i -s 100 -p 2345 -U <username>` may cause a instantaneous increase in the number of queries, but a larger interval will cause irate to fail to recognize this change.) `irate` is based on the last two data points of metrics.

The promqls in `Query rate`, `Transaction rate` and `Query rate per database` panels have been fixed just now.

@jesperpedersen PTAL

jesperpedersen:
Yeah, this is much better :) However, it seems like "Query rate per database" has a defined value after the benchmark is done - for a while (could be the `1m`) - can see it in your graphs as well. I tried to switch the scraping rate, but it is still the same issue.

We should switch "Send traffic" and "Receive traffic" to `irate` as well.

jesperpedersen:
Look at "Query rate per database" -- your line stays at ~8.5k/sec after the benchmark has completed - as per "Query rate" and "Transaction rate". The latter two goes to 0.

An-DJ:
> Look at "Query rate per database" -- your line stays at ~8.5k/sec after the benchmark has completed - as per "Query rate" and "Transaction rate". The latter two goes to 0.

I have run the benchmark `pgbench -i -s 100 -p 2345 -U <username>` and get panels as follows:

image

It seems normal...

An-DJ:
@jesperpedersen I have changed the `irate` interval into "[5s]" to solve the long-tail problem. PTAL

jesperpedersen:

PR #200, opened 2022-02-08
Admin master key wrong message

fluca1978:
While using `pgagroal-admin` I found some odd cases that could be improved:
- warn the user when the inserted password is not valid, otherwise the user will see `Master Key` on the screen without any hint;
- print out where the vault is stored
- catch SIGINT during the insertion of the master key.

Since I'm not a day-by-day C developer, please review carefully.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -55,12 +55,6 @@
 #define ACTION_REMOVE_USER 4
 #define ACTION_LIST_USERS  5
 
-static char CHARS[] = {'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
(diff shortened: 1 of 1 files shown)


fluca1978:
I've done a few changes. In the beginning I deleted the `is_valid_password`, but after that I realized that having a function that do checks, not only about length, but about quality, would be useful. Therefore, I split `CHARS` into its part and created `do_check_master_key` to check the quality of the master key password.
Then, I refactored the generating function to use the concatenation of all the parts as the old `CHARS`.
The only part that remains, if we want, is to force the generation to validate the password, or better, to extract at least one char from every piece.

In the cae you want to integrate, squash or cherry pick.

fluca1978:
Ok, added also the capability to automatically generate a password picking different set of chars, so that it will always contain numbers, letters, specials, and so on.

jesperpedersen:
An earlier commit remove all password strength checking for user based account - leaving it to the DBA to set the guidelines. So, I think we should do the same in this case. And, hence `CHARS` is only going to be used for generated keys.

jesperpedersen:
Yeah, the cherry-pick looks like it, but I would like to have a proper pull request for it, so rebase, fix indentation and squash, please :)

fluca1978:
Opened issue #206. I'm closing this one and opening a new pull request since most of the work done here is not accepted.

PR #207, opened 2022-02-21
[#206] Remove is_valid_password function and provide better messages.

### LINKED ISSUE #206: Better messages when entering master password
In `1.4` when the user enters a master password thru `pgagroal-admin`, in the case the password is incorrect (i.e., wrong size) the program keeps asking the user for a password without producing any error message that can help the user to understand what she is doing wrong.

Related to #200

fluca1978:
Close #206.

See <https://github.com/agroal/pgagroal/pull/200#issuecomment-1032660243>.
The is_valid_password() was checking only the password length and the fact
that was made by ASCII chars.
The check for the length can be done "inline" directly within a loop.
Added a constant with the minimal length of the password, so that it is possible
to insert a warning message for the user in the case she inputs a too short password.

The system also prompts the user for a password with a message that explicitly tells
her the password will not appear on the terminal.

See also the initial work on #200.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -45,6 +45,7 @@
 #include <sys/stat.h>
 
 #define DEFAULT_PASSWORD_LENGTH 64
+#define MIN_PASSWORD_LENGTH 8
 
 #define ACTION_UNKNOWN     0
 #define ACTION_MASTER_KEY  1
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Just needs a bit of indentation in the `while` and `if`.

fluca1978:
Indentation fixed, I need a way to instrument Emacs to understand the correct indentation for this project or I'm going crazy.

Please also note that this does not take care of 0455132a3d87 that, in my opinion, is useful.

jesperpedersen:
Squashed and merged.

Yes, feel free to create another pull request with 0455132a3d87affe9613e685cf323c42afd9487d

PR #211, opened 2022-02-24
[#210] Understand all 'debug' levels for log_level.

### LINKED ISSUE #210: log_level = debug is not undrstood
In the configuration file, the `log_level` can be set to `debug1` but not to `debug`. While this is documented, I would expect and prefer to set `debug` instead of `debug1` that sounds more as a development setting. Also, in the case of `log_level = debug` the level is set to fallback `info` without any warning.

fluca1978:
This makes `log_level` configuration parameter understand
any "debug*" string, like "debug10", as well as "debug".
In particular, if `log_level = debug` (without any number)
the level will be set to `debug1`.
On the other hand, if `log_level = debug99` the
level will be adjusted to `debug5`.

Close #210

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -28,7 +28,7 @@ See a [sample](./etc/pgagroal/pgagroal.conf) configuration for running `pgagroal
 | metrics | 0 | Int | No | The metrics port (disable = 0) |
 | management | 0 | Int | No | The remote management port (disable = 0) |
 | log_type | console | String | No | The logging type (console, file, syslog) |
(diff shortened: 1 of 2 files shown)


jesperpedersen:
The indentation monster is back :(

Do  you have

```
(setq tab-width 3)
(setq-default indent-tabs-mode nil) 
```

?

fluca1978:
> The indentation monster is back :(
> 
> Do you have
> 
> ```
> (setq tab-width 3)
> (setq-default indent-tabs-mode nil) 
> ```
> 
> ?

Yes, but apparently the `3` is not the value you want. Is it fine now?

jesperpedersen:
Squashed, rebased, and merged.

PR #212, opened 2022-02-24
Return value coherent to logging type macros.

fluca1978:
The as_logging_type() function must return
the default logging type PGAGROAL_LOGGING_TYPE_CONSOLE
that has the value 0.
This does not change the behaviour of the application
but makes the code coherent with other
configuration parsing functions.

I don't believe there is the need for an issue for such a commit, since it's just syntax sugar.


jesperpedersen:
Merged.

Issue #45, opened 2020-02-04
Log rotation


jesperpedersen:
Implement log rotation for files - could be based on a date or size.

Will require `log_path` to understand date modifiers like `%Y`, `%m`, `%d`, `%H`, `%M` and `%S`.

dmig-alarstudios:
use `logrotate` for that. it requires just reopening the log file on SIGHUP.

Issue #44, opened 2020-02-04
Log formatting


jesperpedersen:
Implement log formatting codes that can style the logging output. Could borrow similar codes as PostgreSQL.

May require changes to zf_log.

PR #213, opened 2022-02-24
Log rotation and formatting

fluca1978:
This is a possible not-so-efficient implementation of log rotation and formatting (see #45  and #44).
The idea is that `log_path` supports `strftime(3)`, so that rotation can be enabled. Also added `log_rotation_age` (seconds) and `log_rotation_size` (bytes) and `log_line_prefix` (string).
Example of final configuration:
```
log_mode  = create
log_rotation_size = 2M
log_rotation_age = 1m
log_line_prefix  = PGAGROAL-%Y-%m-%d-%H:%M:%S'
```

There are now several `log_rotation_xxx` functions that help keeping the code cleaner. If any of `log_rotation_age` or `log_rotation_size` parameters are enabled, and the logging is to a file, the rotation is enabled, otherwise is disabled.
Whenever a new entry is loggeg, the application looks for the need to rotate.
This is not efficient, since under heavy load the system could spend too much time in trying to understand if rotation is required.
If a rotation is required, the log file is closed and a new one is opened.
The `log_rotation_age` is used in conjunction with a global variable that keeps track of when the next rotation will happen.
Since rotation is evaluated only when a new log entry is issued, rotation will not happen automatically until the next log entry arrives. This means that is an almost-rotation.

Drawbacks:
- rotation happens when a new log entry arrives. Probably we should improve this by means of a kind of "try every n entries", like inspecting for rotation only after 10 entries are logged to improve performances.
-  `log_line_prefix` does not allow spaces, since the `extract_key_value` does not handle spaces too.

We should also evaluate to store global variables `next_log_rotation_age` and `current_log_path` within the configuration structure, because they could be useful also outside of the `logging.c` file (I'm not sure).

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -30,7 +30,9 @@ See a [sample](./etc/pgagroal/pgagroal.conf) configuration for running `pgagroal
 | log_type | console | String | No | The logging type (console, file, syslog) |
 | log_level | info | String | No | The logging level (fatal, error, warn, info, debug1, ..., debug5) |
 | log_path | pgagroal.log | String | No | The log file location |
-| log_mode | append | String | No | Append to or create the log file (append, create) |
+| log_rotation_age | -1 | String | No | The age that will trigger a log file rotation. If expressed as a positive number, is managed as minutes. Supports suffixes: 'H' (hours), 'D' (days), 'W' (weeks), 'M' (months) and 'Y' (years) |
+| log_rotation_size | -1 | String | No | The size of the log file that will trigger a log rotation. Supports suffixes: 'B' (bytes), the default if omitted, 'K' (kilobytes), 'M' (megabytes), 'G' (gigabytes). |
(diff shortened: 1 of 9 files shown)


jesperpedersen:
I think most people will use `info` as their level, which means very little logging will happen.

So, having the overhead of each log entry is ok - debugging with `debug5` will be very specific so minimal log size there as well. I don't think we need to support spaces in `log_line_prefix`, and likewise I don't think we need to account for gaps in the file naming.

All in all looks good - just need to whack the indentation monster

jesperpedersen:
BTW, you could move the documentation to the .h file ? And, rebase on master ?

fluca1978:
Hope the code is clean now.

jesperpedersen:
You could do a `git log` to get the ids of your patches. Then `git format-patch` all of them. Then

```
git fetch upstream
git branch -m old_logging_rotation
git checkout master
git rebase -i upstream/master
git checkout -b logging_rotation master
```

and then start with `git apply`s. At the end `git push -f origin logging_rotation`.

fluca1978:
See #216

PR #216, opened 2022-02-28
[#45][#44] Log rotation and formatting support.

### LINKED ISSUE #44: Log formatting

### LINKED ISSUE #45: Log rotation

fluca1978:
See initial work in #213.
This patch implements the log rotation and formatting support.
There are now three new configuration options:
- log_line_prefix is a strftime(3) compatible string that is used
  as a prefix whenever a new line is logged to console or file;
- log_rotation_age provides the age (in seconds) of when the log must
  be rotated. There is raw support for human strings like '1d' for
  'one day';
- log_rotation_size provides the size in bytes when the log must be
  rotated.

Log rotation is evaluated at every new log line issuing, that means
the age and the size are not honored effectively until a new log line
is issued. When the system decides to rotate the log, the log file
is flushed and a new one is opened. To allow rotation, log_path
supports a strftime(3) compatible string like
so that whnever a new log file must to be created, it can be
placed side by side the closed one.
In the case the new log file already exists, the log_mode is
used to decide if the log must be truncated or used in append mode.

In the case the log_mode is set to console or to syslog, the
log rotation is automatically disabled, therefore the overhead of
the rotation machinery will be applied only when logging to a file.

Example of configuration:

log_type  = file
log_level = info
log_mode  = create
log_rotation_size = 2M
log_rotation_age = 1m
log_line_prefix  = PGAGROAL-%Y-%m-%d-%H:%M:%S'

Close #44
Close #45

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -29,7 +29,10 @@ See a [sample](./etc/pgagroal/pgagroal.conf) configuration for running `pgagroal
 | management | 0 | Int | No | The remote management port (disable = 0) |
 | log_type | console | String | No | The logging type (console, file, syslog) |
 | log_level | info | String | No | The logging level (fatal, error, warn, info, debug, debug1 thru debug5). Debug level greater than 5 will be set to `debug5`. Not recognized values will make the log_level be `info` |
-| log_path | pgagroal.log | String | No | The log file location |
+| log_path | pgagroal.log | String | No | The log file location. Can be a strftime(3) compatible string. |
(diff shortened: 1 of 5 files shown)


jesperpedersen:
I did some minor changed to formatting, and API docs. Merged.

Issue #217, opened 2022-04-03
Return 400 Bad Request for any method except 'GET'


nikitabugrovsky:
**Is your feature request related to a problem? Please describe.**

At the moment if we send request to pgagroal `/metrics` endpoint using any method except 'GET',  we receive 403 Forbidden. 403 indicates that the server understands the request but refuses to authorize. That is not exactly the case with the pgagroal exporter.

**Describe the solution you'd like**

Introduce new check in the code that will allow to return 400 Bad Request for any method except 'GET'.

Ref: [Hypertext Transfer Protocol -- HTTP/1.1](https://www.w3.org/Protocols/rfc2616/rfc2616.html)

PR #218, opened 2022-04-03
[#217] Return 400 Bad Request for any method except 'GET'

nikitabugrovsky:


jesperpedersen:
Merged.

PR #225, opened 2022-05-04
[#224] Prefill a new server when doing a `switch-to`

### LINKED ISSUE #224: prefill after switch-to
Don't know if this is a valuable feature, but with `1.5` when I do a `switch-to` and/or a `reset-server` the prefill seems to be dropped.
I mean, when I start `pgagroal` the "initial" connections from `pgagroal_databases.conf` are established. When I do a `switch-to` to another server, the new server does not get the "initial" connection number but only the "minimum".
The same when I come back with another `switch-to` the original server.
Is there a reason or should `pgagroal` honor the "initial" connection every time a server is changed as primary?

fluca1978:
When a `switch-to` operation is issued, the old server is flushed
(i.e., connections are closed) and new connections to the new primary
server are established.
When such connections are established, the system does a full prefill
against the new primary, so that when the `switch-to` completes, the
new primary is prefilled as the old one was.

Adds also a better detail about the server that is going to be
flushed.

Close #224

### DIFF
--- a/src/include/pool.h
+++ b/src/include/pool.h
@@ -94,9 +94,11 @@ pgagroal_flush(int mode, char* database);
 /**
  * Flush the pool for a specific server
  * @param server The server
+ * @param prefillNextServer true if the next selected server must be prefilled.
+ *                          (.e.g, when doing a switch-to)   
  */
 void
-pgagroal_flush_server(signed char server);
+pgagroal_flush_server(signed char server, bool prefillNextServer);
 
 /**
  * Prefill the pool
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -193,7 +193,7 @@ start:
 
             if (!fork())
             {
-               pgagroal_flush_server(server);
+               pgagroal_flush_server(server, false);
             }
 
             if (config->failover)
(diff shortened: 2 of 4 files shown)


jesperpedersen:
Looking at it - there are 2 places where `pgagroal_flush_server` is called from - 1) The failover scenario, and 2) An unresponsive primary.

For 1) we know that a new primary has been selected. For 2) we know that the primary is the same.

So, I think we can just change `pgagroal_flush_server` to check for the primary (`pgagroal_get_primary`) and if it is different than the value passed in invoke `pgagroal_prefill(true)`. That way we don't have to change the function signatures, and it should be a 3-4 line patch :)

Thoughts ?

fluca1978:
I've crated #226 as a possible implementation of your suggestion.
One thing I don't understand is why `pgagroal_flush_server` accepts a `unsigned char` as a server index, while other parts like `pgagroal_get_primary` handles an integer.
I've inserted a check in the extraction of the new primary, that could be avoided since if `pgagroal_get_primary` fails we don't need to prefill in any case, but the code looks stronger.

jesperpedersen:
The `pgagroal_get_server` uses an integer as its out parameter to highlight that the value can be negative if the function fails to find a primary.

fluca1978:
Got it, thanks.
I'm closing this same PR since #226 looks more what you have in mind.

PR #226, opened 2022-05-04
[#224] Prefill when doing a `switch-to` or primary change.

fluca1978:
Whenever the primary host is changed, by means of an explicit
`switch-to` command or by a primary failure, the connection flushing
is activated. If possible, the prefill should be also restored on the
new server.
As suggested in
<https://github.com/agroal/pgagroal/pull/225#issuecomment-1117218107>
it would be nice to check if the specified new server is the same as
the old one (failure of the primary) or a different one (`switch-to`),
and in the case they are different the prefill is forced.

### DIFF
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -802,13 +802,14 @@ void
 pgagroal_flush_server(signed char server)
 {
    struct configuration* config;
+   int primary = -1;
 
    pgagroal_start_logging();
    pgagroal_memory_init();
 
    config = (struct configuration*)shmem;
 
-   pgagroal_log_debug("pgagroal_flush_server");
+   pgagroal_log_debug("pgagroal_flush_server %s", config->servers[server].name);
    for (int i = 0; i < config->max_connections; i++)
    {
       if (config->connections[i].server == server)
@@ -845,11 +846,16 @@ pgagroal_flush_server(signed char server)
       }
    }
 
-   if (config->number_of_users > 0 && config->number_of_limits > 0)
+   if (pgagroal_get_primary(&primary))
+      pgagroal_log_debug("No primary defined");
+   else
    {
-      if (!fork())
+      if (config->number_of_users > 0 && config->number_of_limits > 0 && server != (unsigned char)primary && primary != -1)
       {
-         pgagroal_prefill(false);
+         if (!fork())
+         {
+            pgagroal_prefill(true);
+         }
       }
    }
 


jesperpedersen:
I think I like this direction a bit more.

Called the variable `primary` - and set it to `-1`, since we don't know what `server` really is. And, do the `server != (unsigned char)primary` check before the other prefill checks are done.

jesperpedersen:
I would out in an explicit `primary != -1` check as well...

jesperpedersen:
You don't need to enter the https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/pool.c#L848 statement if primary isn't valid, so

```
int primary = -1;

if (pgagroal_get_primary(&primary))
{
    pgagroal_log_debug("No primary defined");
}

if (primary != -1 && server != (unsigned char)primary)
{
  // Line 848 and onwards ...

}

```

jesperpedersen:
Can you squash and force push ?

jesperpedersen:
Ah, there is still a thing.

We don't need to call prefill(false) for an old flushed primary server... which is the "if (primary != -1 && ..." check.

Otherwise it would try and prefill to `MIN_SIZE`

fluca1978:
Just to make it clear, you don't want to prefill _at all_ if the primary is not set or the server has remained the same, right?
That makes me wondering: when it is fine to prefill to the MIN_VALUE?

fluca1978:
Just to make it clear, you don't want to prefill _at all_ if the primary is not set or the server has remained the same, right?
That makes me wondering: when it is fine to prefill to the MIN_VALUE?

jesperpedersen:
This is only for the `flush_server` case, and yes - there should be no prefill (as `MIN_SIZE` could be > 0, and therefore cause connects to an invalid host).

There are a lot of cases where `prefill(false)` makes sense - f.ex. after validation of a connection fails or after idle timeout. This is why people need to consider `MIN_SIZE` carefully as it will tie up resources on the database cluster.

fluca1978:
I'm not really comfortable with squashing a set of published commits, please check if it is fine now.

src/libpgagroal/pool.c
```diff
@@ -845,11 +847,21 @@ pgagroal_flush_server(signed char server)
       }
    }
 
-   if (config->number_of_users > 0 && config->number_of_limits > 0)
+   if (pgagroal_get_primary(&primary))
+      pgagroal_log_debug("No primary defined");
+   else
    {
-      if (!fork())
+      prefill = server != (unsigned char)primary && primary != -1;
```
jesperpedersen, line 854: This needs to go into the L856 if check - and therefore you can remove  the `prefill` variable

jesperpedersen (changes requested):

jesperpedersen:
And, you can just do 2 levels

```
if (primary != -1 && ...
{
   // Existing code
  if (..
```

fluca1978:
Sorry, I'm a little confused: if there is the need to prefill (number of users and limits not zero) we `fork` and issue a `pgagroal_prefill` in the child process, that then exits, while the parent process continues to memory destroy and exits too.
So what is the need to `fork` if not issuing a `pgagroal_prefill`?

jesperpedersen:
I have to check something with uncrustify - otherwise I'll merge :)

jesperpedersen:
Yeah, it could be a `static` function within `pool.c`

jesperpedersen:
Merged.

Issue #223, opened 2022-04-28
pgagroal-cli reload and added server


fluca1978:
Running `pgagroal` version 1.5.0.
TL;DR: if I add a new not-primary server to the configuration, issuing a `reload` makes the server appearing in the list of available (and I can `switch-to`). The old primary is then reported as `Not init` which is not the correct status.

Initial situation, one server with prefill:
```
[pgagroal]
host = *
port = 5432
unix_socket_dir = /tmp

log_type  = console
log_level = info
max_connections = 900
pipeline = performance

[venkman]
host = venkman
port = 5432
primary = on
```

and the prefill:

```
% cat /etc/pgagroal/pgagroal_databases.conf
# DATABASE USER MAX INITIAL MIN
 pgbench pgbench 600 200  5
```

When I start `pgagroal` the prefill works, and the connections are established:

```
% pgagroal-cli details | less
Status:              Running
Active connections:  0
Total connections:   200
Max connections:     900
---------------------
Server:              venkman
Host:                venkman
Port:                5432
State:               Primary
---------------------
Database:            pgbench
Username:            pgbench
Active connections:  0
Max connections:     600
Initial connections: 200
Min connections:     5
```

So far so good.
Now adding a new server in the configuration:

```
% cat /etc/pgagroal/pgagroal.conf
...
[venkman]
host = venkman
port = 5432
primary = on

[spengler]
host = spengler
port = 5432
primary = off
```

if I now issue a `reload` the `details` command reports that even the primary is in state not initialized:

```
% pgagroal-cli reload
% pgagroal-cli details | less
Status:              Running
Active connections:  0
Total connections:   200
Max connections:     900
---------------------
Server:              venkman
Host:                venkman
Port:                5432
State:               Not init
---------------------
Server:              spengler
Host:                spengler
Port:                5432
State:               Not init
---------------------
Database:            pgbench
Username:            pgbench
Active connections:  0
Max connections:     600
Initial connections: 200
Min connections:     5
```

The server `venkman`, that holds the connections, is now `Not init` and this is not correct.
In the logs there is a line:

```
INFO  configuration.c:2738 Restart required for pidfile - Existing /tmp/pgagraol.5432.pid New
```

This is strange: the new server has been added to the list, so `reload` partially worked but the old primary is not in the correct status, so I guess if this is not a reloadable option the old configuration should be retained.
Am I missing something?

jesperpedersen:
[...]

PR #227, opened 2022-05-05
[#223] Keep server state during a `reload`

fluca1978:
When doing a `pgagroal-cli reload` the system reloads the
configuration and copies the server information via `copy_server`
function.
This commit makes the `copy_server` to keep the original server status
among a reload process.
This makes a "primary" as "not init (primary)" status, that is surely
better than "not init".

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -2698,7 +2698,7 @@ copy_server(struct server* dst, struct server* src)
    memcpy(&dst->name[0], &src->name[0], MISC_LENGTH);
    memcpy(&dst->host[0], &src->host[0], MISC_LENGTH);
    dst->port = src->port;
-   atomic_init(&dst->state, SERVER_NOTINIT);
+   atomic_init(&dst->state, src->state);
 }
 
 static void


jesperpedersen:
Should we tackle reordering of server definitions as well in this PR... What do you think ?

fluca1978:
Not sure what you mean with "reordering of server definition": would you like to have the primary at the fixed index zero?

Also, I'm suspecting that the `transfer_configuration` does not take into account all configuration settings, because for instance I'm not seeing log rotation ones. Is that a mistake or there is a specific reason?

fluca1978:
This is going to take more effort than I thought, because I found another strange behavior:

```
[venkman]
primary = on

[spengler]
primary = off
```

and the status is of course

```
venkman = Primary
spengler = Not init
```

than I switch the configuration and reload with `pgagroal-cli reload`
```
[venkman]
primary = off

[spengler]
primary = on
```
and the status reported for both servers is `Not init`, even if there is prefill!
My guess is that when the new configuration is initiated and then loaded from file, something prevents the new server to become the new primary. A brute force solution could be to force the primary status once `transfer_configuration` is done, but I would like to dig more to see if I get the point in the workflow that causes this double `Not init`.

jesperpedersen:
Yeah, you have to look if the existing primary is still defined (if there is one), and merge that info in.

`primary = on` is only a hint -- the real value won't be set until there is a connect to the server (https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/server.c#L152).

Of course, even you change the hint but the server is still primary it needs to stay that way.

fluca1978:
Fixed by #233 , apparently there is nothing more to do here.

PR #229, opened 2022-05-06
[#228] No warning messages about pid file when configuration reloads.

### LINKED ISSUE #228: warning about pid file during `reload` even if it is set to its default value
Running 08f1ca20fd86b46dd380244432c4d0f93d469953 (version 1.5.0).

When issuing a configuration reload with `pgagroal-cli reload` the system displays a log notice about the PID file:

```
 INFO  configuration.c:2743 Restart required for pidfile - Existing /tmp/pgagraol.5432.pid New 
```

The problem is that the pid file is not indicated at all in the configuration, so it will get its default value, that is not changed during the reload action. 
The problem seems to be that the function `restart_string` correctly checks even for null/empty strings, and in the new configuration (the one being reloaded) the `pidfile` is not initialized at all, so the message is triggered.

Either `restart_string` (and similar functions) should be made aware of which arguments can be not filled because they will be set later on, or the `config->pidfile` attribute should be set early on during `pgagroal_read_configuration` instead of when the attribute is going to be used (in `main.c` within `create_pidfile`).

fluca1978:
When configuration is reloaded via `pgagroal-cli reload`, the system
checks all the parameters and warns the user about the need of a
restart if some particular parameters have changed.
The check for the `pidfile` parameter is done via `restart_string()`
that, in the case the pidfile has not been specified in the
configuration file, finds a value in the old configuration (the one
running) and no value at all in the new one.
To avoid the misbehaviour, `pgagroal_validate_configuration()` now
invokes a new function `pgagroal_init_pidfile_if_needed()` that sets
the default value of the `pidfile` in the configuration structure.
This way, if the parameter has its default value, no difference will
be spot and no warning message will be issued.

It is safe to make the pidfile default configuration as the last step
in `pgagroal_validate_configuration()` because this function should be
called whenever the configuration is read, and after all this ensures
the configuration is in a "valid" state. On the other hand, it is not
safe to do such configuration within `pgagroal_read_configuration()`
because the default pidfile is built on top of `unix_socket_dir`, that
at that time could be not set.

Close #228

### DIFF
--- a/src/include/configuration.h
+++ b/src/include/configuration.h
@@ -171,6 +171,17 @@ pgagroal_validate_superuser_configuration(void* shmem);
 int
 pgagroal_reload_configuration(void);
 
+/**
+ * Automatically initialize the 'pidfile'
+ * if none has been specified.
+ * This function is called as last step
+ * from pgagroal_validate_configuration
+ * because it builds the pidfile on the value
+ * of unix_socket_dir.
+ */
+void
+pgagroal_init_pidfile_if_needed(struct configuration* config);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -1140,6 +1140,10 @@ pgagroal_validate_configuration(void* shm, bool has_unix_socket, bool has_main_s
       }
    }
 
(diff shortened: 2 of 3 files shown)


jesperpedersen:
Merged.

Issue #230, opened 2022-05-06
Last configuration line ignored if no new line at the end of file


fluca1978:
Running version 1.5.0.
If the configuration file does not contain a new file char at the end, the last line is ignored.
I discovered this while working on #227 because the system was not loading the `primary` hint from my replica configuration:

```
$ cat /etc/pgagroal/pgagroal.conf
...
[spengler]
host = spengler
port = 5432
primary = on$
```

while having the configuration file with an empty new line at the bottom makes it working:

```
$ cat /etc/pgagroal/pgagroal.conf
...
...
[spengler]
host = spengler
port = 5432
primary = on 
$
```

PR #231, opened 2022-05-06
[#230] Correctly read last line of configuration file if no new-line

fluca1978:
In the case the main configuration file has no new line in it, the
line is ignored by the function `extract_key_value()`.
This can be a little tricky, since the line is ignored and the user
has no clue about it.
With this fix, the last line of `pgagroal.conf` is always interpreted
even if no new line is present.

Fixes the same problem even for `pgagroal_hba.conf` file read within `extract_hba`.

Clse #230


jesperpedersen:
Merged.

Issue #232, opened 2022-05-09
Issuing a reload with a single server makes the status Not Init


fluca1978:
Running pgagroal 7fa65ed55a2e75de29fb38070e68772947b46f01.
The configuration has a single server:
```
$ cat /etc/pgagroal/pgagroal.conf
...
[venkman]
host = venkman
port = 5432
primary = on
```

Once pgagroal has started the situation is:

```
$ pgagroal-cli details
Status:              Running
Active connections:  0
Total connections:   2
Max connections:     5
---------------------
Server:              venkman
Host:                venkman
Port:                5432
State:               Primary
---------------------
...
```

but then if issuing a reload the status of the *only* available server is lost:

```
$ pgagroal-cli reload && pgagroal-cli details
Status:              Running
Active connections:  0
Total connections:   2
Max connections:     5
---------------------
Server:              venkman
Host:                venkman
Port:                5432
State:               Not init
---------------------
...
```

No changes to the configuration.
This is related to #223 and in my opinion should be fixed before that issue.

jesperpedersen:
We should probably have a `find_server` function that will loop through the existing servers, and find the `reload` definition.

In this precise case - we need to make sure that the atomic status flag is kept - if all the other information is the same.

PR #233, opened 2022-05-09
Avoid reload loose status

fluca1978:
This is a first attempt to solve the issue #232.
I've done two separated commits.
In the first one, there are a few new utility functions to catch a possible "misconfiguration" of the same server. This works on a positional basis (so fitrst server against the new first server and so on).
Handles also server reordering, that means it can tell the user the server order has changed.

The problem is to identify a server as "the same". So far I've built `is_same_connection()`, that allows us for changing the server name and primary status.

Now the problems becomes what we want to check: a user could change either a server name or a connection string. The latter is probably something we should avoid, and hence the implementation of `is_same_connection()`.

The second commit adds also another warning if the user has decreased the number of servers withing a reload: this probably means a restart should be required. At least, it look simpler than trying to understand what the user wants to do (e.g., delete a replica).

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -78,7 +78,7 @@ static int restart_limit(char* name, struct configuration* config, struct config
 static int restart_server(struct server* src, struct server* dst);
 
 static bool is_empty_string(char* s);
-static bool is_same_connection(struct server* s1, struct server* s2);
+static bool is_same_server(struct server* s1, struct server* s2);
 /**
  *
  */
@@ -2797,14 +2797,28 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
    return 0;
 }
 
+/**
+ * Checks if the configuration of the first server
+ * is the same as the configuration of the second server.
+ * So far it tests for the same connection string, meaning
+ * that the hostname and the port must be the same (i.e.,
+ * pointing to the same endpoint).
+ * It does not resolve the hostname, therefore 'localhost' and '127.0.0.1'
+ * are considered as different hosts.
+ * @return true if the server configurations look the same
+ */
 static bool
-is_same_connection(struct server* s1, struct server* s2)
+is_same_server(struct server* s1, struct server* s2)
 {
 
    if(!strncmp(s1->host, s2->host, MISC_LENGTH) && s1->port == s2->port)
+   {
       return true;
-
-   return false;
+   }
+   else
+   {
+      return false;
+   }
 }
 
 static void
@@ -2813,7 +2827,7 @@ copy_server(struct server* dst, struct server* src)
    atomic_schar state;
 
    // check the server cloned "seems" the same
-   if (is_same_connection(dst,src))
+   if (is_same_server(dst,src))
    {
       state = atomic_load(&dst->state);
    }
@@ -2912,9 +2926,9 @@ restart_server(struct server* src, struct server* dst)
 {
    char restart_message[MISC_LENGTH];
 
-   if (!is_same_connection(src, dst))
+   if (!is_same_server(src, dst))
    {
-      snprintf(restart_message, MISC_LENGTH, "Server <%s>, parameter  <host>", src->name);
+      snprintf(restart_message, MISC_LENGTH, "Server <%s>, parameter <host>", src->name);
       restart_string(restart_message, dst->host, src->host);
       snprintf(restart_message, MISC_LENGTH, "Server <%s>, parameter <port>", src->name);
       restart_int(restart_message, dst->port, src->port);
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -2738,7 +2738,7 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
    /* states */
 
    // decreasing the number of servers is probably a bad idea
-   if ( config->number_of_servers > reload->number_of_servers)
+   if (config->number_of_servers > reload->number_of_servers)
       restart_int("decreasing number of servers", config->number_of_servers, reload->number_of_servers);
 
    for (int i = 0; i < reload->number_of_servers; i++)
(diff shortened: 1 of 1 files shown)


src/libpgagroal/configuration.c
```diff
@@ -75,9 +75,10 @@ static void copy_user(struct user* dst, struct user* src);
 static int restart_int(char* name, int e, int n);
 static int restart_string(char* name, char* e, char* n);
 static int restart_limit(char* name, struct configuration* config, struct configuration* reload);
+static int restart_server(struct server* src, struct server* dst);
 
 static bool is_empty_string(char* s);
-
+static bool is_same_connection(struct server* s1, struct server* s2);
```
jesperpedersen, line 81: `is_same_server` is better

src/libpgagroal/configuration.c
```diff
@@ -2788,13 +2797,36 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
...
 
+static bool
+is_same_connection(struct server* s1, struct server* s2)
+{
+
+   if(!strncmp(s1->host, s2->host, MISC_LENGTH) && s1->port == s2->port)
+      return true;
+
+   return false;
+}
```
jesperpedersen, line 2808: uncrustify

src/libpgagroal/configuration.c
```diff
@@ -2875,6 +2907,23 @@ restart_limit(char* name, struct configuration* config, struct configuration* re
...
 }
 
+static int
+restart_server(struct server* src, struct server* dst)
+{
+   char restart_message[MISC_LENGTH];
+
+   if (!is_same_connection(src, dst))
+   {
+      snprintf(restart_message, MISC_LENGTH, "Server <%s>, parameter  <host>", src->name);
```
jesperpedersen, line 2917: Extra space in message

jesperpedersen (changes requested):

src/libpgagroal/configuration.c
fluca1978: Function name changed. Also added some documentation.

src/libpgagroal/configuration.c
```diff
@@ -2736,13 +2737,21 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
 
    /* states */
 
-   memset(&config->servers[0], 0, sizeof(struct server) * NUMBER_OF_SERVERS);
+   // decreasing the number of servers is probably a bad idea
+   if (config->number_of_servers > reload->number_of_servers)
+      restart_int("decreasing number of servers", config->number_of_servers, reload->number_of_servers);
+
```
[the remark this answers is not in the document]
fluca1978: I think you want this to stay at the moment, and when we'll have a safe way to handle sevrer reordering we could dig a little more on this.

src/libpgagroal/configuration.c
fluca1978: If there's an uncrustify configuration file to be used for the project I'm not able to find it.
In the meantime I've placed brackets, since indentation looks good to me (I'm using `ellemtel` codying style with 3 space indentation).

src/libpgagroal/configuration.c
fluca1978: Removed.

fluca1978:
One thing that the function `is_same_server()` does not do (and I've documented it) is that it does not detect the change of host as alias, e.g., `localhost` and `127.0.0.1` are considered different.
In order to catch such changes, we should use `getaddrinfo(3)` but I'm not sure if it is required at this level. In the case, please assign an issue.

src/libpgagroal/configuration.c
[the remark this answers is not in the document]
jesperpedersen: We will deal with it later, as there are a number of issues

src/libpgagroal/configuration.c
jesperpedersen: There is a `uncrustify.sh` in the root directory with a configuration file

src/libpgagroal/configuration.c
```diff
@@ -2788,13 +2797,50 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
...
+   atomic_schar state;
+
+   // check the server cloned "seems" the same
+   if (is_same_server(dst,src))
+   {
+      state = atomic_load(&dst->state);
+   }
+   else
+   {
+      state = SERVER_NOTINIT;
```
jesperpedersen, line 2836: Use atomic_init()

src/libpgagroal/configuration.c
```diff
@@ -2788,13 +2797,50 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
...
+ * that the hostname and the port must be the same (i.e.,
+ * pointing to the same endpoint).
+ * It does not resolve the hostname, therefore 'localhost' and '127.0.0.1'
+ * are considered as different hosts.
+ * @return true if the server configurations look the same
+ */
+static bool
+is_same_server(struct server* s1, struct server* s2)
+{
+
```
jesperpedersen, line 2813: Extra line

jesperpedersen (changes requested):

jesperpedersen:
I don't think we need to deal with "renamed" servers for now. The important part is that their configuration is sync'ed correctly, and their index doesn't change due to https://github.com/agroal/pgagroal/blob/master/src/include/pgagroal.h#L172

src/libpgagroal/configuration.c
fluca1978: shame on me, I did not see the `uncrustify.sh` script!
We should also put a guide or a note on how to contribute to code, e.g., run `uncrustify.sh` before submitting a patch.

src/libpgagroal/configuration.c
fluca1978: Fixed.

src/libpgagroal/configuration.c
jesperpedersen: Last section in `CONTRIBUTING.md`

jesperpedersen:
Merged, with formatting and commit message changes.

PR #236, opened 2022-05-12
[#235] Do not emit "restart" messages when adding a new server

### LINKED ISSUE #235: `pgagroal-cli reload` added server wrong warn about restart
Related to #233, when a new server is added to the configuration and the `pgagroal-cli reload` is issued, the system warns the user about a restart:

```
INFO  configuration.c:2881 Restart required for Server <spengler>, parameter <host> - Existing  New spengler
INFO  configuration.c:2869 Restart required for Server <spengler>, parameter <port> - Existing 0 New 5432

```

this is because the `restart_server()` function gets a new server that cannot be compared to a previous one, since it is added, and thus emits the warning. This is not a real bug, but a misleading message.

fluca1978:
The `restart_server()` function checks if two configurations identify
the same server.
When a new server is added, there is no previous configuration to
compare against, and therefore there is no need to emit a restart
warning message between the added server and "nothing".

Close #235

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -2745,7 +2745,12 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
 
    for (int i = 0; i < reload->number_of_servers; i++)
    {
-      restart_server(&reload->servers[i], &config->servers[i]);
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #237, opened 2022-05-12
[#234] `pgagroal-cli reload` transfer also log_rotation settings

### LINKED ISSUE #234: pgagroal-cli reload does not take into account log_rotation settings
When issuing a `pgagroal-cli reload` seting for `log_rotation_size` and similar are not transfer by `transfer_configuration`.

fluca1978:
When issuing a `pgagroal-cli reload` all `log_rotation` related
settings such as `log_rotation_age` and `log_rotation_size` must be
cloned into the reloaded configuration.

This also clone the `log_line_prefix` string to the new "reloaded"
configuration.

Close #234

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -2679,11 +2679,28 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
    /* log_type */
    restart_int("log_type", config->log_type, reload->log_type);
    config->log_level = reload->log_level;
+
    /* log_path */
-   restart_string("log_path", config->log_path, reload->log_path);
+   // if the log main parameters have changed, we need
+   // to restart the logging system
+   if (strncmp(config->log_path, reload->log_path, MISC_LENGTH)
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Do we need to do a lock with `log_lock` ?

fluca1978:
I'm not sure. If we don't lock, the last message on its way during the reload will hit the logs with the old rotation settings, that will be check at the very next log entry. Since we are not transferring the log rotation instant `next_log_rotation_age` I don't think we need to lock here.
What we could have to do, with a `pgagroal-cli reload` is to restart the logging system because it will adjust the path, the rotation parameters and so on. Or even better, restart the logging if any of the log settings in `log_rotation_age`, `log_rotation_size` have decreased and if `log_path` has changed.

jesperpedersen:
Yeah, the conditional restart sounds like a good idea

fluca1978:
The only doubt I have is if we need to use another process to restart the logging or we can leave it within the re-configuration phase as in c433f3c

PR #239, opened 2022-05-19
[#235] Wrap `pgagroal_prefill()`

### LINKED ISSUE #223: pgagroal-cli reload and added server

fluca1978:
In order to make calls to `pgagroal_prefill()` consistent around the
code and avoid repetitions, there is a new function named
`pgagroal_prefill_if_can()` that wraps several checks and does the
forking and prefilling.
Therefore there is no need to check against the configuration limits
and users before calling `pgagroal_prefill()` because
`pgagroal_prefill_if_can()` will do by itself.

Also `pgagroal_prefill_if_can()` will check for a primary server, and
in the case there is none, will avoid issuing the prefill at all.

Last, this introduces also the `pgagroal_can_prefill()` function that
is used in turn by `pgagroal_prefill_if_can()` and aims at testing the
configuration for the presence of limits and users. This function can
be used as a condition checker before doing clean-up stuff right
before calling `pgagroal_prefill_if_can()`.

Close #223

### DIFF
--- a/src/include/configuration.h
+++ b/src/include/configuration.h
@@ -182,6 +182,17 @@ pgagroal_reload_configuration(void);
 void
 pgagroal_init_pidfile_if_needed(struct configuration* config);
 
+/**
+ * Checks if the configuration has a min set of values to
+ * take into account doing a prefill. For example, there must
+ * be users and limits set, otherwise it does not
+ * make any sense to attempt a prefill.
+ * This can be used to wrap the condituion before calling
+ * other prefill functions, e.g., `pgagroal_prefill()`.
+ */
+bool
+pgagroal_can_prefill(struct configuration* config);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/include/configuration.h
+++ b/src/include/configuration.h
@@ -191,7 +191,7 @@ pgagroal_init_pidfile_if_needed(struct configuration* config);
  * other prefill functions, e.g., `pgagroal_prefill()`.
  */
 bool
-pgagroal_can_prefill(struct configuration* config);
+pgagroal_can_prefill(void);
 
 #ifdef __cplusplus
 }
--- a/src/include/pool.h
+++ b/src/include/pool.h
@@ -126,6 +126,24 @@ pgagroal_pool_shutdown(void);
 int
 pgagroal_pool_status(void);
 
+/**
+ * This function wraps around the logic to call `pgagroal_prefill()`.
+ * In order to avoid code repetition, this function can be used safely
+ * wherever there is the possibility to activate the prefill. The function
+ * does check if the configuration allows for a prefill, and in such case
+ * tries to `fork(2)` and executes the prefill.
+ * Also, the function checks for the presence of a primary with
+ * `pgagroal_get_primary()` and refuses to do a prefill if there
+ * is no primary at all.
+ *
+ * @param initial true if the prefill has to be done with the INITIAL
+ *        value of the pgagroal_database.conf file, false if it has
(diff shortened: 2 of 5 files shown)


src/include/configuration.h
```diff
@@ -182,6 +182,17 @@ pgagroal_reload_configuration(void);
...
+/**
+ * Checks if the configuration has a min set of values to
+ * take into account doing a prefill. For example, there must
+ * be users and limits set, otherwise it does not
+ * make any sense to attempt a prefill.
+ * This can be used to wrap the condituion before calling
+ * other prefill functions, e.g., `pgagroal_prefill()`.
+ */
+bool
+pgagroal_can_prefill(struct configuration* config);
```
jesperpedersen, line 194: You don't need `config` here -- just resolve it internally in the function

jesperpedersen (changes requested):

src/include/configuration.h
fluca1978: uhm, I don't fully agree. Having the `config` parameter clearly indicates this function is tied "only" to the configuration, not to other run-time parameters (e.g., the server is reachable). Second, I think this could also be useful when doing a `reload` to choose among new servers what can be used as a prefill-able one.
By the way, if you believe it is better to have the function without the parameter, I will make the change as a separated commit, so you can choose.

jesperpedersen:
Merged.

Issue #240, opened 2022-05-20
Cleanup configuration parsing


fluca1978:
This is somehow similar to #238: in the `configuration.c` file the function `pgagroal_read_configuration` is full of blocks like this one:

```
               if (!strcmp(key, "host"))
               {
                  if (!strcmp(section, "pgagroal"))
```

that checks for the `host` key parameter outside of the `pgagroal` section.
It seems to me the blocks are repeated quite a lot, and somehow unconsinstently. For example at <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L270> there is an extra check for `section` to be non empty, that is not present in other parts, that instead do an explicit `if ... else` on `section` (e.g., <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L247>.

My proposal is to create two different functions:
- `pgagroal_key_in_main_section( key )` will return true if `key` is defined within the `[pgagroal]` section;
- `pgagroal_key_in_section( key, section )` will return true if the `key` is defined in a custom user-defined `section`.

Therefore, this code:

```
               if (!strcmp(key, "host"))
               {
                  if (!strcmp(section, "pgagroal"))
                  {
                     // A  ...
                 }
                  else if (strlen(section) > 0)
                  {
                     // B   ....
                  }
                 else
                 { 
                    unknown = true;
                }
```

can be rewritten as

```
if ( pgagroal_key_in_main_section( "host" ) )
{  
   // A ...
} 
else if ( pgagroal_key_in_section( "host", section ) )
{
   /// B ...
}
else
{
   unknown = true;
}
```

Clearly this does not change the behavior of the configuration, but removes repetitions and should make `pgagroal_read_configuration()` more simple to new configuration parameter additions.

PR #242, opened 2022-05-23
[#240] Cleanup `pgagroal_read_configuration()`

fluca1978:
This commit introduces a few utility functions withing the
configuration code:
- `key_in_section()` is able to determine if a particular
configuration key is within a specific section, that could either the
main `[pgagroal]` section or a custom (i.e., server specific) one.
- `section_line()` handles a configuration line that appear to be a
section.
- `is_comment_line()` handles a line that starts with a comment.

Thanks to these functions, the code of `pgagroal_read_configuration()`
is shorter and also, to some extent, easier to read.

Moreover, in the case of an unknown configuration setting, the
`printf()` call has been changed to `fprintf()` to `stderr`, so that
the user is informed on the standard error.

Minor changes also to other configuration related functions.

Close [#240]

### DIFF
--- a/src/include/configuration.h
+++ b/src/include/configuration.h
@@ -35,6 +35,12 @@ extern "C" {
 
 #include <stdlib.h>
 
+/*
+ * The main section that must be present in the `pgagroal.conf`
+ * configuration file.
+ */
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

PR #243, opened 2022-05-23
[241] Remove `config` parameter from `pgagroal_init_pidfile_if_needed()`

### LINKED ISSUE #241: Remove parameter `config` in `pgagroal_init_pidfile_if_needed()` function
Functions in code should be all coherent and get the `struct configuration` out of the shared memory, not as an argument.

fluca1978:
Since the `struct configuration*` can be obtained out of shared
memory, functions should not use it as a parameter if not explicitly
needed.
This commit refactors the function `pgagroal_init_pidfile_if_needed()`
to avoid the need to pass a `struct configuration*` parameter.

Close #241

### DIFF
--- a/src/include/configuration.h
+++ b/src/include/configuration.h
@@ -180,7 +180,7 @@ pgagroal_reload_configuration(void);
  * of unix_socket_dir.
  */
 void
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

PR #246, opened 2022-05-25
[#245] Fix warnings about server sections.

### LINKED ISSUE #245: Unknown server parameter warnings
This is a bug that arises from #242 : 
```
$ pgagroal

Unknown key <host> with value <venkman> in section [venkman]
Unknown key <port> with value <5432> in section [venkman]
Unknown key <host> with value <spengler> in section [spengler]
Unknown key <port> with value <5432> in section [spengler]

```

**The system is working fine** but provides a set of warnings for *every* property within the server configuration sections.

The problem is that the new function `key_in_section()` does set the `unknown` fag when testing a global paramter (e.g., `host` before re-checking it for local section.

fluca1978:
Due to #242, when the `key_in_section()` function checks a paramter
that can be both global and local (e.g., `host`, `port`), the first
call to the function could set the `unknown` flag to true, because the
section does not match, and then the flag remains set even for the
other check.

This commit fixes the problem by resetting the `unknown` flag every
time the function `key_in_section()` is invoked. This is safe because
the function returns a false value if the section does not matches,
thus another if..else if branch is executed.

Close #245

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -199,7 +199,7 @@ pgagroal_read_configuration(void* shm, char* filename)
 
                //printf("\nSection <%s> key <%s> = <%s>", section, key, value);
 
-               if (key_in_section( "host", section, key, true, &unknown))
+               if (key_in_section( "host", section, key, true, NULL))
                {
                   max = strlen(value);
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

Issue #105, opened 2020-09-13
metrics: pgagroal_limit gauge


mrechte:
Would it be possible to add a value for number of waiting clients when pool limit is reached (for a given user/database) ?

jesperpedersen:
Yes, that should be possible.

The clients will time out based on `blocking_timeout` of course - if enabled.

fluca1978:
I've tried to design a possibile implementation, but since I'm really new to prometheus exporting, this could wrong.
The idea is to keep track with a counter of every time a connection goes suspended due to `blocking_timeout`, decrementing the counter immediatly after because the workflow of the connection is going to be restarted.
It seems to work on my manual tests, at least the datum is exported into the `/metrics/ page.

jesperpedersen:
Yes, your pull request looks correct for creating a new Prometheus metric.

This issue is about creating the blocking information for each limit entry defined. But, we need both metrics in the end :)

PR #252, opened 2022-05-26
[#105] Use atomic_load for pgagroal_limit_awaiting

fluca1978:
Since the array of connections_awaiting is defined as of type atomic,
use `atomic_load` and consequently `append_ulong`.


jesperpedersen:
Thanks !

PR #255, opened 2022-05-30
[#254] Document `pgagroal-cli` commands that work on all databases.

### LINKED ISSUE #254: not documented: `pgagroal-cli enable` and `pgagroal-cli disable` do not require a database name
`pgagroal-cli enable` and `pgagroal-cli disable` do not required a database name. In such case the system will enable or disable all configured databases.
This is not documented at all.

fluca1978:
This makes more clear that those commands that work against a database
do accept also the `*` special string or no argument at all to issue
the command against all available databases.

Close #254


jesperpedersen:
Merged.

Issue #219, opened 2022-04-11
Option to cache Prometheus response


jesperpedersen:
Have an option that will cache a Prometheus response for `X` seconds

fluca1978:
Just curious: do you mean avoid to send out the result of `metrics_page()` at every single request but to serve it as it was `x` seconds before?
Is such functon so heavy to serve that requires caching?
Could it be a static temporary file the solution? I mean write out `metrics_page()` to the file and serve such a file is still valid.

jesperpedersen:
Prometheus requests are light-weight for pgagroal (and pgmoneta, but not for pgexporter).

I would just cache it as a `static` string, together with the last served timestamp.

PR #256, opened 2022-05-30
[#219] Provide Prometheus metrics cache

fluca1978:
This commit refactors the way the Prometheus metrics are served in
order to provide the capability to cache the responses for a specified
amount of time.

A new setting, named `metrics_cache` is defined and parsed as an "age"
string. The value will be stored as a number of seconds.
This required the refactoring of some configuration code, with
particular regard to `as_logging_rotation_age()` in order to extract
the "as age" logic into an utility function that can be used for
different "age"-like strings. Moreover, some methods and configuration
parameters have been changed from `int` to `unsigned int` (e.g., it
does not make any sense to have a log rotation age negative!). The
configuration related code has been changed to intercept wrong
strings, e.g., when `log_rotation_age = -2m` the system disables the
rotation and warns the user as the parameter is unknown. Also `+1m` is
considered a bogus string.

The `struct prometheus` has been expanded with an inner struct, named
`struct prometheus_cache` that holds the cache payload and the
timestamp the cache will be valid until.

The `metrics_page()` has been refactored to check if the cache is
configured and valid, in such case the response is served out of
cache. On the other hand, the cache is reset and the response is
served as a new one, with all the inner methods adding the result to
the cache for the next time.
A few utility functions have been introduced:
- `is_metrics_cache_configured()` checks that the current
configuration allows for the usage of cache, i.e., `metrics_cache`
have been set;
- `is_metrics_cache_valid()` checks hat the cache does handle some
good stuff to be served;
- `metrics_cache_reset()` to reset the cache, once it has been set
within the shared memory;
- `metrics_cache_add()` appends data to the cache. This methods is
invoked whenever a new Prometheus message is spurt. In the case the
overall amount of data overflows, the cache is made not valid. It is
safe to call this function everywhere because it does nothing if the
caching system is not enabled.

If the cache overflows, it is set to "invalid" so that is not gong to
be served anymore. The max size of the cache is defined as a constant
value within the source code, and is set to 8kB (that should be
reasonable since the average Prometheus response on a unloaded server
is around 4kB).

The documentation has been updated to reflect changes.

Close #239


fluca1978:
I've added also another cache controlling parameter: `metrics_cache_max_size`. Thanks to this, the user can decide to cache whatever is within `x` bytes, so can fine control the size of the cached response. This however does not allocate nor control memory.

doc/CONFIGURATION.md
```diff
@@ -39,6 +39,8 @@ See a more complete [sample](./etc/pgagroal.conf) configuration for running `pga
 | port | | Int | Yes | The bind port for pgagroal |
 | unix_socket_dir | | String | Yes | The Unix Domain Socket location |
 | metrics | 0 | Int | No | The metrics port (disable = 0) |
+| metrics_cache_max_age | 0 | String | No | The number of seconds to cache the metrics response. Can be a string including a modifier suffix, e.g., `2M` for two minutes. A value of zero disables caching. |
+| metrics_cache_max_size | 8192 | String | No | An optional value, in bytes, to control the metrics cache. If a single responses is greater than this value, the cache will be temporarely disabled. This value does not allocate any memory, and is not taken into account if `metrics_cache_max_age` is set to `0` (i.e., cache is disabled). It can be expressed as a size-string, e.g., 2K as for 2 kilobytes, |
```
jesperpedersen, line 43: We should probably use 256k as the default - e.g. 262144

src/include/pgagroal.h
```diff
@@ -228,6 +228,30 @@ struct prometheus_connection
    atomic_ullong query_count;           /**< The number of queries per connection */
 } __attribute__ ((aligned (64)));
 
+/**
+ * Max size of the cache (in bytes).
+ * If the cache request exceeds this size
+ * the caching should be aborted in some way.
+ */
+#define PROMETHEUS_MAX_CACHE_SIZE 8192
```
jesperpedersen, line 236: This should be something like 1M - e.g. 1024*1024

We will need a PROMETHEUS_DEFAULT_CACHE_SIZE as well at 256 * 1024

src/include/pgagroal.h
```diff
@@ -228,6 +228,30 @@ struct prometheus_connection
...
+ * settings.
+ *
+ * The `valid_until` field stores the result
+ * of `time(2)` but instead of using a `time_t`
+ * there is an `atomic_ullong` to allow for
+ * concurrency management.
+ */
+struct prometheus_cache
+{
+   char data[PROMETHEUS_MAX_CACHE_SIZE];  /**< the payload */
```
jesperpedersen, line 251: This should be a dynamic value - e.g. `char*`

src/include/pgagroal.h
```diff
@@ -228,6 +228,30 @@ struct prometheus_connection
...
+   atomic_ullong valid_until;             /**< when the cache will become not valid */
```
jesperpedersen, line 252: We will need a `size_t cache_size` and an `atomic_schar prometheus_lock;` as well

src/libpgagroal/configuration.c
```diff
@@ -265,6 +267,20 @@ pgagroal_read_configuration(void* shm, char* filename, bool emitWarnings)
...
+               else if (key_in_section("metrics_cache_max_age", section, key, true, &unknown))
+               {
+                  if (as_seconds(value, &config->metrics_cache_max_age, PGAGROAL_PROMETHEUS_CACHE_DISABLED))
+                  {
+                     unknown = true;
+                  }
+               }
+               else if (key_in_section("metrics_cache_max_size", section, key, true, &unknown))
+               {
+                  if (as_bytes(value, &config->metrics_cache_max_size, PROMETHEUS_MAX_CACHE_SIZE))
```
jesperpedersen, line 279: We need to pass in `PROMETHEUS_DEFAULT_CACHE_SIZE` as well -- capped at `PROMETHEUS_MAX_CACHE_SIZE`

src/libpgagroal/prometheus.c
```diff
@@ -1110,53 +1119,76 @@ metrics_page(int client_fd)
...
    int status;
    struct message msg;
+   struct prometheus* prometheus;
+
+   prometheus = (struct prometheus*)prometheus_shmem;
 
    memset(&msg, 0, sizeof(struct message));
 
-   now = time(NULL);
+   // can serve the message out of cache?
```
jesperpedersen, line 1128: If we are using the cache then we need to do locking with `prometheus_lock` such that we don't serve a partial result

src/libpgagroal/prometheus.c
```diff
@@ -2021,3 +2065,164 @@ append_ullong(char* orig, unsigned long long l)
...
+   if (config->metrics == 0)
+   {
+      return false;
+   }
+
+   if (config->metrics_cache_max_age != PGAGROAL_PROMETHEUS_CACHE_DISABLED)
+   {
+      return true;
+   }
+   else
```
jesperpedersen, line 2094: Just return false, no need for the `else`

src/libpgagroal/prometheus.c
```diff
@@ -2021,3 +2065,164 @@ append_ullong(char* orig, unsigned long long l)
...
+ *
+ */
+static void
+metrics_cache_reset(void)
+{
+   struct prometheus* prometheus;
+
+   prometheus = (struct prometheus*)prometheus_shmem;
+
+   memset(prometheus->cache.data, 0, PROMETHEUS_MAX_CACHE_SIZE);
```
jesperpedersen, line 2226: Use `cache.size`

jesperpedersen (changes requested):

jesperpedersen:
Maybe we don't need the `prometheus_lock` if we can find a way that 2+ processes doesn't initialize the result at the same time... Maybe it doesn't matter...

jesperpedersen:
The overhead of the lock is minimal though and scraping shouldn't be done in a loop anyway

src/include/pgagroal.h
fluca1978: Agree. I was not sure about the cache size to use.

src/libpgagroal/configuration.c
fluca1978: Done.

src/libpgagroal/prometheus.c
fluca1978: Fine, I was not sure about styling...

src/libpgagroal/prometheus.c
```diff
@@ -2021,3 +2065,164 @@ append_ullong(char* orig, unsigned long long l)
...
+ *
+ *
+ * @param data the payload string to store
+ * @param updateTimestamp true if this invocation will update the
+ * `valid_until` value, that is this is the last call to append data to
+ * the cache.
+ *
+ */
+static void
+metrics_cache_add(char* data, bool updateTimestamp)
```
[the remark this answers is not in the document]
jesperpedersen: Well, if `add|append` overflows with `MAX_CACHE_SIZE` it will deallocate and set to `NULL` - in that case `timestamp` won't update

src/include/pgagroal.h
fluca1978: Forgive me if this is trivial, but I've a doubt. If we change the sized array to a `char *` it means we are going to `malloc` and `free` the pointer as the cache content grows and shrinks.
But such pointer is `mmap`ed into the `struct prometheus`, so if we don't allocate memory before the `mmap` call, processes will only share the pointer, not the content of the cache.
And `metrics_page()` is served by a `fork`ed process.
Doesn't this mean that `char *` needs special handling in this case?

jesperpedersen: It is in a shared memory segment, so it will only be allocated once - and `memset` when we need it.

fluca1978: I'm not so sure, for example when the configuration changes and the `pgagroal-cli reload` imposes a grow on the cache size, the `struct prometheus` is not going to be reallocated, so the inner carhe is not too, but it has to grow.
However, I will start implementing it and see when and where we are facing troubles.
Surely this is a little more than "caching to a string"!

fluca1978:
Commit ad85093d90d97357e6a88e95b749700881e7415c does pretty much everything, I hope.
Please note that `metrics_cache_alloc()` must initialize the shared memory of the cache. I cannot see a way to do it in `pgagroal_init_prometheus()` even increasing `tmp_p_size` by the `metrics_cache_size_to_alloc()`. If this is not correct, as I suspect, please advice.

Another problem is when `pgagroal-cli reload` is issued and `metrics_cache_max_size` is changed: since `struct prometheus` have been already initialized, how can we handle a re-`metrics_cache_alloc()` (clearly with a previous destruction of the memory)? S far, I issue an hint to restart `pgagroal`.

fluca1978:
One possible solution to the `pgagroal-cli reload` and `metrics_cache_max_size` problem could be this:
- within `metrics_page()` when the cache is locked we check if `cache.size` is different from `config->metrics_cache_max_size`
- if equal, do nothing
- if different, free `cache.data` and realloc as shared memory
- proceed with the `metrics_cache_append()` workflow as usual

fluca1978:
I'm closing this because we are diverging too much in my opinion.
I don't like force pushing, so I'm going to open a new PR.
By the way, I liked the usage of `struct prometheus` the most instead of having its own shared memory.

jesperpedersen:
Yeah, but you can only have one FMA per struct, which means that it had to be a static char array allocating 256k of memory even when it isn't going to be used

PR #258, opened 2022-06-03
[#257] Refactoring of the configuration age and size parsing.

### LINKED ISSUE #257: Refactor configuration age parsing to provide support for other configuration options
Related to the work in #256 : the `as_log_rotation_age()` function parses a string like `2h` to extract the corresponding number of seconds.
It is required to extract an utility function to parse the above string independently from the logging related stuff, so that it can be used for other configuration parameters that will be introduced.
So far, this is mandatory to introduce the `metrics_cache` parameter for #219 and #256.

fluca1978:
This commit introduces two utility functions:
- `as_seconds()` that parses an "age" string (e.g., "2d") and returns the
   extracted value as seconds;
- `as_bytes()` parses a "size" string (e.g., "2m") and returns the
  number of bytes the string represents.

Such utility functions can be used as backbone for introducing new configuration
parameters, so far they are used as delegates
for `as_log_rotation_age()` and `as_log_rotation_size()`.

Also, converts the `log_rotation_age` and `log_rotation_size` configuration
parameters to `unisgend int`: it does not make any sense to have a negative
rotation value.

In the case a configuration parameter for an age or a size is expressed
as a negative value, a warn about the "unknown" configuration value
is issued.

Documentation updated.

Close #257


jesperpedersen:
Merged.

PR #260, opened 2022-06-03
[#259] Omit configuration warnings when running cli

### LINKED ISSUE #259: `pgagroal-cli` reporting configuration errors
If the configuration contains errors, the `pgagroal_read_configuration()` function emits on `stderr` some warnings.
Since `pgagroal-cli` exploits the same function, the same warning are emitted also on `stderr` whenever `pgagroal-cli` is called:

```
% pgagroal-cli status

Unknown key <log_rotation_age> with value <-2d> in section [pgagroal]
Unknown key <log_rotation_size> with value <-3m> in section [pgagroal]
```

I don't have a particular opinion here, but I would like `pgagroal-cli` to not emit the same warnings the `pgagroal` daemon emits, since the daemon could have "fixed" errors by itself (e.g., disable the log rotation).
I would suggest just to print out a summary message (e.g., "you have unknown values in configuration") or disable such warnings at all.

fluca1978:
`pgagroal-cli` reparses the configuration file as the daemon.
Due to that, if there is a problem with the configuration a warning
is emitted to the stderr.
This commit avoids the `pgagroal-cli` to report warnings on the command line
since such problems could have been alread fixed by the running daemon.

Close #259

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -237,7 +237,7 @@ main(int argc, char** argv)
 
    if (configuration_path != NULL)
    {
-      ret = pgagroal_read_configuration(shmem, configuration_path);
+      ret = pgagroal_read_configuration(shmem, configuration_path, false);
       if (ret)
       {
          printf("pgagroal-cli: Configuration not found: %s\n", configuration_path);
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Merged.

PR #262, opened 2022-06-09
[#219] Provide Prometheus metrics cache

### LINKED ISSUE #219: Option to cache Prometheus response

fluca1978:
This commit refactors the way the Prometheus metrics are served in
order to provide the capability to cache the responses for a specified
amount of time.

Two new configuration settings have been introduced:
- `metrics_cache_max_age`, expressed in seconds, defines for how long
a cached response can be served without having to compute a new one
set of metrics. This parameter is required for the caching machinery
to be enabled.
- `metrics_cache_max_size` allows the user to avoid caching of
responses greater than the expressed value in bytes.

In any case, a cache response is tied to 1MegaByte, by a value set in
the source code.

If the cache overflows the 1MB memory limit, it is immediatly marked
as invalid, that is the system is not going to serve the cached
result. However, as soon as a new Prometheus request arrives, the
cache will be re-filled from scratch, so if the current answer is
lower than 1MB and lower than the optional `metrics_cache_max_size`,
the response will be cached.
Whenever an overflow of the cache happens, the user is warned. Due to
the nature the cache is populated, the user could see more than one
single advice in logs (at level DEBUG).

The cache is built as a shared memory segment (see #256 and in
particular
<https://github.com/agroal/pgagroal/pull/256#issuecomment-1150239298>).
The cache is implemented by means of the `struct prometheus_cache`
that has a dynamically allocated (`mmap()`) segment to handle the
payload. This means that the `metrics_cache_max_size` parameter will
drive the amount of memory dynamically allocated to be used as a cache.

New static utility functions have been added:
- `metrics_cache_append(char* data)` that appends the data to the
cache in the case it is safe to do it, that is no overflow in the
cache size. In case the data to append makes the cache to overflow,
the cache is invalidated;
- `metrics_cache_finalize(void)` set the timestamp for the cache
vlidation.
- `metrics_cache_invalidate(void)` set the cache as invalid by
zero-filling the data. Used for example when an overflow is detected.
- `metrics_cache_size_to_alloc(void)` computes the size, in bytes,
required for the cache payload allocation. The size is set to the
default, or `metrics_cache_max_size` if configured and in any case
nothing more than the max size of 1MB.

New "public" utility function:
- `pgagroal_init_prometheus_cache()` will initialize the cache shared
memory according to the user configuration and the result of
`metrics_cache_size_to_alloc()`.

The `metrics_page()` method has been refactored to check if the cache
is valid, in such case the response is served directly out of
cache. If the cache is not valid, the function gets the cache lock and
invokes several times the append function (in inner method
calls). Last, it finalizes the cache and release the lock.

If the optional configuration parameter `metrics_cache_max_size`
changes, the system does not reallocate memory, rather it issue a
warning about the need for a restart.

The documentation has been updated to reflect changes.

Close #219

### DIFF
--- a/src/libpgagroal/prometheus.c
+++ b/src/libpgagroal/prometheus.c
@@ -91,6 +91,13 @@ static char* append_int(char* orig, int i);
 static char* append_ulong(char* orig, unsigned long i);
 static char* append_ullong(char* orig, unsigned long long l);
 
+static bool is_metrics_cache_configured(void);
+static bool is_metrics_cache_valid(void);
+static bool metrics_cache_append(char* data);
+static bool metrics_cache_finalize(void);
+static size_t metrics_cache_size_to_alloc(void);
+static void metrics_cache_invalidate(void);
+
 void
 pgagroal_prometheus(int client_fd)
 {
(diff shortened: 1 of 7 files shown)


fluca1978:
> You will need to lock the cache in all cases, otherwise you could end up serving a zero cache - so the lock has to be moved in metrics_page.

I'm not sure, but it seems to me that the lock is fine where it is. However, the lock has to be granted before serving out the cache, that is true, so the `metrics_page()` now, as by 31e4e731835a1c3dea9abee116f8410147516339, acquires the lock before either serving or building a response.

> curl: (56) Illegal or missing hexadecimal sequence in chunked-encoding

Interesting, I tried it out with a browser without having the problem. It turned out that the problem is that we announce the HTTP response to be chunked, so either the cache should spurt the whole answer or, as in the current commit, it uses `send_chunk()` to send out the cache content.
At least, this seems to work with `curl` and browsers. However, the cached answer has an extra '\n' that should not cause any trouble but I've not figured out from where it originates.

fluca1978:
Sorry for the noise, commit 447b19abc96b1843b4b4b1f8a0a2375c5253c3b0 should fix all.

I've fixed the `send_chunk` and extra newline problem: when having a response in cache, we don't send chunks but a single shot.
I've also added an extra before-error cache invalidation.
Last, the lock should be fine.

works with `curl` and browsers

fluca1978:
Should this work, once it has been accepted, bump the version number at <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/prometheus.c#L1123>?

src/libpgagroal/prometheus.c
```diff
@@ -1110,54 +1117,105 @@ metrics_page(int client_fd)
...
+      // can serve the message out of cache?
+      if (is_metrics_cache_configured() && is_metrics_cache_valid())
+      {
+         // serve the message directly out of the cache
+         pgagroal_log_debug("Serving metrics out of cache (%d/%d bytes valid until %lld)",
+                            strlen(cache->data),
+                            cache->size,
+                            cache->valid_until);
+
+         //status = send_chunk(client_fd, cache->data);
```
jesperpedersen, line 1140: You can remove this line :)

jesperpedersen (changes requested):

src/libpgagroal/prometheus.c
fluca1978: shame on me!

jesperpedersen:
Merged - great work !

PR #264, opened 2022-06-13
[#263] Use a macro to perform the process-sleep.

### LINKED ISSUE #263: Use a C macro to sleep?
I think this could improve code readibility: there are several places (not many) where there is a code like fhis:

```
      else
      { 
      /* Sleep for 1ms */
        struct timespec ts; 
        ts.tv_sec = 0; 
        ts.tv_nsec = 1000000L; 
        nanosleep(&ts, NULL); 

       goto retry; 
      } 
```

such code performs a sleep and a jump, and is used when asking for a lock acquisition.
Cannot we wrap the code in a macro so that the above becomes:

```
      else
          SLEEP_AND_GOTO(1000000L,retry)
```

Could this improve source code readibility and avoid duplication of code?

fluca1978:
Introduces two macros:
- `SLEEP` to perform a nanosleep call with the specified amount of
time expressed as nanoseconds;
- `SLEEP_AND_GOTO` does the same of the above but adds also a `goto`
to the specified label.

Example of usage:

      SLEEP_AND_GOTO(1000000L,retry)

is the same as

      {
         /* Sleep for 1ms */
         struct timespec ts;
         ts.tv_sec = 0;
         ts.tv_nsec = 1000000L;
         nanosleep(&ts, NULL);

         goto retry;
     }

Close #263

### DIFF
--- a/src/libpgagroal/pipeline_transaction.c
+++ b/src/libpgagroal/pipeline_transaction.c
@@ -144,10 +144,7 @@ transaction_start(struct ev_loop* loop, struct worker_io* w)
    if (is_new)
    {
       /* Sleep for 5ms */
-      struct timespec ts;
-      ts.tv_sec = 0;
-      ts.tv_nsec = 5000000L;
-      nanosleep(&ts, NULL);
+      SLEEP(5000000L)
    }
 
    return;
--- a/src/include/pgagroal.h
+++ b/src/include/pgagroal.h
@@ -132,6 +132,47 @@ extern "C" {
       __typeof__ (b) _b = (b);  \
       _a < _b ? _a : _b; })
 
+/*
+ * Common piece of code to perform a sleeping.
+ *
+ * @param zzz the amount of time to
+ * sleep, expressed as nanoseconds.
+ *
+ * Example
+   SLEEP(5000000L)
+ *
+ */
+#define SLEEP(zzz)                  \
+   {                                \
+      struct timespec ts_private;   \
+      ts_private.tv_sec = 0;        \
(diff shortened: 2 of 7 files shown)


src/libpgagroal/pipeline_transaction.c
```diff
@@ -144,13 +144,10 @@ transaction_start(struct ev_loop* loop, struct worker_io* w)
...
       /* Sleep for 5ms */
-      struct timespec ts;
-      ts.tv_sec = 0;
-      ts.tv_nsec = 5000000L;
-      nanosleep(&ts, NULL);
-   }
+      SLEEP(5000000L)
 
-   return;
+      return;
```
jesperpedersen, line 149: You moved the `return;` statement under the `if`

jesperpedersen (changes requested):

src/libpgagroal/pipeline_transaction.c
fluca1978: It took me some time to understand how did I mess such branch.
It turned out it was `uncrustify` that added the extra curly braces after I wrote the code as:

```
if (is_new)
   SLEEP(...)

return;
```

Since I run uncrusityf as the last thing before pushing, I did not noticed the code changed to:

```
if (is_new)
{
   SLEEP(..)
  return;
}
```

Commit 6a62e4b6a0a62d68df62ec887430424a8641a3f6 fixes this by adding an extrac curly braces pair to the `if`.
I'm not an expert in using `uncrustify` but probably we should document this or fix (if it can understand it is a macro, I suspect the absence of `;` is causing the problem).

PR #266, opened 2022-06-13
[#265] Match HTML `p` tags in home page

### LINKED ISSUE #265: Match HTML `p` tags in home page
The `prometheus.c` `home_page`  functuion outputs a set of information that have a structure like:

```
<h2>tite</h2>
bla bla
<p>
```

while it should be something like:

'''
<h2>title</h2>
<p>
bla bla
</p>
```

so that `p` tags are always matched. This does not cause at the moment any problem with browsers, but makes the HTML code cleaner and more correct.

### PR DESCRIPTION
**Pull request by fluca1978**
The `home_page()` function provides an HTML page with the description
of all the metrics exported to Prometheus.
The HTML code was missing the `<p>` and `</p>` pairs, only one `<p>`
was present at the end of each section.
This commit fixes the problem and indents the code of one space within
each paragraph.

Close #265

### CODE REVIEW

**Comment by jesperpedersen** [comment]
Nice.

You would expand it with the full header ?

```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
                     http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
<head>
  <title>pgagroal exporter</title>
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
</head>
```

fluca1978:
> <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"

as far as I know, HTML 5 requires a simple `DOCTYPE`, in fact the validator is raising an error with your `DOCTYPE` example.
I've appended other stuff and checked the validator passes.

jesperpedersen:
Merged.

PR #276, opened 2022-06-16
[#275] `pgagraol-cli` gives priority to remote options when specified

### LINKED ISSUE #275: `pgagroal-cli` remote management not detected on localhost
See <https://github.com/agroal/pgagroal/discussions/269#discussioncomment-2963480>: when `pgagroal-cli` is used on the same host of the pooler, it detects the main configuration file and does not honor `host` and `port` options if specified.

fluca1978:
If the user specifies both `-h` and `-p` as command line parameters
she wants to connect remotely and not via Unix socket, even if the
configuration file is present (e.g., running on `localhost`).

Close #275

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -150,6 +150,7 @@ main(int argc, char** argv)
    char un[MAX_USERNAME_LENGTH];
    char* server = NULL;
    struct configuration* config = NULL;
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #279, opened 2022-06-20
[#278] `pgagroal-cli` username/password allowed only on remote conns

### LINKED ISSUE #278: `pgagroal-cli` incompatible (or incomplete) options
Using `pgagroal` 019ea11d89b549a21cfb28d578325728335e6420, while working on #253 I discovered by accident that the following seems to work:

```
$ pgagroal-cli status -U luca -P secret
Status:              Running
Active connections:  0
Total connections:   1
Max connections:     4
Segmentation fault (core dumped)
```

and also something more interesting that could reveal the issue of the segmentation fault:

```
$ pgagroal-cli details -U luca -P secret
...
Connection    0:     Free            2022-06-20 07:07:24        21     pgbench pgbench 
Connection    1:     Not initialized                                     
Connection    2:     Not initialized                                     
Connection    3:     Not initialized                                     
free(): invalid pointer
Aborted (core dumped)
```

The point here is that the `-P` and `-U` options should not be taken into account at all when not providing a remote connection information. So the `pgagroal-cli` program should warn the user that she has mixed arguments.
I suspect also that the core dumped is related to some remote connection information not properly allocated and that is going to be freed.

fluca1978:
This commit ensures that when the user specifies either a `-U` or `-P`
flag on the command line, `pgagroal-cli` ensures that also `-h` and
`-p` have been specified, otherwise it claims the remote connection
cannot be established.

Close #278

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -229,6 +229,14 @@ main(int argc, char** argv)
       }
    }
 
+   // if the user has specified either a username or a password
+   // there must be all the other pieces for a remote connection
+   if ((username != NULL || password != NULL) && !remote_connection)
+   {
+      printf("pgagroal-cli: you need also -h and -p options to perform a remote connection\n");
+      exit(1);
+   }
+
    // and she cannot use "local" and "remote" connections at the same time
    if (configuration_path != NULL && remote_connection)
    {


jesperpedersen:
Merged.

PR #283, opened 2022-06-28
[#282] Reduce code duplication while reading configuration.

### LINKED ISSUE #282: Reduce configuration read branches in `main`
I've seen that in the `main` of `pgagroal.c` there are blocks that read the various configuration files, that look all the same:

```
if ( <user has specified a path> )
{ 
    if ( <cannot load the path> )
   { 
      error;
  }
else 
{
    if ( <cannot load default path> )
    {
       error;
    }
}
```

As an example, see <https://github.com/agroal/pgagroal/blob/master/src/main.c#L406>:

```
if (configuration_path != NULL)
{
    if (pgagroal_read_configuration(shmem, configuration_path, true))
    {
        printf("pgagroal: Configuration not found: %s\n", configuration_path);
#ifdef HAVE_LINUX
        sd_notifyf(0, "STATUS=Configuration not found: %s", configuration_path);
#endif
        exit(1);
    }
}
else
{
    if (pgagroal_read_configuration(shmem, "/etc/pgagroal/pgagroal.conf", true))
    {
        printf("pgagroal: Configuration not found: /etc/pgagroal/pgagroal.conf\n");
#ifdef HAVE_LINUX
        sd_notify(0, "STATUS=Configuration not found: /etc/pgagroal/pgagroal.conf");
#endif
        exit(1);
    }
    configuration_path = "/etc/pgagroal/pgagroal.conf";
}
```

Since the branches are identical, I propose a change like the following (untested yet):

```
configuration_path = configuration_path != NULL ? configuration_path : "/etc/pgagroal/pgagroal.conf";
    if (pgagroal_read_configuration(shmem, configuration_path, true))
    {
        printf("pgagroal: Configuration not found: %s\n", configuration_path);
#ifdef HAVE_LINUX
        sd_notifyf(0, "STATUS=Configuration not found: %s", configuration_path);
#endif
        exit(1);
    }

```

so that we set the path variable first, then a single branch to test for errors, and in any case we exit, so it does not mind if the variable already contains the path.
This should improve code readability and provide less room for copy and paste errors.

fluca1978:
This commit introduces a few constants to indicate the default files
`pgagroal` is going to read the configuration from when the user does
not specify anything else on the command line.

All the constants have the name `PGAGROAL_DEFAULT_<what>_FILE`, like
for example `PGAGROAL_DEFAULT_CONF_FILE` and
`PGAGROAL_DEFAULT_HBA_FILE`.

The idea is to give priority to what the user has specified on the
command line, and then if nothing has been specfiied, use the default
file path. This was already done, but in a more verbose way.

This commit uniforms the error checking for all the cases, when the
file has been specified or it is the default one, so that for
instance, the default file is checked against the same type of
troubles the user could have introduced into custom files.

The online help has been updated to indicate where the default file
for every option is located.

The command line tool `pgagroal-cli` has been adjusted to use the new
constants and get a coherent online help.

Close #282

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -96,6 +96,7 @@ usage(void)
    printf("\n");
    printf("Options:\n");
    printf("  -c, --config CONFIG_FILE Set the path to the pgagroal.conf file\n");
+   printf("                           Default: %s\n", PGAGROAL_DEFAULT_CONF_FILE);
    printf("  -h, --host HOST          Set the host name\n");
    printf("  -p, --port PORT          Set the port number\n");
    printf("  -U, --user USERNAME      Set the user name\n");
@@ -285,7 +286,7 @@ main(int argc, char** argv)
    }
    else
    {
-      ret = pgagroal_read_configuration(shmem, "/etc/pgagroal/pgagroal.conf", false);
+      ret = pgagroal_read_configuration(shmem, PGAGROAL_DEFAULT_CONF_FILE, false);
       if (ret)
       {
          if (!remote_connection)
@@ -296,7 +297,7 @@ main(int argc, char** argv)
       }
       else
       {
-         configuration_path = "/etc/pgagroal/pgagroal.conf";
+         configuration_path = PGAGROAL_DEFAULT_CONF_FILE;
 
          if (logfile)
          {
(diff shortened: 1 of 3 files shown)


jesperpedersen:
Merged.

Issue #215, opened 2022-02-25
process title is too much short


fluca1978:
Version 1.4 (`HEAD`):
```
% pgagroal -d       
% ps -auxw | grep pgagr
luca       16998  0.0  0.3  69316  2656 ?        Ss   05:59   0:00 pgagroal: ma
```

the process title should be `pgagroal: main`, according to https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/utils.c#L774 but it is truncated at https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/utils.c#L777  because `max_process_title_size` is lower. The problem appears to be here https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/utils.c#L762 because the `max_process_title` is computed on the basis of `argv` and therefore it depends on the length of the command line.
In other words, invoking `pgagroal` will result in `max_process_title_suze` to be `9`, invoking `pgagroal -d` will make `max_process-title_size` to be 12, and `/usr/local/bin/pgagroal -d` will make `max_process_title_size` bigger.

I don't know if we should care about `argv` at such level, I don't know if there's a way to get a safe length for the process title and then to use such as `max_process_title_size` default value.

jesperpedersen:
Note, that this is a PITA area to deal with - see the numerous updates to the function in question.

Make sure that you verify any patches on a harden OS like RHEL 8 or Rocky8 - as they are likely to give you a different result; especially if the string is larger than `argv`.

fluca1978:
I'm not expert in this area, but as far as I can tell,  there's no way to update the process title in Linux without overwriting environment, while other platforms provide functions (.e.g, BSD `setproctitle`). I suspect that you are referring to the fact that the hardened systems will complain about a process title changing over its initial size.
I will investigate on this as soon as I've time.

jesperpedersen:
Yeah, having `setproctitle` on Linux would be much easier... Don't break too many keyboards on this though

fluca1978:
Seems however to work on Rocky Linux 8.5. I'm trying to figure out when it is not working.

PR #284, opened 2022-06-28
[#215] Process title length

fluca1978:
This is a brute force approach to set the process title with the
appropriate length.
The idea is to compute the size of the process title as it has been
requested, and set it to the computed size.

This works on Rocky Linux release 8.6 (Green Obsidian).

It could break some aggressive security-enforced environments.

See #215

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -254,6 +254,20 @@ pgagroal_base64_decode(char* encoded, size_t encoded_length, char** raw, int* ra
 void
 pgagroal_set_proc_title(int argc, char** argv, char* s1, char* s2);
 
+/**
+ * Sets the process title for a given connection.
+ *
+ * Uses `pgagroal_set_proc_title` to build an information string
+ * with the form
+ *    user@host:port/database
+ *
+ * @param argc the number of arguments
+ * @param argv command line arguments
+ * @param connection the struct connection pointer for the established connection.
+ */
+void
+pgagroal_set_connection_proc_title(int argc, char** argv, struct connection* connection);
+
 #ifdef DEBUG
 
 /**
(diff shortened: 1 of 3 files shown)


jesperpedersen:
I'm getting

```
pgagroal: main  agroal.conf -a pgagroal_hba.conf -u pgagroal_users.conf
```

now instead of

```
pgagroal: main
```

using Fedora 36

fluca1978:
Shame on me, I was focusing on having a command line *shorter* than what was required.
Commit 29a0cb53a732e1b0f1103f3317867fcab9ab5e5a should fix the problem for both commad lines shorter or longer.

I also added a separated commit 52a585a047d58039fd240a6220893567beeedaec that creates a process title in the form `user@host:port/database` like for instance `pgbench@venkman:5432/pgbench`.
I pushed it as a separated commit because I've a few doubts about it:
- it could be expensive to seek for the primary every time the process title has to be updated (this happens on every connection);
- it could the case of introducing a setting, like `update_process_title` with a few values (.e.g, `minimal`, `verbose`) to carefully decided what to state in the title

jesperpedersen (changes requested):

fluca1978:
Great! Commit e2b226fe991f43cc2a5596dfab08b7bffb96ef17 does everything.
I'm still wondering if it is the case to introduce a process title setting, just to avoid breaking security environments that do not allow `argv` to be overwritten brutally.
My idea would be:
- `update_process_title = strict` do not exceed `argv` initial length (i.e., as it was before this work)
- `update_process_title = minimal` do the `user/database` process title (i.e., `pgagroal_set_proc_title`)
- `update_process_tutle = verbose` do the `user@host:port/database`

Clearly, it is some amount of working for a transient information...

jesperpedersen:
Yeah, we could do that.

I think having `process_title` is enough with the following values

```
off - Don't touch anything
strict - <= strlen(argv)
minimal - user/database
verbose - user@host:port/database (default)
```

This is for platforms that don't have `setproctitle` system call - like Linux.

jesperpedersen:
Merged.

PR #286, opened 2022-07-01
[#285] Introduce `update_process_title` setting

### LINKED ISSUE #285: Create an `update_process_title` setting
See discussion related to #215 here <https://github.com/agroal/pgagroal/pull/284#issuecomment-1171195925>.
Provide a configuration setting, likely `update_process_title`, that can have values within the range:

```
off - Don't touch anything
strict - <= strlen(argv)
minimal - user/database
verbose - user@host:port/database (default)
```

and call `pgagroal_set_proc_title` and friends depending on such configuration setting.

fluca1978:
This commit adds the `update_process_title` configuration setting.
Values for such setting are:
- `never` or `off`, means the user does not want to update the process
title at all;
- `strict` (used on Linux and systems that do no provide a native call
to update the process title) allows for changing the process title
without overflowing the length of the initial command line;
- `minimal` sets the process title to `user/database` even if this
exceeds the initial command line length;
- `verbose` sets the process title to `user@host:port/database` even
if this overflows the initial command line length.

By default the setting is configured as `verbose`, but this could
break some aggressively secured environments, where no native way to
set the process title is provided.

The `pgagroal_set_proc_title` function has been refactored so that, if
the update policy is set to `never` the process title will never be
updated. This can be confusing because even the "main" process will
not have a title update.
Likely, if the policy is `strict`, the function will never try to
overflow the initial command line length.

On those systems that do provide a native way to set the process
title, there is no difference between `strict` and `minimal` and the
`minmal` policy is always adopted.

Documentation updated.

Close #285

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -66,6 +66,7 @@ static int as_logging_rotation_age(char* str, unsigned int* age);
 static int as_validation(char* str);
 static int as_pipeline(char* str);
 static int as_hugepage(char* str);
+static unsigned int as_update_process_title(char* str, unsigned int* policy, unsigned int default_policy);
 static int extract_value(char* str, int offset, char** value);
 static void extract_hba(char* str, char** type, char** database, char** user, char** address, char** method);
 static void extract_limit(char* str, int server_max, char** database, char** user, int* max_size, int* initial_size, int* min_size);
@@ -140,6 +141,8 @@ pgagroal_init_configuration(void* shm)
    config->allow_unknown_users = true;
 
    atomic_init(&config->su_connection, STATE_FREE);
+
+   config->update_process_title = UPDATE_PROCESS_TITLE_VERBOSE;
 
    return 0;
 }
@@ -568,6 +571,13 @@ pgagroal_read_configuration(void* shm, char* filename, bool emitWarnings)
                   if (as_bool(value, &config->track_prepared_statements))
                   {
                      unknown = true;
+                  }
+               }
+               else if (key_in_section("update_process_title", section, key, true, &unknown))
+               {
+                  if (as_update_process_title(value, &config->update_process_title, UPDATE_PROCESS_TITLE_VERBOSE))
+                  {
+                     unknown = false;
                   }
                }
                else
@@ -2416,6 +2426,9 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
    config->metrics_cache_max_age = reload->metrics_cache_max_age;
    restart_int("metrics_cache_max_size", config->metrics_cache_max_size, reload->metrics_cache_max_size);
    config->management = reload->management;
+
+   config->update_process_title = reload->update_process_title;
+
    /* gracefully */
 
    /* disabled */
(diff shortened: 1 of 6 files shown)


src/libpgagroal/configuration.c
```diff
@@ -570,6 +573,13 @@ pgagroal_read_configuration(void* shm, char* filename, bool emitWarnings)
                      unknown = true;
                   }
                }
+               else if (key_in_section("update_process_title", section, key, true, &unknown))
+               {
+                  if (as_update_process_title(value, &config->update_process_title, UPDATE_PROCESS_TITLE_STRICT))
```
jesperpedersen, line 578: _VERBOSE, right ?

jesperpedersen (changes requested):

src/libpgagroal/configuration.c
jesperpedersen: It sounds like the best solution is to stay on `_STRICT` in this case then.

I'll await your decision :)

fluca1978: After reasoning a bit, I decided to put `_VERBOSE` as the default even when the configuration is not understood. Seems to me coherency is better here.

jesperpedersen:
```
src/libpgagroal/utils.c:826:4: warning: 'snprintf' size argument is too large; destination buffer has size 128, but size argument is 256 [-Wfortify-source]
   snprintf(info, MAX_USERNAME_LENGTH + MISC_LENGTH, "%s@%s:%d",
```

fluca1978:
I need to check, because I did not get such warning in the build (or I missed it).

jesperpedersen:
This was with `clang 14`

fluca1978:
I'm using clang 13.0.1, that seems the most shiny version available on rocky linux 8.6.

In any case, I decided to use a dynamically allocated string to set up all the title information, please check no warnings are issued when building 601759c8c864dad60f3cd9eca33e3d3d9cc2945a

fluca1978:
yeah, I realized this after I pushed.
I need some more time to rethink a decent way to use an array with the correct length.
Probably declaring `char info[ MISC_LENGTH + MAX_USERNAME_LENGTH ]` or something like that.

jesperpedersen:
That, or pass `info` back to `worker.c` are likely the best options

fluca1978:
I took a different approach here, hope it passes without warnings your clang version (no warnings here).

Since `pgagroal_set_proc_title` trims the title at 256 bytes, on Linux systems, I thought it is not worth at all computing a dynamically allocated string in `verbose` mode to get it trimmed anyway. So I introduced a `MAX_PROCESS_TITLE_LENGTH` constant (`256` bytes) and used it as a limit in both functions.

The effect is that the `verbose` title could become incomplete when the username and the database name are really long, but I guess `256` chars for a process title are a lot in any case. On BSD this means that the process title will be longer than `256` but the `verbose` part will be trimmed to `256` in any case.

I think this is a good compromise.

src/libpgagroal/utils.c
```diff
@@ -797,13 +817,17 @@ pgagroal_set_connection_proc_title(int argc, char** argv, struct connection* con
 {
    struct configuration* config;
    int primary;
-   char info[MISC_LENGTH];
+   char info[MAX_PROCESS_TITLE_LENGTH];
```
jesperpedersen, line 820: We should `memset(&info, 0, sizeof(info));` this

src/libpgagroal/utils.c
```diff
@@ -797,13 +817,17 @@ pgagroal_set_connection_proc_title(int argc, char** argv, struct connection* con
...
+   {
+      // cannot find the primary, this is a problem!
+      pgagroal_set_proc_title(argc, argv, connection->username, connection->database);
+      return;
+   }
 
-   pgagroal_get_primary(&primary);
-
-   snprintf(info, MISC_LENGTH, "%s@%s:%d",
+   snprintf(info, MAX_PROCESS_TITLE_LENGTH, "%s@%s:%d",
```
jesperpedersen, line 830: To be sure that the string is `\0` terminated it should be

```
   snprintf(info, MAX_PROCESS_TITLE_LENGTH - 1, "%s@%s:%d",
```

jesperpedersen (changes requested):

src/libpgagroal/utils.c
fluca1978: Shame on me, right.

src/libpgagroal/utils.c
fluca1978: Fixed.

jesperpedersen:
Merged.

PR #289, opened 2022-07-25
[#253] Refactor commands in `pgagroal-cli`

### LINKED ISSUE #253: Refactor command names in `pgagroal-cli`
The `pgagroal-cli` command has, according to me, some inconsistency within the commands.
For example:
- `gracefully` does a graceful shutdown;
- `cancel-shutdown` does cancel the above request (but no "gracefully" appears in the name)
- `stop` does a not graceful halt of the system.

I propose to choose a common word, e.g., `shutdown` and add a subcommand to specify the shutdown method:
- `shutdown` will result to be the same as `gracefully`
- `shutdown immediately` will result the same as `stop`
- `cancel-shutdown` will remain the same and now it is clear which commands are related to the shutdown.

Same thing for the `flush-xxx` commands that can become `flush` with a subcommand related to the mode, e.g., `flush gracefully`.

The `reset` command is tied to the Prometheus info, and `reset-server` is the one that does reset a configured server. Appears to me it is better to have the `reset` command to be tied to the working of pgagroal, not of Prometheus, because it is more important and the metrics could be an optional feature (meaning it is not enabled).

Old commands should remain, to avoid breaking working clients and batches, and should report on `stderr` that the command has been deprecated.

fluca1978:
Now `pgagroal-cli` has a set of "logically" grouped commands and
subcommands. For example, all the commands related to shutting down
the pooler are under the `shutdown` command, that can operate with
subcommands like `gracefully`, `immediate` or `cancel`.

In order to provide this capability, two new functions have been
introduced:
- `parse_command()` accepts the command line and seek for a command,
possibly its subcommand, and an optional "value" (often the database
or server name).
- `parse_deprecated_command()` does pretty much the same thing but
against the old command. Thanks to this, old commands can still work
and the user will be warned about their deprecation, but the interface
of `pgagroal-cli` is not broken.

Both functions requires to know the offset at which start seeking for
a command, and that depends on the number of options already parsed
via `getopt_long()`. Since the `&option_index` is valued only for long
options, I decided to use the `optind` global value, see
getopt_long(3).
This value is initialized with the "next thing" to seek on the command
line, i.e., the next index on `argv`.

In the case the command accepts an optional database name, the
database value is automatically set to '*' (all databases) in case the
database name is not found on the command line.
Therefore:
   pgagroal-cli flush idle
is equivalent to
   pgagroal-cli flush idle '*'

On the other hand, commands that require a server name get the value
automatically set to "\0" (an invalid server name) in order to "block"
other pieces of code. Moroever, if the server has not been specified,
the command is automatically set to "unknown" so that the help screen
is shown.

The `pgagroal-cli` has a set of `pgagroal_log_trace()` calls whenever
a command is "parsed", so that it is possible to quickly follow the
command line parsing.

Also, since the `pgagroal-cli` exists if no command line arguments
have been specified, the safety check aboutt `argc > 0` around the
command line parsing have been removed.

In the case the user specified an unknown command, she is warned on
stdout before printing the `usage()` help screen.

Deprecated commands are notified to the user via a warning message,
printed on stderr, that provides some hints about the correct usage of
the new command.

Documentation has been updated and improved.
A specific documentation section about deprecated
commands have been introduced.

Close #253

### DIFF
--- a/contrib/shell_comp/pgagroal_comp.bash
+++ b/contrib/shell_comp/pgagroal_comp.bash
@@ -2,16 +2,37 @@
 
 # COMP_WORDS contains
 # at index 0 the executable name (pgagroal-cli)
-# at index 1 the command name (e.g., flush-all)
+# at index 1 the command name (e.g., flush)
+# at index 2, if required, the subcommand name (e.g., all)
 pgagroal_cli_completions()
 {
 
     if [ "${#COMP_WORDS[@]}" == "2" ]; then
         # main completion: the user has specified nothing at all
         # or a single word, that is a command
-        COMPREPLY=($(compgen -W "flush-idle flush-gracefully flush-all is-alive enable disable stop gracefully status details switch-to reload reset reset-server config-get config-set" "${COMP_WORDS[1]}"))
+        COMPREPLY=($(compgen -W "flush is-alive enable disable shutdown status details switch-to conf clear" "${COMP_WORDS[1]}"))
+    else
+        # the user has specified something else
+        # subcommand required?
+        case ${COMP_WORDS[1]} in
+            flush)
+                COMPREPLY+=($(compgen -W "gracefully idle all" "${COMP_WORDS[2]}"))
+                ;;
+            shutdown)
+                COMPREPLY+=($(compgen -W "gracefully immediate cancel" "${COMP_WORDS[2]}"))
+                ;;
+            clear)
+                COMPREPLY+=($(compgen -W "server prometheus" "${COMP_WORDS[2]}"))
+                ;;
+	    conf)
+		COMPREPLY+=($(compgen -W "reload get set" "${COMP_WORDS[2]}"))
+		;;
+        esac
     fi
+
+
 }
+
 
 
 pgagroal_admin_completions()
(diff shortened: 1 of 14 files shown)


fluca1978:
In the case this is a good implementation, we could refactor similarly also `pgagroal-admin` too.

fluca1978:
> We should keep the `reset` command for now

The problem with having `reset` is that we will not be able to "catch" the usage of the deprecated command in place of the new one. That's why I changed the word into `clear`. Any suggestion here is welcome.

>, and I don't think we should display the deprecated message in the 1.x branches.

I can add a check into `parse_deprecated_command` to emit the warning only if `pgagroal` version is greater than a constant.


> The rest looks good, and changing `pgagroal-admin` as well would be nice.

I will open an issue about that and start working on it, but I need to move the utility functions into a shared file (i.e, making them not static), therefore that is going to influence also this implementation.

> Another thing could be shell completion based on the new commands.

Yeah, but I've no idea at the moment on how to do that.

jesperpedersen:
Yes, we can use your `VERSION` pull request for this; set to `PGAGRAOL_VERSION_MAJOR` to 2

fluca1978:
I think I've a complete piece of code, but I'm postponing this when the new version will be released.

jesperpedersen:
Lets hold this for 1.6

fluca1978:
Commit 53884cf88f1faab5ea2acc70da9126424ee5bc15 is another attempt at get the command refactoring done.
Please take into account that this changes the documentation, so we should keep it separated by the currently available documentation on the repository by mean of a dedicated branch (e.g., `development`).

jesperpedersen:
Lets promote the new CLI commands, but allow the old ones.

Since we are not in 2.x mode; there should not be a warning using the "old" commands.

Could you rebase ?

fluca1978:
I've rebased so far, but at this point I would wait for #329 to be implemented (and I don't know when it will happen, since it is quite challenging).

With regard to the warning, right now it is set to be placed when the version is less than 1.6 (not yet released), we can pump it to 2 so that the warning is already there but not shown before version 2 is released.

However, one problem I see is that the documentation changes accordingly, so we risk to have an online documentation that does not reflect the current stable release.

jesperpedersen:
Ok, lets wait for `config-set`.

I think we should change the documentation to the new commands - the old commands will be in place for existing installations.

fluca1978:
@jesperpedersen can you please give ade40240317bad155dbf1e40866c96257b688b90 a try? I think this is complete now.

jesperpedersen:
Cool, I'll give this a review next week

jesperpedersen:
Merged.

We can consider to just remove the deprecated commands... but that is a discussion in https://github.com/agroal/pgagroal/discussions/categories/ideas

Issue #290, opened 2022-07-27
Refactor `pgagroal-admin` commands


fluca1978:
Related to #253 and discussion in #289 (see <https://github.com/agroal/pgagroal/pull/289#issuecomment-1196884782>): `pgagroal-admin` could have a more consistent set of commands grouped by *command* and *subcommand*.
An example could be:  `user` and then `add` or `mod` or `del` In the previous example `user` is the command and the other three are the subcommands.

It should also provide a set of warning for deprecated commands only when `pgagroal` is greater than a specific version number,

jesperpedersen:
:+1:

PR #293, opened 2022-07-28
[#292] Provide application version as a numeric value too

### LINKED ISSUE #292: Make the version number comparable
In `pgagroal.h` the version of the pooler is defined as a constant string:

```
#define VERSION "1.5.0"
```

this makes hard to compare version numbers within the program. I propose to split the version number into three numeric parts, providing an utility function to get the string-yfied version of the version number to display.

This is useful also to decide which part of code are deprecated or not depending on the version, as per discussion and needs in #289, #290, and #253.

fluca1978:
This commit introduces two compile time macros:
- PGAGROAL_VERSION is a string and contains the application version
that will be shown in various application messages.
- PGAGROAL_VERSION_NUM is a numeric value that can be used to compare
the running version with another number, so for example to limit
features or messages.

`PGAGROAL_VERSION` replaces the `VERSION` macro that was defined in
the main header file.

Moreover, the message printed at level INFO when booting the pooler
has been enhanced to show also the version that is starting, like for
instance:
    pgagroal: v1.5.0 started on *:5432

See #253, #289, #290

Close #292

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -69,6 +69,15 @@ add_compile_options(-D_GNU_SOURCE)
 add_compile_options(-Wno-deprecated)
 add_compile_options(-Wno-deprecated-declarations)
 
+#
+# version number and string representation
+# the numeric value is computed as
+#   patch + minor * 1000 + major * 10000
+#
+add_compile_options(-DPGAGROAL_VERSION="1.5.0")
+add_compile_options(-DPGAGROAL_VERSION_NUM=15000)
+
+
 set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined")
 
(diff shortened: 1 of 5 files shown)


jesperpedersen:
Lets do

```
PGAGROAL_VERSION_MAJOR 1
PGAGROAL_VERSION_MINOR  5
PGAGROAL_VERSION_PATCH  0
```

instead.

fluca1978:
I've seen that in `CmakeLists.txt` there is already all the stuff we need: <https://github.com/agroal/pgagroal/blob/master/CMakeLists.txt#L6>. Cannot we declare that as compile time macros and use them within the program instead of defining *again* them in the source files?

By the way, I've already all the code for using three constants and functions to convert them into a unique number and string, so it just a matter of taste.

jesperpedersen:
Yeah, we can use the constants from `CMakeLists.txt`

fluca1978:
Work discarded, see #294.

Issue #291, opened 2022-07-27
Command completion for main shells


fluca1978:
Provide shell autocompletion of commands for at least `bash` and `zsh`, and possibly other shells.
See #290 and #253, it probably does make sense to provide completion once those have been completed.

jesperpedersen:
:+1:

PR #294, opened 2022-07-28
[#291] Make the version number comparable

### LINKED ISSUE #292: Make the version number comparable

fluca1978:
Three new constants are introduced here:
- PGAGROAL_MAJOR_VERSION (number)
- PGAGROAL_MINOR_VERSION (number)
- PGAGROAL_PATCH_VERSION (number)
- PGAGROAL_VERSION       (string)

that represent, as integer literals and string composition,
the parts of the version number.
The values are copied from the variables in the `CmakeLists.txt` file
that creates the building flow, so changing the values in a single
place changes and propagates the changes all over the source code.

A few utility functions have been introduced to compose the three
constant into a unique number that can be used to compare the version
currently running, so that features and/or message can be targeted as
conditionals.

Introduced functions are:
- pgagroal_version_as_number() returns an unique sortable version
number. For example, version 1.5.0 is returned as 15000, version 1.6.2
as 16002, and so on;
- pgagroal_version_number() returns the currently running version
number, that is invokes pgagroal_version_as_number() with the
predefined constants.

The `pgagroal_version_as_number()` accepts individual values, while
`pgagroal_version_number()` provides the currently running version
number. The idea is that, thanks to both the functions, it is possible
to check for a feature that depends on a version in a way like the
following one:

   if (pgagroal_version_numer() >=
             pgagroal_version_as_number(2,1,0))
       // version is greater than 2.1.0 !

The utility function `pgagroal_version_ge()` does exactly the above:
it accepts the three parts of a version number and checks if the
currently running version is greater or equal (hence, 'ge') of the
specified triplet, so that the above piece of code becomes:

   if (pgagroal_version_ge(2,1,0))
     // version greater or equal 2.1.0

Close #292
See #253, #289, #290, #293

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -69,6 +69,14 @@ add_compile_options(-D_GNU_SOURCE)
 add_compile_options(-Wno-deprecated)
 add_compile_options(-Wno-deprecated-declarations)
 
+#
+# version number and string management
+#
+add_compile_options(-DPGAGROAL_MAJOR_VERSION=${VERSION_MAJOR})
+add_compile_options(-DPGAGROAL_MINOR_VERSION=${VERSION_MINOR})
+add_compile_options(-DPGAGROAL_PATCH_VERSION=${VERSION_PATCH})
+add_compile_options(-DPGAGROAL_VERSION="${VERSION_STRING}")
+
 set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined")
 
 if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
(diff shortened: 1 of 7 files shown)


fluca1978:
I force pushed the same commit with the correct refence to the issue, that is *292*, not *291*.

jesperpedersen:
Nice !

jesperpedersen:
Merged.

PR #300, opened 2022-08-11
Improve pid file management

fluca1978:
These commits improve a little the management of the pid file by `main` application.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -851,7 +851,7 @@ read_superuser_path:
 #ifdef HAVE_LINUX
       sd_notifyf(0, "STATUS=Could not bind to %s/%s", config->unix_socket_dir, MAIN_UDS);
 #endif
-      exit(1);
+      goto error;
    }
 
    if (!has_unix_socket)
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -1954,17 +1958,20 @@ create_pidfile(void)
 
    if (strlen(config->pidfile) > 0)
    {
-      // check pidfile is not there
-      if (access(config->pidfile, F_OK) == 0)
+      pid = getpid();
+
+      fd = open(config->pidfile, O_WRONLY | O_CREAT | O_EXCL, 0644);
```
jesperpedersen, line 1963: Can you change this to `0640` ?

jesperpedersen (changes requested):

jesperpedersen:
Just squash the commits with both issues referenced.

src/main.c
fluca1978: sure

fluca1978:
> Just squash the commits with both issues referenced.

ok, but I like more to keep commits separated...

jesperpedersen:
Could you split it into separate PRs then ? One for each issue - otherwise I can do it as part of the merge.

fluca1978:
> Could you split it into separate PRs then ? One for each issue - otherwise I can do it as part of the merge.

I don't know how to split commits once I've rebased them and force pushed. I guess this is fine as a single PR.

My personal idea is to keep related commits into a PR/branch, and commits as short as possible to avoid "jumbo patches".

jesperpedersen:
Merged.

PR #304, opened 2022-08-11
[#303] Allows for 'B' suffix in multiletter size specification.

### LINKED ISSUE #303: Allows for two letter size specification ending with 'b'
The function `as_bytes` accepts suffixes like 'M' for megabytes.
It would be nice and more natural to allow the user to specify also 'MB', that means allow for a double letter suffix only when the last letter is a 'B'.

fluca1978:
Allows the `as_bytes` function to parse also compound suffixes like
`MB` for megabytes.

Documentation updated.

Close #303


jesperpedersen:
Merged.

PR #305, opened 2022-08-11
[#296][#297] Improve configuration error management

### LINKED ISSUE #296: Better checks for `pgagroal.conf`
This happens on current HEAD 1296cc4216c73119a1ff4c3a3ffd0c610ca04f69; assume the following dumb configuration file `pgagroal.conf`:

```
$ cat /etc/pgagroal/pgagroal.conf
[pgagroal]
host = *
port = 5432
unix_socket_dir = /tmp

management=6432

log_type  = console
log_level = info
log_path  = /var/log/pgagroal/pgagroal.log
log_mode  = create

log_connections    = on
log_disconnections = on
log_rotation_size = "2M"# in megabytes
log_rotation_age = 1m# one minute
log_line_prefix  = "PGAGROAL #%Y-%m-%d-%H:%M:%S" # the prefix of the log

max_connections = 4

pipeline = 'performance'
# solo per la pipeline transaction
allow_unknown_users = false
blocking_timeout = 10000

metrics = 8000
metrics_cache_max_age = 10s
metrics_cache_max_size = 20k

#idle_timeout = 0

update_process_title = minimal


[venkman]
host = venkman
port = 5432
#primary = on


[pgagroal]
log_level = debug

[venkman]
host = miguel
port = 5432

[a]
host = venkman
port = 5432

[b]
host = venkman
port = 5432
```

There are a few mistakes by purpose in it:
1) the `pgagroal` section has been defined twice;
2) the `venkman` section has been defined twice;
3) server `a` and `b` point to the very same endpoint.

As a result:

```
$ pgagroal-cli details
Status:              Running
Active connections:  0
Total connections:   0
Max connections:     4
---------------------
Server:              venkman
Host:                venkman
Port:                5432
State:               Not init
---------------------
Server:              venkman
Host:                miguel
Port:                5432
State:               Not init
---------------------
Server:              a
Host:                venkman
Port:                5432
State:               Not init
---------------------
Server:              b
Host:                venkman
Port:                5432
State:               Not init
---------------------
Database:            pgbench
Username:            pgbench
Active connections:  0
Max connections:     2
Initial connections: 1
Min connections:     1
---------------------
```

and the `log_level` last defined in the repetition of `pgagroal` section is applied.

I do propose to change the configuration parsing so that:
- a section can be seen only once, this fixes 1 and 2
- the same endpoint cannot be used multiple times with different names, this fixes 3. The rationale is that this can lead to dumb behaviours when failover is enabled. 

With regard to the latter, at least produce a warning when the same endpoint is found in the configuration, so that the user can be warned properly, in particular when the failover is active.

### LINKED ISSUE #297: Reduce the number of error paths in configuration parsing
While working on #296 I saw that `configuration.c` has a few functions that have error paths like <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L1291>:

```
error:

   free(master_key);
   free(password);
   free(decoded);

   if (file)
   {
      fclose(file);
   }

   return 1;

masterkey:

   free(master_key);
   free(password);
   free(decoded);

   if (file)
   {
      fclose(file);
   }

   return 2;

above:

   free(master_key);
   free(password);
   free(decoded);

   if (file)
   {
      fclose(file);
   }

   return 3;
```

The above error paths are identical except for the return value.
Since it is safe to call `free(3)` on a null pointer, as it happens to be, and since the only change is on the return value, I propose to keep track of the return value and merge all the error paths into a single one to improve code readibility.

fluca1978:
This commit introduces a check for `pgagroal.conf` to avoid duplicated
sections.
A new structure, named `struct config_section` has been introduced to
handle the parsing of every section in the configuration file. The aim
of the struct is to temporarily store the line number and the name of
the section as the parsing goes on.
At the end of parsing, all the found sections are compared to see if
there are duplicates.

The function `pgagroal_read_configuration()` has been refactored so
that it returns a negative value in the case of file problem (e.g.,
file not found or too many sections within the file), or a positive
value to indicate how many duplicated sections have been found.

Several constants have been defined:
- PGAGROAL_CONFIGURATION_STATUS_FILE_TOO_BIG if the file contains
too much sections or users or entries, (e.g., for sections if it
overflows  NUMBER_OF_SERVERS + 1);
- PGAGROAL_CONFIGURATION_STATUS_FILE_NOT_FOUND to indicate that the
file does not exists;
- PGAGROAL_CONFIGURATION_STATUS_CANNOT_DECRYPT to indicate a mastre
key problem;
- PGAGROAL_CONFIGURATION_STATUS_KO to indicate a generic error;
- PGAGROAL_CONFIGURATION_STATUS_OK to indicated everything is fine.

The `main` function has been refactored too to adapt to the new return
values and to provide better descriptive messages to the user. In
particular a new temporary variable `message` is now used to format a
coherent error message between the application output and the systemd
notification.

Documentation updated to better explain limits on the number of
sections. Also, place brief documentation to explain every
configuration file limits.

Moreover, when validating the main configuration, there is a nested
loop to check against duplicated servers, i.e., servers that have the
very same endpoint (host and port). The check relies on
`is_same_server' and therefore it has all the limitation that the
latter has, most notably it does not performs a DNS lookup, so at the
moment the usage of same servers with different hostnames cannot be
detected.
The detection of duplicated server is especially useful to allow
`switch-to` and the failover to not perform useless operations.

Since now there are constants to define the return values for
different error conditions in the configuration parsing, the `main`
and the `pgagroal-cli` have been refactored to use such return values
in a more readable and useful way.

Documentation of the functions have been updated.

Functions that read the configuration have different return values
depending on the particular error that occured. However, the error
paths are all alike, and it changes only the return value.

Since calling `free(3)` on a NULL pointer has no effect at all, it is
safe to merge all the error paths in a single one and keep track only
of the return value to emit.

This improves code readability.

Close #296.
Close #297

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -7,11 +7,16 @@ The configuration of `pgagroal` is split into sections using the `[` and `]` cha
 The main section, called `[pgagroal]`, is where you configure the overall properties
 of the connection pool.
 
-Other sections doesn't have any requirements to their naming so you can give them
+The other sections configure each one server. There can be no more than `64` server sections, therefore
+no more than `64` servers configured as backends.
+Server sections don't have any requirements to their naming so you can give them
 meaningful names like `[primary]` for the primary [PostgreSQL](https://www.postgresql.org)
 instance.
 
-All properties are in the format `key = value`.
+In any case, it is not possible to have duplicated sections, that means section names must be unique within
+the configuration file.
+
+All properties within a section are in the format `key = value`.
 
 The characters `#` and `;` can be used for comments. A line is totally ignored if the
 very first non-space character is a comment one, but it is possible to put a comment at the end of a line.
@@ -32,6 +37,9 @@ pipeline = 'performance' # no need to quote since it does not contain any spaces
 See a more complete [sample](./etc/pgagroal.conf) configuration for running `pgagroal` on `localhost`.
 
 ## [pgagroal]
+
+This section is mandatory and the pooler will refuse to start if the configuration file does not specify one and opnly one. Usually this section is place on top of the configuration file, but its position within the file does not really matter.
+The available keys and their accepted values are reported in the table below.
 
 | Property | Default | Unit | Required | Description |
 |----------|---------|------|----------|-------------|
@@ -88,6 +96,9 @@ __Danger zone__
 
 ## Server section
 
+Each section with a name different from `pgagroal` will be treated as an host section.
+There can be up to `64` host sections, each with an unique name and different combination of `host` and `port` settings, otherwise the pooler will complain about duplicated server configuration.
+
 | Property | Default | Unit | Required | Description |
 |----------|---------|------|----------|-------------|
 | host | | String | Yes | The address of the PostgreSQL instance |
@@ -122,6 +133,8 @@ host    all      all   all      all
 
 Remote management users needs to have their database set to `admin` in order for the entry to be considered.
 
+There could be up to `64` HBA entries in the configuration file.
+
 # pgagroal_databases configuration
 
 The `pgagroal_databases` configuration defines limits for a database or a user or both. The limits are the number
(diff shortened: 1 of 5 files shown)


jesperpedersen:
Merged.

PR #307, opened 2022-08-12
[#306] Remove `atoi()` in `as_logging_level()`

### LINKED ISSUE #306: Improve function `as_logging_level()()`
In function `as_logging_level()` there is usage of `atoi()` which is not ideal. Since we already have `as_int()` that catches parsing errors and overflows, we should remove `atoi()` in favor of `as_int()`.

fluca1978:
Use the internal function `as_int()` that checks for overflows and not
parsable strings.

Close #306


jesperpedersen:
Merged.

PR #314, opened 2022-08-23
[#313] Use '%lu' format specifier for file limit

### LINKED ISSUE #313: use `unsigned long int` formatter in error message about file limit
At line <https://github.com/agroal/pgagroal/blob/master/src/main.c#L862> and following systemd one, the `%ld` format specifier is used for `rlimit_cur`, that is of type ` long unsigned int` (thus `%lu`).

fluca1978:
On Linux, `rlimit_t` is of type `unsigned long int`, and therefore the
appropriate format specifier to use is `%lu`.

Close #313


jesperpedersen:
Merged.

PR #317, opened 2022-08-24
[#316] Use `err(3)`

### LINKED ISSUE #316: use `err(3)` functions
There is a lot of code that contains `printf` or `fprintf(stderr)` and `exit` (e.g., <https://github.com/agroal/pgagroal/blob/master/src/main.c#L528>).

I propose to change such bunch of code with the `err(3)` set of functions, like `errx`. Advantages:
- it will bring coherency around the code
- makes the code a little more readable
- no need for the extra new line at the end of the string
- provides automatically short application name in front of the message, as for example
```
$ pgagroal
pgagroal: Server section [a] duplicated at lines 15 and 19 of file </etc/pgagroal/pgagroal.conf>
pgagroal: 1 problematic or duplicated sections (file </etc/pgagroal/pgagroal.conf>)

$ /usr/local/bin/pgagroal
pgagroal: Server section [a] duplicated at lines 15 and 19 of file </etc/pgagroal/pgagroal.conf>
pgagroal: 1 problematic or duplicated sections (file </etc/pgagroal/pgagroal.conf>)
```

The `pgagroal:` part is automatically appended even if the application is invoked with a full path, so we can remove it (and forget about it) from every message.

fluca1978:
Whenever in the code there is a `printf` + `exit` or similar piece of
code, use `err(3)` functions instead.
This brings coherency, removes the need for the extra new line, the
program name and provides automatic expansion of errors when needed.

This also changes the `create_pidfile` static function so that it
terminates the current program (main) if the pid file cannot be
created. To emphasize this, the function has been renamed to
`create_pidfile_or_exit`. Also the `remove_pidfile` changed to emit a
warning if the pid file cannot be removed

Close #316

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -43,6 +43,7 @@
 #include <unistd.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <err.h>
 
 #define DEFAULT_PASSWORD_LENGTH 64
 #define MIN_PASSWORD_LENGTH 8
@@ -169,8 +170,7 @@ main(int argc, char** argv)
 
    if (getuid() == 0)
    {
-      printf("pgagroal: Using the root account is not allowed\n");
-      exit(1);
+      errx(1, "Using the root account is not allowed");
    }
 
    if (argc > 0)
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Merged.

Thanks for the contribution !

PR #320, opened 2022-08-25
[#319] Remove `atoi` from `pgagroal-cli`

### LINKED ISSUE #319: Remove `atoi` from `pgagroal-cli`
In case of remote connection, `pgagroal-cli` uses `atoi` to get the port number <https://github.com/agroal/pgagroal/blob/master/src/cli.c#L451>.
This should be removed in favor of `strtol` or similar.

fluca1978:
When doing a remote connection to the pooler, the user has to specify
the port with `-p` flag. The value is converted with `strtol` and the
ranges are checked.
Also, since pgagroal cannot be run as root, it cannot bind to ports
lower than 1024, therefore if the user specifies a wrong port number,
it gets a different error message rather than "No route to host".

Close #319

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -46,6 +46,7 @@
 #include <sys/mman.h>
 #include <sys/types.h>
 #include <err.h>
+#include <errno.h>
 
 #include <openssl/ssl.h>
 
@@ -153,6 +154,7 @@ main(int argc, char** argv)
    char* server = NULL;
    struct configuration* config = NULL;
    bool remote_connection = false;
+   long l_port;
 
    while (1)
    {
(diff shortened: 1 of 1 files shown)


src/cli.c
```diff
@@ -448,9 +450,27 @@ main(int argc, char** argv)
          else
          {
             /* Remote connection */
-            if (pgagroal_connect(host, atoi(port), &socket))
+
+            l_port = strtol(port, NULL, 10);
+            if ((errno == ERANGE
+                 && (l_port == LONG_MAX || l_port == LONG_MIN))
+                || (errno != 0 && l_port == 0))
```
jesperpedersen, line 457: Could you put `&&` and `||` on the previous lines ?

src/cli.c
```diff
@@ -448,9 +450,27 @@ main(int argc, char** argv)
...
+            {
+               warnx("Specified port %s out of range", port);
+               goto done;
+            }
+
+            // cannot connect to port less than 1024 becayse pgagroal
```
jesperpedersen, line 463: Spelling

jesperpedersen (changes requested):

src/cli.c
fluca1978: fixed

src/cli.c
fluca1978: fixed

jesperpedersen:
Merged.

Thanks for the contribution !

PR #323, opened 2022-08-25
[#322] OpenBSD and FreeBSD configuration files

### LINKED ISSUE #322: Path for configuration files in BSD systems
This is related to f5058d0 and #315 : on OpenBSD configuration files should be kept in `/etc`, not `/usr/local/etc`.

fluca1978:
Change the cmake configuration to use `/usr/local/etc` as the folder
for the configuration files only when the system detected is FreeBSD.

Also change the macros in the source header to use `/usr/local/etc`
only when FreeBSD is detected.

Last, this installs configuration files only if they are not already
present at the time of build.

Close #322

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -66,8 +66,8 @@ add_custom_command(
 #
 # Install configuration and documentation
 #
-install(FILES etc/pgagroal.conf DESTINATION etc/pgagroal)
-install(FILES etc/pgagroal_hba.conf DESTINATION etc/pgagroal)
+install(FILES etc/pgagroal.conf DESTINATION share/doc/pgagroal/etc)
+install(FILES etc/pgagroal_hba.conf DESTINATION share/doc/pgagroal/etc)
 
 install(DIRECTORY . DESTINATION share/doc/pgagroal FILES_MATCHING PATTERN "*.md" PATTERN "etc" EXCLUDE PATTERN "images" EXCLUDE PATTERN "man" EXCLUDE)
 install(DIRECTORY images/ DESTINATION share/doc/pgagroal/images FILES_MATCHING PATTERN "*.png")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -44,6 +44,8 @@ else()
 
   if (${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
     add_compile_options(-DHAVE_OPENBSD)
+  elseif (${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD")
+     add_compile_options(-DHAVE_FREEBSD)
   endif()
 
   #
--- a/src/include/pgagroal.h
+++ b/src/include/pgagroal.h
@@ -49,10 +49,10 @@ extern "C" {
 
 #define MAIN_UDS ".s.pgagroal"
 
-#ifdef HAVE_LINUX
-    #define PGAGROAL_DEFAULT_CONFIGURATION_PATH "/etc/pgagroal/"
+#ifdef HAVE_FREEBSD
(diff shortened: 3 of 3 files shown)


fluca1978:
Probably this has to be improved, because there must be an easier and smater way to check if the files are there before overriding the configuration.
Note that this was a problem also in Linux: I never realized that a `make install` was installing configuration files under `/usr/local/etc` every time without nuking my `/etc/pgagroal` configuration.

jesperpedersen:
It is probably better to put them in `share/doc/pgagroal/etc` always.

`pgagroal.spec` should reflect that as well.

fluca1978:
Uhm, then I need assistance here to get the job done.
If you know what to do, steal the issue!

jesperpedersen:
You change

```
install(FILES etc/pgagroal.conf DESTINATION etc/pgagroal)
```

to

```
install(FILES etc/pgagroal.conf DESTINATION share/doc/pgagroal/etc/)
```

And, actually the `.spec` file is ok in either case...

fluca1978:
Done.
This however means the user will not be able to start `pgagroal` until she creates the configuration files by hand. Also I think we should rename the files as `.conf.sample`?

src/include/pgagroal.h
```diff
@@ -49,10 +49,10 @@ extern "C" {
 
 #define MAIN_UDS ".s.pgagroal"
 
-#ifdef HAVE_LINUX
-    #define PGAGROAL_DEFAULT_CONFIGURATION_PATH "/etc/pgagroal/"
+#ifdef HAVE_FREEBSD
+  #define PGAGROAL_DEFAULT_CONFIGURATION_PATH "/usr/local/etc/pgagroal/"
```
jesperpedersen, line 53: Spacing

src/include/pgagroal.h
```diff
@@ -49,10 +49,10 @@ extern "C" {
...
 #else
-   #define PGAGROAL_DEFAULT_CONFIGURATION_PATH "/usr/local/etc/pgagroal/"
+  #define PGAGROAL_DEFAULT_CONFIGURATION_PATH "/etc/pgagroal/"
```
jesperpedersen, line 55: Spacing

jesperpedersen (changes requested):

jesperpedersen:
I think that is ok, since `pgagroal` and PostgreSQL won't run on the same server anyway. The `GETTING_STARTED` configuration can be used as a template.

Since the sample configuration is the `doc` directory I don't think we need the `.sample` extension

doc/CMakeLists.txt
```diff
@@ -66,8 +66,9 @@ add_custom_command(
 #
 # Install configuration and documentation
 #
-install(FILES etc/pgagroal.conf DESTINATION etc/pgagroal)
-install(FILES etc/pgagroal_hba.conf DESTINATION etc/pgagroal)
+install(FILES etc/pgagroal.conf DESTINATION share/doc/pgagroal/etc)
+install(FILES etc/pgagroal_hba.conf DESTINATION share/doc/pgagroal/etc)
+
```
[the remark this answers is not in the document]
fluca1978: fixed

src/CMakeLists.txt
```diff
@@ -44,6 +44,10 @@ else()
 
   if (${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
     add_compile_options(-DHAVE_OPENBSD)
+  else()
```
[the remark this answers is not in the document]
fluca1978: fixed

src/include/pgagroal.h
fluca1978: sorry, I thought uncrustify was catching this

src/include/pgagroal.h
fluca1978: I've used the same indentation level as before, and removed extra spaces.

jesperpedersen:
Merged.

PR #331, opened 2022-11-24
[#330] Fix exit code handling in `pgagroal_cli`

### LINKED ISSUE #330: Potentially wrong exit status in some `pgagroal-cli` operations
While working on #238 against HEAD, I saw that <https://github.com/agroal/pgagroal/blob/master/src/cli.c#L726> the exit status of the reading over the socket are ignored, and always a success is returned.
While it could be very difficult that a problem arises, in that case the `pgagroal-cli` will report a wrong exit status (0 = success).
Same happens here <https://github.com/agroal/pgagroal/blob/master/src/cli.c#L711>.

fluca1978:
This commit fixes some minor socket-driven communications where the return code was not propagated correctly to the application exit code, and therefore some erros could get a wrong exit code marked as "success".

Meanwhile, little changes the error messages for a failure exit code, since a failure does not automatically imply a "no connection" status, rather could be a "connection problem" (i.e., some data has been sent/received over the socket, some did not).

Last, reduce the code to print the status of the command when using the `--verbose` flag. Use `warnx(3)` to print the verbose exit message, so that it can be redirected to another place since it is printed on STDERR.

Close #330

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -610,7 +610,7 @@ done:
    {
       if (action != ACTION_UNKNOWN && exit_code != 0)
       {
-         printf("No connection to pgagroal on %s\n", config->unix_socket_dir);
+         printf("Connection error on %s\n", config->unix_socket_dir);
       }
    }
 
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #332, opened 2022-11-24
Link to libatomic on Linux

df7cb:
On some architectures, atomic_load() needs -latomic to work, notably armel, mipsel, powerpc, riscv64.

https://buildd.debian.org/status/logs.php?pkg=pgagroal&ver=1.5.0-1


jesperpedersen:
Is there a test we can do ? Or, are we forced to include it on all Linux platforms ?

jesperpedersen:
Done in #347

PR #333, opened 2022-11-28
[#328] Introduce subcommand `config-get`.

### LINKED ISSUE #328: Obtain runtime configuration via `pgagroal-cli`
It would be possible to add another command to `pgagroal-cli`, named for instance *`config-get`* suitable to obtain the runtime configuration for a specific setting.
The usage should be like the following:

```
$ pgagroal-cli config-get pipeline
auto
```

Given a name of a configuration parameter, the command should print out the value of such parameter. This will allow an automated or scriptable way to query a pgagroal instance about its configuration.

Since `pgagroal-cli` already talks via socket to the main process, this will also make it easier to obtain runtime information about the configuration remotely.

fluca1978:
This commit introduces the `pgagroal-cli` subcommand `config-get` that allows the user to query the running `pgagroal` instance and obtain the value of a configuration parameter.

A new management action, named ACTION_CONFIG_GET, is added to manage the communication over the socket between `pgagroal-cli` and the main `pgagroal` process.
A few utility functions to handle the communication have been introduced to send data over the socket and receive it on the socket.

Moreover, different utility functions to print out the value of a configuration setting have been introduced, so that for instance all the boolean values are always mapped to "on" and "off".

The `pgagroal-cli config-get` command accepts a key to search. Usually the key is searched into the global configuration namespace, that is the section `[pgagroal]`. However, it is possible to define other keys that can be handled:
- `server.<server-name>.<what>` will search 'what' within the server section named `[server-name]`, as an example `server.venkman.host`;
- `limit.<database>.<what>` will search 'what' within the pgagroal_databases.conf file entries (runtime) for the specified database name;
- `hba.<user>.<what>` will search 'what' among the runtime HBA configuration for the user 'user'.

Note that in the case of 'hba.' and 'limit.', since the username or database name can appear multiple times, only the first occurency is matched. The match happens in a simple way, so for instance the database name 'foo' does not match an entry with 'all'.

It is also possible to specify a key starting with `pgagroal.`, that will instrument the search within the global namespace. Therefore `max_entries` and `pgagroal.max_entries` are the same search terms.

If a key is not found, the main `pgagroal` process issue a warning in the logs, while the application `pgagroal-cli` does receive an empty answer and therefore does nothing.
Even in the case of a non-existing key, `pgagroal-cli` exists normally, to indicate that the query has been performed well, simply with no meaningful answer.

If `pgagroal-cli` is running with `--verbose` flag, the application prints the configuration entry with the key and the value, like in

    pgagroal-cli get-config max_entries
    max_entries = 300

this is useful for humans, but not for automated systems, so the default way of answering a query is by printing only the value

    pgagroal-cli get-config max_entries
    300

Documentation updated.
Shell completions updated.

Close #328

### DIFF
--- a/contrib/shell_comp/pgagroal_comp.bash
+++ b/contrib/shell_comp/pgagroal_comp.bash
@@ -9,7 +9,7 @@ pgagroal_cli_completions()
     if [ "${#COMP_WORDS[@]}" == "2" ]; then
         # main completion: the user has specified nothing at all
         # or a single word, that is a command
-        COMPREPLY=($(compgen -W "flush-idle flush-gracefully flush-all is-alive enable disable stop gracefully status details switch-to reload reset reset-server" "${COMP_WORDS[1]}"))
+        COMPREPLY=($(compgen -W "flush-idle flush-gracefully flush-all is-alive enable disable stop gracefully status details switch-to reload reset reset-server config-get" "${COMP_WORDS[1]}"))
     fi
 }
 
--- a/contrib/shell_comp/pgagroal_comp.zsh
+++ b/contrib/shell_comp/pgagroal_comp.zsh
@@ -6,7 +6,7 @@ function _pgagroal_cli()
 {
     local line
     _arguments -C \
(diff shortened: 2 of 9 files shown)


jesperpedersen:
Rebased, and merged.

Issue #337, opened 2022-12-14
Fix compilation problems


fluca1978:
Current head has problems when compiling with gcc 12 and `-W` error option, as reported in comment https://github.com/agroal/pgagroal/pull/336#issuecomment-1348346821 .

PR #338, opened 2022-12-14
Update install instructions to pin to latest stable release

brettgoss:
This updates the release build install instructions to pin to the latest stable release (1.5.x), since master is currently unstable, as mentioned in #337.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,11 @@ Alternative [clang 8+](https://clang.llvm.org/) can be used.
 
 ### Release build
 
-The following commands will install `pgagroal` in the `/usr/local` hierarchy
+The following commands will install the latest stable release of `pgagroal` in the `/usr/local` hierarchy
 and run the default configuration.
 
 ```sh
-git clone https://github.com/agroal/pgagroal.git
+git clone https://github.com/agroal/pgagroal.git --branch 1.5.x
 cd pgagroal
 mkdir build
 cd build


jesperpedersen:
I think it would be better to information like this in a `doc/DEVELOPMENT.md` file since we can assume that people using the `master` branch are developers.

Maybe we can move a lot of the `README.md` information into `doc/DEVELOPMENT.md`.

Would you like to take a stab at it ?

jesperpedersen:
Remember to add yourself to the `AUTHORS` file as well

fluca1978:
I think it would be nice to have a floating tag `latest` (or something) that points to the latest stable release tag.
In this case we can point the users to checkout `latest` without having to deal with a particular version number.

brettgoss:
> since we can assume that people using the `master` branch are developers.

@jesperpedersen Ah, I had assumed the 'Release Build' instructions were for the stable install of pgagroal. I now see that there is an [install doc](https://github.com/agroal/pgagroal/blob/master/doc/tutorial/01_install.md), however it's fairly opinionated to dnf, and not obviously mentioned in the readme so easy to miss. 

Perhaps a section for `Using pgagroal` (similar to what [tini](https://github.com/krallin/tini#using-tini) does) above the development instructions? And I guess the development instructions could also be extracted as you suggest, but IMO that isn't necessary as long as it's clear that those instructions are in fact for development only.

My thought is that for someone discovering this project via the GH Repo, they would likely just skim the readme for install instructions, and the `Release build` sounds like it would be the most stable way to install. 

I'm happy to help with whatever direction you think is best, with the caveat that I don't have a ton of context about this project - just wanted to contribute since my org was using the release build instructions to install pgagroal, which broke our builds with the breaking change committed to `master` the other day. I think it's likely that others might have made the same mistake of not realizing that `master` was not stable.

jesperpedersen:
Ideally, `master` is stable with the up-coming features. However, it may be unstable during certain times in order to help development in a certain area.

`comfig-get` and `config-set` are the main features of 1.6 so any help with testing is appreciated.

But, if you are looking for a true stable branch then the latest release branch should be used - currently 1.5.x

Moving the development information from README.md to doc/DEVELOPMENT.md and adding a link to the tutorial page would be an idea. Feel free to hack something up, and we will look at it.

jesperpedersen:
Please, reopen and update if you think it is still needed

PR #341, opened 2022-12-15
[#339] Count how many settings remain unchanged in configuration reload.

### LINKED ISSUE #339: Understand when a configuration reload is not fully succesfull
The function `transfer_configuration` returns always success even when there are parameters that need a restart. As an example: <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L2481>.

This means that whenever the function is called, there is no hint to the caller that the application of the configuration has failed. I propose to return the value of not-applied parameters instead, so that:
- 0 means everything is applied
- -n means n settings are not applied but the configuration has been reloaded
- +x means reload errors

This can be useful also in the work I'm trying in #333 because I'm using the same function `transfer_configuration` to issue a parameter change on the fly.

fluca1978:
The internal function `transfer_configuration` now returns a negative value in the case of unchanged settings: such value counts the number of parameters that has remained untouched, for example because they require a restart.
The idea is that there are parameter that cannot be changed on the fly and that do not prevent the configuration to be swapped, but it could be interesting to take the count of unchanged configuration settings.

Close #339

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -1935,7 +1935,7 @@ pgagroal_reload_configuration(void)
       goto error;
    }
 
-   if (transfer_configuration(config, reload))
+   if (transfer_configuration(config, reload) > 0)
    {
       goto error;
    }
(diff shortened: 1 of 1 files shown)


fluca1978:
I'm slowly changing m,y mind about this: what about leaving `transfer_configuration` to return a generic non-zero value when an error happens and accept an `int*` to store the unchanged count?
This could simplify the usage of the function, like `if ( transfer_configuration(..)){ error}`.

jesperpedersen:
Rebased, and merged.

Issue #325, opened 2022-08-25
MacOS X support


jesperpedersen:
Identified as

```
System is Darwin
```

2010YOUY01:
I'm trying to make it run on OSX, it can build so far, but there are several issues:
1. ld linker on OSX doesn't seem to have '-z' option, it will throw errors `ld: unknown option: -z
clang: error: linker command failed with exit code 1 (use -v to see invocation)`, so those lines to set the option are disabled https://github.com/agroal/pgagroal/blob/master/src/CMakeLists.txt#L142 I'm not sure if it will cause any problem.
2. How can I perform something like an integration test to make sure everything works?

jesperpedersen:
1) Hmm, checks like https://github.com/agroal/pgagroal/blob/master/src/CMakeLists.txt#L202 should catch this, and disable the linker flag. Any idea of why it fails ?

2) Unfortunately, it is a manual setup - you can follow the tutorials in order to get a basic setup - https://github.com/agroal/pgagroal/tree/master/doc/tutorial

Thanks for working on this !

2010YOUY01:
> 1. Hmm, checks like https://github.com/agroal/pgagroal/blob/master/src/CMakeLists.txt#L202 should catch this, and disable the linker flag. Any idea of why it fails ?
> 2. Unfortunately, it is a manual setup - you can follow the tutorials in order to get a basic setup - https://github.com/agroal/pgagroal/tree/master/doc/tutorial
> 
> Thanks for working on this !

Cool! I will setup and test it later.

For 1. according to the compiler manual
> /usr/bin/clang --help | grep "\-Wl"
  -Wl,<arg>               Pass the comma separated arguments in <arg> to the linker
> /usr/bin/clang --help | grep "\-z "
  -z <arg>                Pass -z <arg> to the linker

`-Wl` and `-z` options in the compiler seem only forward it to the linker, and it didn't care if the linker supports that. (same for gcc)
The doc for linker on Linux said it support `-z` option, and OSX doesn't.

So both on Linux and OSX, https://github.com/agroal/pgagroal/blob/master/src/CMakeLists.txt#L202 `HAS_RELRO` is set.
I tried changing it to `check_linker_flag("-z relro" HAS_RELRO)`. On Linux it's set, on OSX it's not set, which is the expected behavior.
If that's the case, this issue can be fixed first.

jesperpedersen:
[...]

PR #351, opened 2023-03-17
Darwin support

2010YOUY01:
For issue https://github.com/agroal/pgagroal/issues/325
This is an awesome project! I (and possibly other contributors) would like to do development on MacOS X, so I'm trying to make it build on MacOS X.
This patch is currently separated into small commits for review, and will be squashed later if it's possible to merge.

### [Fix linker in cmake.](https://github.com/agroal/pgagroal/commit/f6257afee85d8fb8ca81f0319fd0288c6433a29e)
The semantics of `check_c_compiler_flag(-Wl,-z,relro HAS_RELRO)` is to check if you pass `-Wl -z relro` to the compiler and if it will throw an error, the compiler will simply pass it to the linker and didn't actually check if linker will throw an error for `-z relro` option.  e.g. I tested `check_c_compiler_flag(-Wl,random,var FOO)` and `FOO` is set to true, so the current implementation might have false positive and make the project's cmake fail on MacOS X. Detail also discussed at https://github.com/agroal/pgagroal/issues/325#issuecomment-1470825193
### [find OpenSSL and headers on Darwin](https://github.com/agroal/pgagroal/commit/6e8744d3ab9d9d1dcb3d337a00dfc3fd3b57468e)
Setup include path for OpenSSL otherwise `openssl/ssl.h` can't be found when compiling.
Defined `_DARWIN_C_SOURCE` to locate system headers
Fixed `pgagroal_set_proc_title` under MacOS X, I tested it using `ps`, and works fine.
### [fix rlim_t type print issue.](https://github.com/agroal/pgagroal/commit/d7f7e8ac877496bb695ab49237bdef12681b3889)
`rlim_t` type seems might be either 4 bytes or 8 bytes on different architectures, and its 8 byte size in MacOS X caused problem.
```
 error: format specifies type 'long' but the argument has type 'unsigned long long' [-Werror,-Wformat]
      errx(1, "max_connections is larger than the file descriptor limit (%ld available)", flimit.rlim_cur - 30);
                                                                         ~~~              ^~~~~~~~~~~~~~~~~~~~
                                                                         %llu
```
 Check size at compile time seems hard to do, and since pgagroal is currently deployed ok and might be only used for development under MacOS X, here just simply did type casting to fix the error.
### [Darwin CI setup and formatting.](https://github.com/agroal/pgagroal/commit/8a11fc5f6988e69892de47e3905f5f960d19c317)
CI, AUTHOR file, and several formatting issues caught by [uncrustify.sh](https://github.com/agroal/pgagroal/blob/master/uncrustify.sh)

## Testing
Deployed on local machine `Darwin Kernel Version 19.6.0`, run several simple queries through `pgagroal` proxy, and tested the concurrent connection using `pgbench`, no error reported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
     - master
 
 jobs:
-  build:
+  build-linux:
 
     runs-on: ubuntu-latest
     
@@ -46,3 +46,45 @@ jobs:
     - name: CLANG/make
       run: make
       working-directory: /home/runner/work/pgagroal/pgagroal/build/
+
+
+
+  build-macos:
+
+    runs-on: macos-latest
+    
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install Homebrew
+      run: /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+    - name: Update system
+      run: brew update
+    - name: Install openssl
+      run: brew install openssl
+    - name: Install libev
+      run: brew install libev
+    - name: Install rst2man
+      run: brew install docutils
+    - name: Install clang
+      run: brew install llvm
+    - name: GCC/mkdir
+      run: mkdir build
+      working-directory: /Users/runner/work/pgagroal/pgagroal/
+    - name: GCC/cmake
+      run: export CC=/usr/bin/gcc && export OPENSSL_ROOT_DIR=`brew --prefix openssl` && cmake -DCMAKE_BUILD_TYPE=Debug ..
+      working-directory: /Users/runner/work/pgagroal/pgagroal/build/
+    - name: GCC/make
+      run: make
+      working-directory: /Users/runner/work/pgagroal/pgagroal/build/
+    - name: rm -Rf
+      run: rm -Rf build/
+      working-directory: /Users/runner/work/pgagroal/pgagroal/
+    - name: CLANG/mkdir
+      run: mkdir build
+      working-directory: /Users/runner/work/pgagroal/pgagroal/
+    - name: CLANG/cmake
+      run: export CC=/usr/bin/clang && export OPENSSL_ROOT_DIR=`brew --prefix openssl` && cmake -DCMAKE_BUILD_TYPE=Debug ..
+      working-directory: /Users/runner/work/pgagroal/pgagroal/build/
+    - name: CLANG/make
+      run: make
+      working-directory: /Users/runner/work/pgagroal/pgagroal/build/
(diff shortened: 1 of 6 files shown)


jesperpedersen:
Nice !

Lets start by doing a squash and force push - plus remove unrelated changes. Feel free to open multiple pull requests...

2010YOUY01:
Squash and force pushed. I forget to mention all modifications are required for the project to build on MacOS X.

src/libpgagroal/configuration.c
```diff
@@ -2629,7 +2629,6 @@ transfer_configuration(struct configuration* config, struct configuration* reloa
    // changes the pgagroal-cli probably will not be able to connect in any case!
    restart_string("unix_socket_dir", config->unix_socket_dir, reload->unix_socket_dir, false);
 
-
```
jesperpedersen, line 2632: Lets remove this change from this pull request

src/libpgagroal/message.c
```diff
@@ -1275,7 +1275,7 @@ ssl_read_message(SSL* ssl, int timeout, struct message** msg)
          err = SSL_get_error(ssl, numbytes);
          switch (err)
          {
-            case SSL_ERROR_NONE: 
+            case SSL_ERROR_NONE:
```
jesperpedersen, line 1278: Lets revert this as well

src/libpgagroal/message.c
```diff
@@ -1371,7 +1371,7 @@ ssl_write_message(SSL* ssl, struct message* msg)
 
          switch (err)
          {
-            case SSL_ERROR_NONE: 
+            case SSL_ERROR_NONE:
```
jesperpedersen, line 1374: Lets revert this as well

jesperpedersen (changes requested):

jesperpedersen:
Just some white space changes removed in order to minimize the patch...

2010YOUY01:
Sure, removed. And thanks for reviewing!

jesperpedersen:
Merged.

Thank you for your contribution ! It was very helpful !

Feel free to port the patch to the "sister" projects - pgmoneta and pgexporter :)

2010YOUY01:
I haven't looked into pgmoneta and pgexporter yet. Let's first create issues linked to this PR, so that it will be easier for other contributors to use this patch as a reference if they want to port it themselves. 
However, I have already ported and used pgprtdbg. It's very handy for understanding the underlying pg protocol communication of pgagroal, will send a patch soon!

jesperpedersen:
@2010YOUY01 No problem !

PR #358, opened 2023-03-27
[#355] Add line number field in various configuration structs

### LINKED ISSUE #355: Line number in structure limits but not in hba?
While working on #239, I noted that the `struct limit` has a line number field <https://github.com/agroal/pgagroal/blob/master/src/include/pgagroal.h#L291> while other similar structures, most notably `struct hba` does not.
The line number is useful to report back errors in the configuration, e.g., <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L1303>.

In order to keep everything consistent, I would like to either remove the line number or to add it in similar structures, at leas in `struct hba`.

fluca1978:
Added a `lineno` field into structures `struct hba` and `struct server`.
This simplifies error reporting, since it is now possible to drive the user to the line number that is causing a problem.

Thanks to this, it is possible to provide better messages to the users about misconfigurations.
For example, if the user mispells an entry into the HBA file:

```
FATAL configuration.c:1162 Unknown HBA type: HOSTSS (/etc/pgagroal/pgagroal_hba.conf:7)
```

Changes to a few log/warn messages to become more easy to read.

Close #355

### DIFF
--- a/src/include/pgagroal.h
+++ b/src/include/pgagroal.h
@@ -237,6 +237,7 @@ struct server
    int port;               /**< The port of the server */
    bool tls;               /**< Use TLS if possible */
    atomic_schar state;     /**< The state of the server */
+   int lineno;             /**< The line number within the configuration file */
 } __attribute__ ((aligned (64)));
 
 /** @struct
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Rebased, and merged.

Issue #361, opened 2023-03-29
max_connection_age


jesperpedersen:
Create a `max_connection_age` setting that will define how long a connection will live - in seconds.

This configuration option will be checked once the connection is returned to the pool, or during the idle check.

Default is `-1` - e.g. not enabled.

fluca1978:
What is the aim of killing a connection? I mean, isn't a way of augment latency, even if in favor of security?

If I get it right, the workflow should be:
- in `pgagroal_return_connection` check if `timestamp` is greater than `max_connection_age`
- if the above is true, `goto kill_connection` but probably there is something more to do in order to cleanup the connection

jesperpedersen:
It is an option to remove a connection after a maximum amount of time.

So, it is `pgagroal_return_connection` - check - then `goto kill_connection` in that workflow

2010YOUY01:
I also want to make sure I get it right:
So the current implementation has a `timestamp` field inside `struct connection`, and an `idle_timeout` field inside `struct configuration`, they're used to make sure a connection won't stay in `STATE_FREE` (after returning back to the pool) longer
 than `idle_timeout`.
This `max_connection_age` is to make sure the time from a connection successfully get from `pgagroal_get_connection()` (turned into `STATE_INUSE`) to `pgagroal_return_connection()` or idle check/validation won't longer than this specified time period.

jesperpedersen:
@2010YOUY01 Yes

2010YOUY01:
Awesome, may I work on this issue?

fluca1978:
@2010YOUY01 if you work on this, there is a function in the configuration code, named `as_seconds` that can parse the parameter.
Remember to keep the parameter consisten across configuration reloads, see ``metrics_cache_max_age` for an hint.

hao-tian-xu:
@2010YOUY01 Hi, I asked to find the issue to work on. Sorry for this..

jesperpedersen:
This has been merged - thanks @hao-tian-xu !

PR #365, opened 2023-04-03
[#361] Add a `max_connection_age` setting

hao-tian-xu:
This defines how long a connection will live in seconds
- Add a `max_connection_age` member to `struct configuration`. Its implementation and related functions are similar to `idle_timeout`
- Add a `start_time` member to `struct connection`. Its implementation is similar to `timestamp`

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -61,6 +61,7 @@ The available keys and their accepted values are reported in the table below.
 | log_disconnections | `off` | Bool | No | Log disconnects |
 | blocking_timeout | 30 | Int | No | The number of seconds the process will be blocking for a connection (disable = 0) |
 | idle_timeout | 0 | Int | No | The number of seconds a connection is been kept alive (disable = 0) |
+| max_connection_age | 0 | Int | No | The maximum number of seconds that a connection will live (disable = 0) |
 | validation | `off` | String | No | Should connection validation be performed. Valid options: `off`, `foreground` and `background` |
 | background_interval | 300 | Int | No | The interval between background validation scans in seconds |
 | max_retries | 5 | Int | No | The maximum number of iterations to obtain a connection |
--- a/doc/man/pgagroal.conf.5.rst
+++ b/doc/man/pgagroal.conf.5.rst
@@ -95,6 +95,9 @@ blocking_timeout
 idle_timeout
   The number of seconds a connection is been kept alive (disable = 0). Default is 0
 
+max_connection_age
+  The maximum number of seconds that a connection will live (disable = 0). Default is 0
+
 validation
   Should connection validation be performed. Valid options: off, foreground and background. Default is off
 
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -267,6 +267,11 @@ start:
          }
       }
 
+      if (config->connections[*slot].start_time == -1)
+      {
+         config->connections[*slot].start_time = time(NULL);
+      }
+
       config->connections[*slot].timestamp = time(NULL);
 
       atomic_store(&prometheus->client_wait_time, difftime(time(NULL), start_time));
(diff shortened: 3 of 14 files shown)


hao-tian-xu:
- I've tried two ways for background checks. One is to add code to idle timeout functions, and another is to add separate functions. Both can work.
  - I currently prefer the separate method to have different time intervals of background checks.
  - If it's ok, I'll add new TRACKER and STATE, and a new Prometheus metric.
- A `start_time` member is added to `struct connection` as the start timestamp. It is initialized in `pgagroal_return_connection` if prefilled or in `pgagroal_get_connection` if not.
- I did a few simple tests. Any suggestions on how to do comprehensive testing?
- Another trivial question: should I set the default value to 0 to be consistent with `idle_timeout`?

src/libpgagroal/pool.c
```diff
@@ -579,6 +590,65 @@ pgagroal_idle_timeout(void)
...
+   now = time(NULL);
+   prefill = false;
+
+   pgagroal_log_debug("pgagroal_max_connection_age");
+
+   /* Here we run backwards in order to keep hot connections in the beginning */
+   for (int i = config->max_connections - 1; i >= 0; i--)
+   {
+      free = STATE_FREE;
+      age_check = STATE_IDLE_CHECK; // TODO
```
jesperpedersen, line 615: Yes, do a new `STATE_MAX_CONNECTION_AGE`

jesperpedersen (changes requested):

hao-tian-xu:
Thanks for the feedback! I added new STATE, TRACKER, Prometheus metric, and documentation for `max_connection_age`

doc/CONFIGURATION.md
```diff
@@ -60,7 +60,8 @@ The available keys and their accepted values are reported in the table below.
 | log_connections | `off` | Bool | No | Log connects |
 | log_disconnections | `off` | Bool | No | Log disconnects |
 | blocking_timeout | 30 | Int | No | The number of seconds the process will be blocking for a connection (disable = 0) |
-| idle_timeout | 0 | Int | No | The number of seconds a connection is been kept alive (disable = 0) |
+| idle_timeout | 0 | Int | No | The number of seconds a connection is been kept alive since becomes idle (disable = 0) |
+| max_connection_age | 0 | Int | No | The number of seconds a connection is been kept alive since starts (disable = 0) |
```
jesperpedersen, line 64: "... since it times out"

doc/man/pgagroal.conf.5.rst
```diff
@@ -93,7 +93,10 @@ blocking_timeout
   The number of seconds the process will be blocking for a connection (disable = 0). Default is 30
 
 idle_timeout
-  The number of seconds a connection is been kept alive (disable = 0). Default is 0
+  The number of seconds a connection is been kept alive since becomes idle (disable = 0). Default is 0
+
+max_connection_age
+  The number of seconds a connection is been kept alive since starts (disable = 0). Default is 0
```
jesperpedersen, line 99: Ditto

jesperpedersen (changes requested):

doc/CONFIGURATION.md
hao-tian-xu: If I get it right, 
- `idle_timeout` is "The number of seconds a connection can remain idle before it is automatically closed"
- `max_connection_age` is "The number of seconds a connection can remain active before it is automatically closed if it is not being used"

jesperpedersen: It is the maximum seconds that a connection will live - it will be checked upon returned to the pool, or during idle timeout

hao-tian-xu: Thanks for the feedback! I updated the `max_connection_age` docs and kept the `idle_timeout` docs untouched.

Checking upon returned to the pool is added to `pgagroal_return_connection`.

Both ways of checking are tested using a basic setup.

src/libpgagroal/pool.c
```diff
@@ -364,9 +369,27 @@ pgagroal_return_connection(int slot, SSL* ssl, bool transaction_mode)
...
    int state;
    struct configuration* config;
+   time_t now;
+   signed char in_use;
+   signed char age_check;
 
    config = (struct configuration*)shmem;
 
+   /* Kill the connection, if it lives longer than max_connection_age */
+   now = time(NULL);
```
jesperpedersen, line 379: We need some extra checks here, right ?

Is `max_connection_age > 0`

jesperpedersen (changes requested):

src/libpgagroal/pool.c
hao-tian-xu: Thanks for the review! I've tested thses cases this time:
- `idle_timeout` off, `max_connection_age` on, one prefill, no connection / get one connection and then close
- `idle_timeout` on, `max_connection_age` off, one prefill, get one connection and then close
- `idle_timeout = 20`, `max_connection_age = 40`, one prefill, get two connections and then close

jesperpedersen: Excellent

jesperpedersen:
Can you rebase ?

hao-tian-xu:
Sure, rebased and fixed conflicts

jesperpedersen:
Merged.

Issue #366, opened 2023-04-16
Prometheus exporter `/metrics` endpoint does not produce correct output 


nikitabugrovsky:
**Describe the bug**

The prometheus exporter `/metrics` endpoint does not produce correct output for
```
#TYPE pgagroal_query_count count
```
That "count" at the end is invalid. It has to be "counter". 
Prometheus and the signalfx cannot parse that correctly.

Offending lines are:

https://github.com/agroal/pgagroal/blob/dc04446f0ea4c62854e130072eb580c361a6fd40/src/libpgagroal/prometheus.c#L1416

https://github.com/agroal/pgagroal/blob/dc04446f0ea4c62854e130072eb580c361a6fd40/src/libpgagroal/prometheus.c#L1449

**Version**

pgagroal 1.5.1

PR #367, opened 2023-04-16
[#366] Fix /metrics endpoint does not produce correct output

nikitabugrovsky:


jesperpedersen:
Merged.

PR #377, opened 2023-08-08
GSoC: Adding basic Caching system

SudeepRed:
This PR adds query caching functionality to pgagroal.
## Implementation
Functions to be implemented:
- [x] pgagroal_query_cache_init();
- [x] pgagroal_query_cache_get();
- [x] pgagroal_query_cache_invalidate();
- [x] pgagroal_query_cache_update();
- [x] pgagroal_query_cache_add();
- [x] pgagroal_query_cache_clear();
## Testing
Functions to be Tested:
- [x] pgagroal_query_cache_init();
- [x] pgagroal_query_cache_get();
- [x] pgagroal_query_cache_invalidate();
- [x] pgagroal_query_cache_update();
- [x] pgagroal_query_cache_add();
- [x] pgagroal_query_cache_clear();

To test the provided functions on your local machine, execute the command (assuming you are using the debug build)
 ```
./pgagroal -c pgagroal.conf -a pgagroal_hba.conf -t 
```
 If you want to exclude cache tests, simply remove the `t` flag from the command.

### DIFF
--- a/src/libpgagroal/query_cache.c
+++ b/src/libpgagroal/query_cache.c
@@ -92,18 +92,14 @@ error:
    return 1;
 }
 struct hashEntry*
-pgagroal_query_cache_get(struct query_cache* cache, struct hashTable** Table, struct hashEntry* key)
+pgagroal_query_cache_get(struct query_cache* cache, struct hashTable** Table, char* key)
 {
-   for (int i = 0; i < cache->max_elements; i++)
+
+   struct hashTable* s = NULL;
+   HASH_FIND_STR(*Table, key, s);
+   if (s)
    {
-
-      int x = strncmp(cache->cache[i].key->key, key->key, key->length);
-
-      if (x == 0)
-      {
-
-         return cache->cache[i].data;
-      }
+      return s->data;
    }
    return NULL;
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by jesperpedersen** [comment]
Remember to update your progress log in the forum thread

**Comment by SudeepRed** [comment]
> I don't think our key will be char key... I would expect that it is char* -> void* and void* is query_cache_max_size - or maybe query_cache_entry_max_size - leave the former to total size of the query cache - that way you can calculate the max number of cache entries...
This is just the initial rough implementation. This will change a lot in the coming days. I was just testing out and pushing the changes as I was implementing 
> Always, squash and force push your changes
Yes, I will do this just before merging since I might need to revert in the future. Also, why don't we `squash and merge` the PRs? IT basically does the same thing right? One commit on the main branch

**Comment by jesperpedersen** [comment]
Yes, one commit per feature - e.g. all of this work is ideally one commit

File: src/libpgagroal/query_cache.c [hunk 1]
```diff
@@ -203,10 +200,26 @@ pgagroal_query_cache_add(struct query_cache* cache, struct hashTable** Table, st
       memset(copy_data->value, 0, data->length);
       memcpy(copy_data->value, data->value, data->length);
       copy_data->length = data->length;
+      int idx = cache->max_elements;
+      pgagroal_log_info("Adding to cache %d", idx);
+
+      cache->cache[idx].key = copy_key;
+      cache->cache[idx].data = copy_data;
+      cache->max_elements = idx + 1;
+      pgagroal_log_info("fromcache to cache %s", cache->cache[idx].key->key);
```
jesperpedersen, line 209: More logging...

SudeepRed: Yeah this is something im testing, will remove after we resolve this issue https://github.com/agroal/pgagroal/discussions/376#discussioncomment-7480504

PR #379, opened 2023-09-06
Fix PG protocol handling on big-endian

df7cb:
Replace the homegrown swap-bytes machinery by system functions that work correctly independently from the host byte order. pgagroal_read/write_* were wrong on big-endian machines before.

Problem noticed on Debian s390x where connects to pgagroal were just hanging indefinitely.

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -166,21 +166,6 @@ void
 pgagroal_write_string(void* data, char* s);
 
 /**
- * Is the machine big endian ?
- * @return True if big, otherwise false for little
- */
-bool
-pgagroal_bigendian(void);
-
-/**
- * Swap
- * @param i The value
- * @return The swapped value
- */
-unsigned int
-pgagroal_swap(unsigned int i);
-
-/**
  * Print the available libev engines
  */
 void
--- a/src/libpgagroal/security.c
+++ b/src/libpgagroal/security.c
@@ -40,6 +40,11 @@
 #include <utils.h>
 
 /* system */
+#ifdef __APPLE__
+#include <machine/endian.h>
+#else
+#include <endian.h>
+#endif
 #include <stdatomic.h>
 #include <stdbool.h>
 #include <stdio.h>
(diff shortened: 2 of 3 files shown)


df7cb:
Added missing `#include <endian.h>` in security.c now.

jesperpedersen:
Maybe `#if HAVE_OSX` and `#include <machine/endian.h>` ?

https://stackoverflow.com/questions/20813028/endian-h-not-found-on-mac-osx

df7cb:
I was googling around and found like half a dozen different answers. Will give this one a shot now.

jesperpedersen:
Keep an eye on the CI environment

jesperpedersen:
Its `HAVE_OSX`

jesperpedersen:
Should it be a "wrapper" include, or an `#if` for this ?

df7cb:
Sorry the commit where I had tried `HAVE_OSX` had a silly typo elsewhere. But it looks like `__APPLE__` worked, and it found the header, but `htobe32` is still not available.

jesperpedersen:
@df7cb Any updates ?

df7cb:
I believe my patch is correct, but MacOS lacks the modern `htobe*` functions.

jesperpedersen:
@df7cb Maybe we should merge this on `master` and see if somebody with a Mac steps up and fixes it such that it can be backported ?

Likely the issue is the same with pgmoneta and pgexporter...

jesperpedersen:
Thinking about it some more - we should leave the existing functionality for `#ifdef HAVE_OSX` in, and use your changes for all other platforms.

Could you rework your patch for that ?

df7cb:
Fwiw, this might be a possible fix: https://gist.github.com/yinyin/2027912 (but I currently lack resources to actually try it)

jesperpedersen:
Could you rebase ? And, what is the status ?

df7cb:
To be honest, the status is "I have a fix that works on sane operating systems, and MacOS is clearly not"...

I don't have the resources to pick this up again, sorry.

jesperpedersen:
That is quite fair !

Thanks for making the report / patch in the first place !

PR #389, opened 2023-11-07
[#388] Add `conf ls` to get the configuration file locations

### LINKED ISSUE #388: Add a command to `pgagroal-cli` get configuration file locations
I propose to add a command to `pgagroal-cli` to get information about the configuration files used by the deamon.
My proposal is to add `conf ls` (or `conf files`), so that for instance:

```
$ pgagroal-cli conf ls
Configuration file:        /etc/pgagroal/pgagroal.conf
HBA file:                  /etc/pgagroal/pgagroal_hba.conf
Limit file:                /etc/pgagroal/pgagroal_databases.conf
Frontend users file:       /etc/pgagroal/pgagroal_users.conf
Admins file:               /etc/pgagroal/pgagroal_frontend_users.conf
Superuser file:            /etc/pgagroal/pgagroal_admins.conf
```

In this way, even remotely, it is possible to know where the configuration files are located on the system running `pgagroal`.

fluca1978:
Introducing the command `conf ls` to `pgagroal-cli` that prints out where the configuration files are located.
A new management key, action and three new management methods have been added to implement the logic.
The idea is that, once the action is required, the management prints on the socket every single `struct configuration*` path variables, and then the management read method reads back in the same order to the same list of values to be printed on the output screen.

In order to achieve this, two new static functions have been added to the managemenet in order to write and read from the socket a single path. A path is written with a size and the following path, assuming it will never be larger than MAX_PATH or an error will be thrown. In the case a path is NULL or empty, an empty string will be written on the socket and so the final result is that the entry will be printed as empty in the command output.

Documentation updated.

Shell completions updated.

Fixes also a few log_trace in the cli application.

Close #388

### DIFF
--- a/contrib/shell_comp/pgagroal_comp.bash
+++ b/contrib/shell_comp/pgagroal_comp.bash
@@ -25,7 +25,7 @@ pgagroal_cli_completions()
                 COMPREPLY+=($(compgen -W "server prometheus" "${COMP_WORDS[2]}"))
                 ;;
 	    conf)
-		COMPREPLY+=($(compgen -W "reload get set" "${COMP_WORDS[2]}"))
+		COMPREPLY+=($(compgen -W "reload get set ls" "${COMP_WORDS[2]}"))
 		;;
 	    status)
 		COMPREPLY+=($(compgen -W "details" "${COMP_WORDS[2]}"))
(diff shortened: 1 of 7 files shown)


jesperpedersen:
Merged.

Issue #403, opened 2024-02-19
Improve error messages in `pgagroal-cli`


fluca1978:
While working on #399 I discovered that `pgagroal-cli` does not provide good error messages.
For example.
- `pgagroal-cli conf` -> `pgagroal-cli: unknown command conf` while it should be something like *command conf requires a subcommand*
-  `pgagroal-cli conf get` -> `pgagroal-cli: unknown command conf` while it should be something like *command **conf get** requires an argument name*

Similar behaviour for other commands.
The problem is that the handling of command line arguments in `pgagroal-cli` is done with a catch-all error block <https://github.com/agroal/pgagroal/blob/master/src/cli.c#L652> so every rule that cannot fully parse jumps into such block and provides no useful information than a generic error.

My idea is to trap errors near every parsing rule, so that the error message can have context-aware information to send back to the user.

fluca1978:
Same issue happens in `pgagroal-admin`:

```
% pgagroal-admin user
pgagroal-admin: unknown command or subcommand <user>
```

while it should report something like *command user requires a subcommand*.

decarv:
Hello, @fluca1978. 

As we have discussed, I am ready to start working on this issue. In the near future, I will outline my proposed approach for us to discuss further, so we're aligned on the solution's direction before I dive deeper into implementation.

decarv:
While investigating this issue, I encountered a segmentation fault when running pgagroal-cli when no command is specified a command, but a password is specified.

It appears the segfault occurs because the program attempts to free a password that points to a position in *argv.

**Steps to reproduce**
```shell
./pgagroal-cli -h localhost -p 2345 -U test -P test
```
**Proposed Solution:**
This can be addressed by setting the initial value of `do_free` to `false` and then setting it to `true` only if `pgagroal_get_password` is called.

Here is the relevant segment of the proposed fix (please notice my comments):

```C
// NOTE(decarv): adjust below
bool do_free = false;

// ...

/* Password */
if (password == NULL)
{
    // NOTE(decarv): This inner check seems unnecessary since `password` is confirmed to be NULL above. Am I missing something?
    if (password != NULL) 
    {
        free(password);
        password = NULL;
    }

    printf("Password : ");
    password = pgagroal_get_password();
    printf("\n");
    
    // NOTE(decarv): add below
    do_free = true;
}
// NOTE(decarv):  Suggested to remove the following else block to avoid resetting `do_free` unnecessarily.
else
{
   do_free = false;
}
```

[...]

PR #408, opened 2024-03-02
Fix SegFault in pgagroal-cli

decarv:
`pgagroal-cli` was raising segmentation fault when no command is specified, but a password is specified.

The fix for this followed [this](https://github.com/agroal/pgagroal/issues/403#issuecomment-1973895687) suggestion and removes do_free completely. `password` is now always dynamically allocated.

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -163,7 +163,6 @@ main(int argc, char** argv)
    char* password = NULL;
    bool verbose = false;
    char* logfile = NULL;
-   bool do_free = true;
    int c;
    int option_index = 0;
    size_t size;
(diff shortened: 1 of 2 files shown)


src/cli.c
```diff
@@ -676,7 +669,7 @@ main(int argc, char** argv)
    pgagroal_stop_logging();
    pgagroal_destroy_shared_memory(shmem, size);
 
-   if (do_free)
+   if (password != NULL)
```
jesperpedersen, line 672: `free() will work with a `NULL` so we don't need this check

jesperpedersen:
You need to add your name to `AUTHORS` as well

jesperpedersen (changes requested):

jesperpedersen:
Rebased, and merged.

Welcome to the community !

decarv:
> Rebased, and merged.
> 
> Welcome to the community !
> 
> Thanks for your contribution !

Thank you, @jesperpedersen! Looking forward to more contributions!

Issue #410, opened 2024-03-08
Deprecated `pgagroal-cli` commands `config-get` and `config-set` do not work as expected


decarv:
**Describe the bug**

Deprecated `pgagroal-cli` commands should still execute the corresponding new commands, but this is not true for `config-get` and `config-set`. 

**To Reproduce**

Run in the terminal the commands and get the error message `unknown command config-get`.

```shell
$ ./pgagroal-cli -v config-get port
pgagroal-cli: command <config-get> has been deprecated by <conf get> since version 1.6
2024-03-07 22:38:59 TRACE cli.c:459 Command: <conf get> [(null)]
pgagroal-cli: unknown command config-get
```

```shell
$ ./pgagroal-cli -v config-set port 2346
pgagroal-cli: command <config-set> has been deprecated by <conf set> since version 1.6
2024-03-07 22:41:56 TRACE cli.c:466 Command: <conf set> [(null)] = [(null)]
pgagroal-cli: unknown command config-set
```

**Version**

pgagroal-cli 1.7.0

**Diagnosis**

Neither `config_key` or `config_value` are ever passed to `parse_deprecated_command`, thus being impossible to parse these values. This function needs to be modified.

decarv:
Sure! I plan to open a PR soon that should close #403 and this issue. But this issue should be a quick fix and thus quicker to review. I will work on it.

jesperpedersen:
See if it worth to split the patch into the separate issue.

Otherwise, just reference both issues in the commit message

decarv:
@jesperpedersen Splitting the patch was impractical.

I just referenced both issues.

jesperpedersen:
That is quite ok

fluca1978:
Close via 11a1f47

PR #411, opened 2024-03-09
[#403][#410] Improve error messages for pgagroal-cli

decarv:
This commit improves the error messages for pgagroal-cli and pgagroal-admin commands by modifying the parse_command function. 

It also tries to simplify adding new commands. Now adding a command with the structure "<command> [subcommand] [arg] [arg] ..." requires the addition of the command enum and its struct in the `command_table`.

This means that the command parsing now involves the interpretation of a command_table defined in each cli.c and admin.c files.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -68,6 +68,94 @@ static int remove_user(char* users_path, char* username);
 static int list_users(char* users_path);
 static char* generate_password(int pwd_length);
 
+const struct pgagroal_command command_table[] =
+{
+   {
+      .command = "master-key",
+      .subcommand = "",
+      .accepted_argument_count = {0},
+      .deprecated = false,
+      .action = ACTION_MASTER_KEY,
+      .log_message = "<master-key>",
+   },
+   {
+      .command = "user",
+      .subcommand = "add",
+      .accepted_argument_count = {0},
+      .deprecated = false,
+      .action = ACTION_ADD_USER,
+      .log_message = "<user add> [%s]",
+   },
+   {
+      .command = "user",
+      .subcommand = "edit",
+      .accepted_argument_count = {0},
+      .deprecated = false,
+      .action = ACTION_UPDATE_USER,
+      .log_message = "<user edit> [%s]",
+   },
+   {
+      .command = "user",
+      .subcommand = "del",
+      .accepted_argument_count = {0},
+      .deprecated = false,
+      .action = ACTION_REMOVE_USER,
+      .log_message = "<user del> [%s]",
+   },
+   {
+      .command = "user",
+      .subcommand = "ls",
+      .accepted_argument_count = {0},
+      .deprecated = false,
+      .action = ACTION_LIST_USERS,
+      .log_message = "<user ls>",
+   },
+   {
+      .command = "add-user",
+      .subcommand = "",
+      .accepted_argument_count = {0},
+      .deprecated = true,
+      .action = ACTION_ADD_USER,
+      .log_message = "<deprecated: use 'user add'> [%s]",
+      .deprecated_since_major = 1,
+      .deprecated_since_minor = 6,
+      .deprecated_by = "user add",
+   },
+   {
+      .command = "update-user",
+      .subcommand = "",
+      .accepted_argument_count = {0},
+      .deprecated = true,
+      .action = ACTION_UPDATE_USER,
+      .log_message = "<deprecated: use 'user edit'> [%s]",
+      .deprecated_since_major = 1,
+      .deprecated_since_minor = 6,
+      .deprecated_by = "user edit",
+   },
+   {
+      .command = "remove-user",
+      .subcommand = "",
+      .accepted_argument_count = {0},
+      .deprecated = true,
+      .action = ACTION_REMOVE_USER,
+      .log_message = "<deprecated: use 'user del'>",
+      .deprecated_since_major = 1,
+      .deprecated_since_minor = 6,
+      .deprecated_by = "user del",
+   },
+   {
+      .command = "list-users",
+      .subcommand = "",
+      .accepted_argument_count = {0},
+      .deprecated = true,
+      .action = ACTION_LIST_USERS,
+      .log_message = "<deprecated: use 'user ls'>",
+      .deprecated_since_major = 1,
+      .deprecated_since_minor = 6,
+      .deprecated_by = "user ls",
+   },
+};
+
 static void
 version(void)
 {
@@ -117,7 +205,8 @@ main(int argc, char** argv)
    bool generate_pwd = false;
    int pwd_length = DEFAULT_PASSWORD_LENGTH;
    int option_index = 0;
-   int32_t action = ACTION_UNKNOWN;
+   size_t command_count = sizeof(command_table) / sizeof(struct pgagroal_command);
+   struct pgagroal_parsed_command parsed = {.cmd = NULL, .args = {0}};
 
    while (1)
    {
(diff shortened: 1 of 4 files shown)


src/include/utils.h
```diff
@@ -375,6 +375,62 @@ pgagroal_backtrace(void);
 
 #endif
 
+/**
+ * Command parsing utils for pgagroal-cli and pgagroal-admin.
+ */
+
+#define COMMAND_UNKNOWN (-1)
+
+struct pgagroal_parsed_command
```
fluca1978, line 384: I would love to have this struct documented, so that it is clear to everyone what fields mean.
Also this structure should be after the `pgagroal_command`, since it is the result of having `pgagroal_command` being parsed.

src/include/utils.h
```diff
@@ -375,6 +375,62 @@ pgagroal_backtrace(void);
...
+ * allowing for clear definition and handling of all supported commands.
+ */
+struct pgagroal_command
+{
+   const char* command;
+   const char* subcommand;
+   int accepted_argument_count[MISC_LENGTH];
+
+   char* default_argument;
+   char* log_trace;
```
fluca1978, line 423: I would name this field as `log_message` or something like that. The fact that we, at the moment, log as `trace` has nothing to do with the message itself.

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
+   char* command = NULL;
+   char* subcommand = NULL;
+   int command_match[MISC_LENGTH] = {COMMAND_UNKNOWN};
+   int command_match_count = 0;
 
-   // sanity check: if no arguments, nothing to parse!
+   /* Exit if nothing to parse */
    if (argc <= offset)
    {
+      sprintf(cmd->error_message, "A command is required\n");
```
fluca1978, line 975: `snprintf`

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
+   for (int i = 0; i < command_count; i++)
    {
+      if (strncmp(command, command_table[i].command, MISC_LENGTH) == 0)
+      {
+         command_match[command_match_count++] = i;
+      }
+   }
+   if (command_match_count == 0)
+   {
+      sprintf(cmd->error_message, "Invalid command '%s'\n", command);
```
fluca1978, line 990: `snprintf`

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
+   if (cmd->command_code == COMMAND_UNKNOWN)
    {
-      // need to evaluate the database or server or configuration key
-      offset++;
-      *key = argc > offset ? argv[offset] : default_key;
-      if (*key == NULL || strlen(*key) == 0)
+      if (EMPTY_STR(subcommand))
       {
-         goto error;
+         sprintf(cmd->error_message, "Command '%s' requires a subcommand\n", command);
```
fluca1978, line 1008: `snprintf`

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
          {
-            goto error;
+            if (strncmp(subcommand, command_table[command_match[i]].subcommand, MISC_LENGTH) == 0)
+            {
+               cmd->command_code = command_match[i]; /* pushes down the matches */
+            }
+         }
+         if (cmd->command_code == COMMAND_UNKNOWN)
+         {
+            sprintf(cmd->error_message, "Invalid subcommand '%s' for command '%s'\n", subcommand, command);
```
fluca1978, line 1024: `snprintf`

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
-                         unsigned int deprecated_since_minor)
-{
-   // sanity check: if no arguments, nothing to parse!
-   if (argc <= offset)
+   /* Iterate until finds an accepted_arg_count that is equal or greater than the typed command arg_count */
+   cmd->arg_count = argc - offset;
+   int j;
+   for (j = 0; j < MISC_LENGTH && command_table[cmd->command_code].accepted_argument_count[j] < cmd->arg_count; j++)
+   {
+      ;
```
fluca1978, line 1035: I think a `for` with an `if ... break` would be a little more readable than this.

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
+   }
+   if (cmd->arg_count < command_table[cmd->command_code].accepted_argument_count[0])
    {
+      sprintf(cmd->error_message, "Too few arguments provided for command '%s%s%s'\n", command,
```
fluca1978, line 1039: `snprintf`

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
+              (command && !EMPTY_STR(subcommand)) ? " " : "", subcommand);
       return false;
    }
-
-   // first of all check if the command is the same
-   // as the first argument on the command line
-   if (strncmp(argv[offset], command, MISC_LENGTH))
+   if (j == MISC_LENGTH || cmd->arg_count > command_table[cmd->command_code].accepted_argument_count[j])
    {
+      sprintf(cmd->error_message, "Too many arguments provided for command '%s%s%s'\n", command,
```
fluca1978, line 1045: `snprintf`

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
-      *value = argc > offset ? argv[offset] : "*";
+      CONSUME(cmd->args[i]);
    }
 
-   // warn the user if there is enough information
-   // about deprecation
-   if (deprecated_by
-       && pgagroal_version_ge(deprecated_since_major, deprecated_since_minor, 0))
+   /* Warn the user if there is enough information about deprecation */
+   if (command_table[cmd->command_code].deprecated
```
fluca1978, line 1057: I'm wondering if we should remove some clutter of repeated `command_table[cmd->command_code]` by declaring a variable like `pgagroal_command template = command_table[ cmd->command_code];` and then use `template.deprecated`, `template.deprecated_by` and so on.

Another approach, could be to store a pointer to the entry in the command table into the command itself, so that it would be possible to `cmd->template->is_deprecated`. In this way, there would be no more need for a lookup into the table once the command has been successfully parsed.

src/libpgagroal/utils.c
```diff
@@ -954,128 +955,124 @@ pgagroal_backtrace(void)
...
-      // do I need also a value?
-      if (value)
+      else
       {
-         offset++;
-         *value = argc > offset ? argv[offset] : default_value;
-
-         if (*value == NULL || strlen(*value) == 0)
+         offset--;
+         subcommand = "";
```
fluca1978, line 1014: I don't understand why `subcommand` is cleared here, since it could have been `CONSUME`d.
In fact:

```
% pgagroal-admin user foo
Invalid subcommand '' for command 'user'


% pgagroal-cli conf foo
Invalid subcommand '' for command 'conf'
```

fluca1978 (changes requested):
This is a very good job and it almost reflect what I had in mind. 
I think there is really a little more to do to mark this as complete.

decarv:
Thank you for your feedback, @fluca1978.

My main goal with this PR was to seek your insights regarding my implementation and identify areas for improvement based on your vision.

Given your response suggests I'm on the right track, I'd like to discuss additional ideas (below) aimed at simplifying our new command handling approach.

As for the issues you highlighted above, I will address them later this evening.

**1. Refining Command Execution through Callbacks**
I propose shifting from directly assigning action, mode, and log traces to utilizing a callback function activated upon command execution. This function would be the already existing implementation of the actions `flush`, `enabledb`, ..., but with a different signature, thereby streamlining the command structure and execution process. The callback could be defined as:

```c
int (*action)(struct pgagroal_parsed_command* cmd) // returns exit_error
```

The `pgagroal_parsed_command` struct would then include all necessary information for action execution, which currently varies in signature but shares common elements like SSL* ssl, int socket, bool verbose, char output_format, etc.

With this, eliminating the pgagroal_{cli|action}_command_code seems beneficial, as it appears to add complexity without clear value. We could determine the command count using the sizeof(array)/sizeof(array[0]) technique, enhancing simplicity and maintainability.

Initially, I hesitated to implement these changes as they would necessitate further modifications to the existing code. What do you think?

**2. Simplifying Error Handling**
I suggest removing the `error_message` field from the command struct, as I believe directly outputting error messages to `stderr` would be more straightforward and efficient. Is retaining `error_message` essential in your view?

src/admin.c
```diff
@@ -174,96 +296,71 @@ main(int argc, char** argv)
...
-
+         errx(1, "Error for <user ls>");
       }
+
    }
 
    exit(0);
 
 error:
+   if (cmd.error_message[0] != 0)
```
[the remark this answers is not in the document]
decarv: `error_message` is stack allocated so `!= NULL` would always be true. In an earlier comment I suggest removing `error_message` entirely and just call `fprintf(stderr, ...)` or `errx(1, ...)`.

src/include/utils.h
decarv: Agreed.

src/admin.c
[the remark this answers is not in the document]
decarv: or `warnx` as suggested below.

src/libpgagroal/utils.c
decarv: My mistake. It should be cleared at the end of this if. To be honest, the code is a little awkward here, I can improve it.

src/libpgagroal/utils.c
decarv: I like the second approach.

decarv:
While attempting to implement the callback based structure, I found myself redesigning the entire file, which wasn't the goal, so I decided to abort that approach. 

Could you check the PR and let me know if it looks good to you? I'm here to help with any adjustments you need.

fluca1978:
@decarv I added a few comments on https://github.com/agroal/pgagroal/commit/e57675d0654f73a1af86510787465cf2b938bc09, please do the review. Also, could you please reword the commit message to include a little more details about the changes? Not a complete description but something more verbose.

fluca1978:
@decarv I think the last commit is fine. Thanks.

@jesperpedersen PTAL commit https://github.com/agroal/pgagroal/commit/19f30dc5809449d22288a2c6cab1a57128f3a791

src/include/utils.h
```diff
@@ -375,6 +375,67 @@ pgagroal_backtrace(void);
 
 #endif
 
+/**
+ * Command parsing utils for pgagroal-cli and pgagroal-admin.
+ */
+
+#define COMMAND_UNKNOWN (-1)
```
jesperpedersen, line 382: Move this to the beginning of the file

src/include/utils.h
```diff
@@ -375,6 +375,67 @@ pgagroal_backtrace(void);
...
+
+struct pgagroal_command;
```
jesperpedersen, line 384: You don't need the prototype definitions

src/include/utils.h
```diff
@@ -375,6 +375,67 @@ pgagroal_backtrace(void);
...
+ * - mode: A value specifying the mode of operation or context in which the command applies.
+ * > deprecated: A flag indicating whether this command is deprecated.
+ * - deprecated_since_major: The major version number in which the command was deprecated.
+ * - deprecated_since_minor: The minor version number in which the command was deprecated.
+ * - deprecated_by: A string naming the command that replaces the deprecated command.
+ *
+ * This struct is key to extending and maintaining the command processing functionality in pgagroal,
+ * allowing for clear definition and handling of all supported commands.
+ */
+struct pgagroal_command
```
jesperpedersen, line 408: It should be

```
#define

struct

pgagroal_

static

#ifdef DEBUG
```

so move things around a bit

jesperpedersen (changes requested):

jesperpedersen:
Otherwise it looks good

src/include/utils.h
decarv: Not used in the project anymore. Removed.

jesperpedersen:
Merged.

decarv:
Thank you @fluca1978 and @jesperpedersen for the help!

PR #414, opened 2024-03-18
Fix connection errors

fluca1978:
Fixes the connection errors:
- when dealing with faulty JSON commands
- when using `status details`


jesperpedersen:
Merged.

Issue #392, opened 2023-12-09
Add application version within the management protocol


fluca1978:
While working on #391 I got the idea that, since `pgagroal-cli` and `pragroal` are communicating over a socket thru the management protocol,  chances are that the two applications are not at the very same version.
In order to better diagnose problems and even receive a better and more constant behavior, we should transmit the version of `pgagroal-cli` and `pgagroal` thru the socket, so that both parties have information about who they are talking to.
Moreover, this could be a chance to even block request made by older versions of any of the two to the other part.

My idea is to add an header-like block to any request sent out from `pgagroal-cli`, having `pgagroal` to get it in the begin of the message (so, at a *constant* offset) and viceversa, having `pgagroal` to send out the same information as the first part of the communication protocol.

This should not require a lot of changes in the management protocol implementation.
However, I would postpone this to #391 because importing JSON output will speed up the debugging of this implementation.

jesperpedersen:
:+1:  to the idea.

We just have to figure out what we want in that header... could also be stuff like the entire size of the management invocation

fluca1978:
> 👍 to the idea.
> 
> We just have to figure out what we want in that header... could also be stuff like the entire size of the management invocation

Well, since we are going to add stuff, I think there is room for placing a size of the request.
At this point, I would also put the client (e.g., `pgagroal-cli` ) that did the request.

jesperpedersen:
Yeah, the name of the client plus its version number would be a good idea.

The response should include the version of the pgagroal instance.

An idea could be to be to create a variable length header where each "section" is identified by a unique `char` - like `...cpgagroal-cli\0v1.6.0\0...` with a response of `...Spgagroal\0V1.6.0...`

fluca1978:
> An idea could be to be to create a variable length header where each "section" is identified by a unique `char` - like `...cpgagroal-cli\0v1.6.0\0...` with a response of `...Spgagroal\0V1.6.0\0...`

I think that changing `write_header` <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/management.c#L1488> (and consequently, `pgagroal_managament_read_header` <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/management.c#L66>) to interact with the *extended* headers.
But in my opinion, sections should be identified by a fixed letter, for example `F` as *from* (the fact that is a server or a client depends on the side we read the data), `V` for *version*, etc.
[...]

PR #407, opened 2024-03-02
WIP: pgagroal-vault

ashu3103:
Continuation of PR #378 

@jesperpedersen PTAL on this PR

## About this commit

1. Hosted an HTTP server (can be advanced to HTTPS) on the host and port provided in a configuration file `pgagroal_vault.conf`. This server only listens to HTTP requests from the client, also the address {`host`, `port`, `admin_user` and `admin_password`} of management port is provided to the vault to connect to the remote management port and forward requests to that port. The `admin_user` password is provided using a `-u` flag which is mandatory. Typical `.conf` looks like this -
```
[vault]
host = localhost
port = 2500

[main]
host = localhost
port = 2347
user = admin
```

2. The handelling of URLs is done in the following manner :-

    - If client provide a URL `http://localhost:2500/info`, The server will repy 
    
    ```
    HTTP/1.1 200 OK
    Content-Type: text/plain
   
    vault is good
    ```
    - If the client provide a URL `http://localhost:2500/users/<_frontend_user_>`, The server will try to connect and request to the management port of pgagroal to fetch the frontend_user password of the `<_frontend_user_>`, If found the server will respond 
    
    ```
    HTTP/1.1 200 OK
    Content-Type: text/plain
    
    Your temporary password is: <_password_>
    ```   
    - else the server will respond
    
    ```
    HTTP/1.1 200 OK
    Content-Type: text/plain
    
    No such user
    ```   
    - In all other cases the server will give `ERROR 404`

3. Now, this server further have to connect to the management port of pgagroal to call `GET_PASSWORD` method and fetch the frontend user password of the user provided.
4. Meanwhile in pgagroal we have a callback function which will keep rotating frontend passwords of all the `frontend_users` periodically after a certain timeout (`rotate_frontend_password_timeout`) [defined in `pgagroal.conf` file].

PR #415, opened 2024-03-18
[#392] Add application name and version in management protocol

EuGig:
In order to improve debugging, [#392] suggested to write the sender application name and its version into the header over which pgagroal-cli and pgagroal communicate.

Introduces the `write_header_info` function in `management.c`

Every time a pgagroal-cli is used, `cli.c` writes into the communication socket through the `write_header` function.
Then pgagroal replies with one of the `pgagroal_write_` functions, depending on the option specified.
Therefore, the strategy is to call `write_header_info` into both.

Arguments of `write_header_info`:
- `char* header`: the header to write sender application name and its version into
- `bool start`: In case the function is called in `write_header`, the block is written after `type` and `slot.`
- `char* command`: server application name.

As for now, this commit adds `write_header_info` only in the  `pgagroal_write_isalive`  function called if the user specified option `ping` while calling the `pgagroal-cli` command.

The plan is to insert `write_header_info` into every `pgagroal_write_` function called by `main.c`, if the strategy is successful and after code reviews.

### DIFF
--- a/src/libpgagroal/management.c
+++ b/src/libpgagroal/management.c
@@ -53,12 +53,17 @@
 #include <openssl/ssl.h>
 
 #define MANAGEMENT_HEADER_SIZE 5
+#define MANAGEMENT_INFO_SIZE 13
+
+#define S "S:"
+#define V ",V:"
 
 static int read_complete(SSL* ssl, int socket, void* buf, size_t size);
 static int write_complete(SSL* ssl, int socket, void* buf, size_t size);
 static int write_socket(int socket, void* buf, size_t size);
 static int write_ssl(SSL* ssl, void* buf, size_t size);
 static int write_header(SSL* ssl, int fd, signed char type, int slot);
+static int write_info(char* buffer, int command, int offset);
 
 static int pgagroal_management_write_conf_ls_detail(int socket, char* what);
 static int pgagroal_management_read_conf_ls_detail(SSL* ssl, int socket, char* buffer);
@@ -71,10 +76,15 @@ static cJSON* pgagroal_managment_json_read_config_get(int socket, char* config_k
 static cJSON* pgagroal_management_json_read_conf_ls(SSL* ssl, int socket);
 static int pgagroal_management_json_print_conf_ls(cJSON* json);
 
+static int pgagroal_executable_version_number(char* version, size_t version_size);
+static int pgagroal_executable_version_string(char** version_string, int version_number);
+static char* pgagroal_executable_name(int command);
+
 int
 pgagroal_management_read_header(int socket, signed char* id, int32_t* slot)
 {
    char header[MANAGEMENT_HEADER_SIZE];
+   char buf_info[MANAGEMENT_INFO_SIZE];
 
    if (read_complete(NULL, socket, &header[0], sizeof(header)))
    {
@@ -85,6 +95,13 @@ pgagroal_management_read_header(int socket, signed char* id, int32_t* slot)
 
    *id = pgagroal_read_byte(&(header));
    *slot = pgagroal_read_int32(&(header[1]));
+
+   if (read_complete(NULL, socket, &buf_info[0], sizeof(buf_info)))
+   {
+      pgagroal_log_warn("pgagroal_management_read_header: %d %s", socket, strerror(errno));
+      errno = 0;
+      goto error;
+   }
 
    return 0;
 
@@ -188,7 +205,6 @@ pgagroal_management_read_payload(int socket, signed char id, int* payload_i, cha
       case MANAGEMENT_CONFIG_GET:
       case MANAGEMENT_GET_PASSWORD:
       case MANAGEMENT_CONFIG_SET:
-
          if (read_complete(NULL, socket, &buf4[0], sizeof(buf4)))
          {
             goto error;
@@ -687,12 +703,33 @@ pgagroal_management_json_read_status_details(SSL* ssl, int socket, bool include_
    int limits = 0;
    int servers = 0;
    char header[12 + MAX_NUMBER_OF_CONNECTIONS];
+   char buf_info[MANAGEMENT_INFO_SIZE];
+   int application;
+   int version;
 
+   memset(&buf_info, 0, sizeof(buf_info));
    memset(&buf, 0, sizeof(buf));
    memset(&disabled, 0, sizeof(disabled));
    memset(&header, 0, sizeof(header));
 
-   cJSON* json = pgagroal_json_create_new_command_object(include_details ? "status details" :  "status", true, "pgagroal-cli");
+   if (read_complete(ssl, socket, &buf_info[0], sizeof(buf_info)))
+   {
+      pgagroal_log_warn("pgagroal_management_json_read_status_details: read: %d %s", socket, strerror(errno));
+      errno = 0;
+      return NULL;
+   }
+
+   application = pgagroal_read_int32(&(buf_info[2]));
+   version = pgagroal_read_int32(&(buf_info[9]));
+
+   char* version_buf = NULL;
+
+   if (pgagroal_executable_version_string(&version_buf, version))
+   {
+      return NULL;
+   }
+
+   cJSON* json = pgagroal_json_create_new_command_object(include_details ? "status details" :  "status", true, pgagroal_executable_name(application), version_buf);
    cJSON* output = pgagroal_json_extract_command_output_object(json);
 
    if (read_complete(ssl, socket, &buf[0], sizeof(buf)))
@@ -889,6 +926,7 @@ pgagroal_management_json_read_status_details(SSL* ssl, int socket, bool include_
    }
 
 end:
+   free(version_buf);
    return json;
 
 error:
@@ -902,9 +940,23 @@ int
 pgagroal_management_write_status(int socket, bool graceful)
 {
    char buf[16];
+   char buf_info[MANAGEMENT_INFO_SIZE];
+
    int active;
    int total;
    struct main_configuration* config;
+
+   if (write_info(buf_info, PGAGROAL_EXECUTABLE, 0))
+   {
+      goto error;
+   }
+
+   if (write_complete(NULL, socket, &buf_info, sizeof(buf_info)))
+   {
+      pgagroal_log_warn("pgagroal_management_write_status: write: %d %s", socket, strerror(errno));
+      errno = 0;
+      goto error;
+   }
 
    memset(&buf, 0, sizeof(buf));
    active = 0;
@@ -1016,8 +1068,9 @@ error:
 int
 pgagroal_management_write_details(int socket)
 {
-   char header[12 + MAX_NUMBER_OF_CONNECTIONS];
    struct main_configuration* config;
+   int offset = 12;
+   char header[offset + MAX_NUMBER_OF_CONNECTIONS];
 
    config = (struct main_configuration*)shmem;
 
@@ -1030,7 +1083,7 @@ pgagroal_management_write_details(int socket)
    for (int i = 0; i < config->max_connections; i++)
    {
       signed char state = atomic_load(&config->states[i]);
-      header[12 + i] = (char)state;
+      header[offset + i] = (char)state;
    }
 
    if (write_complete(NULL, socket, header, sizeof(header)))
@@ -1130,7 +1183,9 @@ error:
 int
 pgagroal_management_read_isalive(SSL* ssl, int socket, int* status, char output_format)
 {
-   char buf[4];
+   char buf[MANAGEMENT_INFO_SIZE + 4];
+   int application;
+   int version;
 
    memset(&buf, 0, sizeof(buf));
 
@@ -1141,12 +1196,22 @@ pgagroal_management_read_isalive(SSL* ssl, int socket, int* status, char output_
       goto error;
    }
 
-   *status = pgagroal_read_int32(&buf);
+   application = pgagroal_read_int32(&(buf[2]));
+   version = pgagroal_read_int32(&(buf[9]));
+
+   *status = pgagroal_read_int32(&buf[MANAGEMENT_INFO_SIZE]);
 
    // do I need to provide JSON output?
    if (output_format == COMMAND_OUTPUT_FORMAT_JSON)
    {
-      cJSON* json = pgagroal_json_create_new_command_object("ping", true, "pgagroal-cli");
+      char* version_buf = NULL;
+
+      if (pgagroal_executable_version_string(&version_buf, version))
+      {
+         goto error;
+      }
+
+      cJSON* json = pgagroal_json_create_new_command_object("ping", true, pgagroal_executable_name(application), version_buf);
       cJSON* output = pgagroal_json_extract_command_output_object(json);
 
       cJSON_AddNumberToObject(output, "status", *status);
@@ -1179,6 +1244,19 @@ int
 pgagroal_management_write_isalive(int socket, bool gracefully)
 {
    char buf[4];
+   char buf_info[MANAGEMENT_INFO_SIZE];
+
+   if (write_info(buf_info, PGAGROAL_EXECUTABLE, 0))
+   {
+      goto error;
+   }
+
+   if (write_complete(NULL, socket, &buf_info, sizeof(buf_info)))
+   {
+      pgagroal_log_warn("pgagroal_management_write_isalive: write: %d %s", socket, strerror(errno));
+      errno = 0;
+      goto error;
+   }
 
    memset(&buf, 0, sizeof(buf));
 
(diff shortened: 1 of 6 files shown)


src/libpgagroal/management.c
```diff
@@ -1113,10 +1114,12 @@ pgagroal_management_read_isalive(SSL* ssl, int socket, int* status, char output_
 int
 pgagroal_management_write_isalive(int socket, bool gracefully)
 {
-   char buf[4];
+   char buf[MANAGEMENT_HEADER_SIZE];
 
    memset(&buf, 0, sizeof(buf));
 
+   write_header_info(buf, 1, "pgagroal ");
```
fluca1978, line 1121: Second argument of `write_header_info` is a `bool`, so use `true` or `false`.

src/libpgagroal/management.c
```diff
@@ -1618,6 +1642,8 @@ write_header(SSL* ssl, int fd, signed char type, int slot)
    pgagroal_write_byte(&(header), type);
    pgagroal_write_int32(&(header[1]), slot);
 
+   write_header_info(header, 0 ,"pgagroal-cli ");
```
fluca1978, line 1645: Second argument of `write_header_info` is a `bool`, so use `true` or `false`.

src/libpgagroal/management.c
```diff
@@ -1610,6 +1613,27 @@ write_ssl(SSL* ssl, void* buf, size_t size)
    return 1;
 }
 
+static int
+write_header_info(char* header, bool start, char* command) {
+   char header_info[42] = " {sender: ";
```
fluca1978, line 1618: I would remove the space after the colon, is it useful?

src/libpgagroal/management.c
```diff
@@ -1610,6 +1613,27 @@ write_ssl(SSL* ssl, void* buf, size_t size)
...
+
+   strcat(header_info, command);
+   strcat(header_info, "version: ");
```
fluca1978, line 1621: I would remove the space after the colon, is it useful?

src/libpgagroal/management.c
```diff
@@ -1610,6 +1613,27 @@ write_ssl(SSL* ssl, void* buf, size_t size)
...
+   strcat(header_info, PGAGROAL_VERSION);
+   strcat(header_info, "}");
+
+   if (start) {
```
fluca1978, line 1625: uncrustify

src/libpgagroal/management.c
```diff
@@ -1610,6 +1613,27 @@ write_ssl(SSL* ssl, void* buf, size_t size)
...
+   else {
+      pgagroal_write_string(header + 4, header_info);
+   }
+
+   pgagroal_log_debug("Header: %s\n", header);
+
+   return 0;
+}
+
 static int
```
fluca1978, line 1637: Please add documentation about the function.

fluca1978 (changes requested):
I cannot get this working, even if it compiles:

```
% pgagroal  
INFO  main.c:1094 pgagroal: 1.7.0 started on *:54322
zsh: segmentation fault (core dumped)  pgagroal
```

Tested on Rocky Linux 9.

I also wonder if `write_header_info` should be placed directly into `wirte_header` and `read_header` in order to get it for free on every command.

src/libpgagroal/management.c
EuGig: Th spaces give a debug message like this one:
`2024-03-18 17:47:16 DEBUG management.c:1635 Header:  {sender: pgagroal version: 1.7.0}
`

src/libpgagroal/management.c
EuGig: Th spaces give a debug message like this one:
`2024-03-18 17:47:16 DEBUG management.c:1635 Header:  {sender: pgagroal version: 1.7.0}
`

src/libpgagroal/management.c
EuGig: will do

src/libpgagroal/management.c
EuGig: will do

src/libpgagroal/management.c
fluca1978: > Th spaces give a debug message like this one: `2024-03-18 17:47:16 DEBUG management.c:1635 Header: {sender: pgagroal version: 1.7.0} `

I see, but we should consider the header size as first, not the header human readability.
I'm not sure we want to send out an header longer than the payload, so I would rather add a single character for the indication of the fields, like `S:pgagroal;V:1.7.0` or even converting the version number into a digit (e.g., 1.7.0 could become 107000 and be sent as an int32).

src/libpgagroal/management.c
fluca1978: Also note there's an extra space in `"pgagroal -cli"`

src/libpgagroal/management.c
EuGig: > I see, but we should consider the header size as first, not the header human readability.
As for now, the header has: 1 byte for `type`, 4 bytes for `slot` and 18-22 bytes in case the application name is `pgagroal` or `pgagroal-cli` respectively. Converting the version to int32 saves us 1 byte (if I'm not mistaken).

EuGig: > I see, but we should consider the header size as first, not the header human readability.

As for now, the header has: 1 byte for type, 4 bytes for slot and 18-22 bytes in case the application name is pgagroal or pgagroal-cli respectively. Converting the version to int32 saves us 1 byte (if I'm not mistaken).

fluca1978:
> `1637 pgagroal_log_debug("Header: %s\n", header);` could be the cause of the segmentation fault, since `header` has also `type` and `slot` which are not strings.
> 

No it's not, even if that line could lead to other problems.
The line:
```
   else {
      pgagroal_write_string(header + 4, header_info);
   }
```

is the cause, and it's probably because `hedaer + 4` is wrong (not allocated when you expect it). I've digged anymore to understand why, but it's not clear to me why you are writing the header in different positions.


> Oh yeah, calling `write_header_info` inside `read-header` can save us some coding. Though, can calling a "writing" function inside another function which is supposed to only read something maybe be a bit confusing?

No, there is some confusion originating from me: `write_header` already calls `write_header_info`, which is what I would expect. I don't get why you should call it also from `read_header`.

EuGig:
> I've digged anymore to understand why, but it's not clear to me why you are writing the header in different positions.

The problem is that `write_header` writes at the beginning of the header the `type` and the `slot`, i.e., 5 bytes. So to not loose this information I thought to write starting from header + 4. However, when pgagroal writes in the socket, it seems it does not include this information anymore.

fluca1978:
> > No, there is some confusion originating from me: write_header already calls write_header_info, which is what I would expect. I don't get why you should call it also from read_header.
> 
> I also got confused. After reflecting a bit, this is what I think can work:
> 
>     1. cli.c calls a function based on a specified option, say `isalive`, which calls `write_header` (info about pgagroal-cli are written in the header)
> 
>     2. main.c reads the header through `pgagroal_management_read_header` and prints for debugging purposes info about pgagraol-cli.
> 
>     3. main.c writes the respond through `pgagroal_write_isalive` (info about pgagroal are written in the header)
> 
>     4. cli.c reads the header through `pgagroal_read_isalive` and prints for debugging purposes info about pgagroal.
> 
> 

The above is correct. The fact that the header is written only from one side is why, initially, I thought that placing the information into the payload was simpler than placing into an header. In other words, the `write_info_header` function is inserting something in the header when the `pgagroal-cli` requests a command, but not when `pgagroal` is answering.


> So, maybe the plan can be summarised as follows:
> 
>     * `write_header_info` must be called inside `write_header` and `pgagroal_write_isalive`.
> 

Correct.

>     * `pgagroal_log_debug` must be called inside `pgagroal_management_read_header` and `pgagroal_management_read_isalive`.
> 

No. The purpose of this work is not to debug who is sending a message, but giving the counterpart to be able to understand who sent the message. While you can add as much logging as you want, the final aim is to make `pgagroal-cli` and `pgagroal` able to understand who originated the message they are handling.


> Maybe I'm missing something and maybe we don't need to insert `write_header_info` inside every `pgagroal_write_*` function. Does that make sense to you?

I think we need to add into every write function, because we need to provide such information in both directions.

EuGig:
> In other words, the write_info_header function is inserting something in the header when the pgagroal-cli requests a command, but not when pgagroal is answering.

As a matter of fact, I'm finding problems in adding `write_header_info` inside `pgagroal_management_write_status` and `pgagroal_management_write_details`. Actually, this two functions write something in the header, so one has to play with offsets a little bit too much. Accordingly, one has to change offsets in `pgagroal_management_read_status` and `pgagroal_management_read_status_details` too.

jesperpedersen:
You could look into expanding the `pgagroal_append()` infrastructure and use that as the base

fluca1978:
> > In other words, the write_info_header function is inserting something in the header when the pgagroal-cli requests a command, but not when pgagroal is answering.
> 
> As a matter of fact, I'm finding problems in adding `write_header_info` inside `pgagroal_management_write_status` and `pgagroal_management_write_details`. Actually, this two functions write something in the header, so one has to play with offsets a little bit too much. Accordingly, one has to change offsets in `pgagroal_management_read_status` and `pgagroal_management_read_status_details` too.


@EuGig did you manage to add the information header in all the other cases? We can study better the `status` and `details` commands.

jesperpedersen:
Header should be in the start of the payload for sure

EuGig:
Just rebased after `5dcfb897d90b78c622a38855ae6ffd5cb2630d9f` fixing [#407].

jesperpedersen:
You need to rebase

https://github.com/progit/progit2/releases

jesperpedersen:
You need to add your name to `AUTHORS` too

EuGig:
Thanks, will do!

jesperpedersen:
See build failure...

jesperpedersen:
https://github.com/agroal/pgagroal/blob/master/contrib/valgrind/README.md

src/libpgagroal/management.c
```diff
@@ -1698,15 +1798,163 @@ write_ssl(SSL* ssl, void* buf, size_t size)
...
+{
+   version_name[0] = ((version_number / 100000) % 10) + '0';
+   version_name[1] = '.';
+   version_name[2] = ((version_number / 1000) % 10) + '0';
+   version_name[3] = '.';
+   version_name[4] = ((version_number / 10) % 10) + '0';
+   version_name[5] = '\0';
+
+   return 0;
+}
```
EuGig, line 1922: I needed this function to add a string into the json output. I know It's kind of ugly.

fluca1978:
I see you added a few piece that looks like:
```
 if (version_buf != NULL)
   {
      pgagroal_executable_version_name(version_buf, version);

      cJSON* app_json = cJSON_CreateObject();
      cJSON_AddItemToObject(output, "application", app_json);
      cJSON_AddStringToObject(app_json, "name", pgagroal_executable_name(application));
      cJSON_AddStringToObject(app_json, "version", version_buf);
      free(version_buf);
   }
```

but we already have `pgagroal_json_create_new_command_object` that accepts the executable name, so I think it would be better to extend such prototype to include also the application version, so you need to extract such information first from the buffer and then create the JSON object with all the stuff altready packed. Note that `pgagroal_json_create_new_command_object` already has the "application" and "version" entries, so you need only to customize their values.


Moreover, I'm unable to compile:
```
% make clean && cmake .. && make && sudo make install
-- pgagroal 1.7.0
-- Build type is Release
-- System is Linux
-- libev found
-- OpenSSL found
-- rst2man found
-- cJSON found version 1.7.17
-- libatomic found
-- systemd found
-- Configuring done
-- Generating done
-- Build files have been written to: /home/luca/pgagroal/build
[  0%] Built target man
[  3%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/configuration.c.o
[  7%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/json.c.o
[ 10%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/logging.c.o
[ 14%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/management.c.o
/home/luca/pgagroal/src/libpgagroal/management.c: In function ‘pgagroal_management_json_read_status_details’:
/home/luca/pgagroal/src/libpgagroal/management.c:1919:20: error: array subscript 5 is outside array bounds of ‘char[5]’ [-Werror=array-bounds]
 1919 |    version_name[5] = '\0';
      |    ~~~~~~~~~~~~~~~~^~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:739:31: note: referencing an object of size 5 allocated by ‘malloc’
  739 |    char* version_buf = (char*)malloc(5 * sizeof(char));
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c: In function ‘pgagroal_management_read_isalive’:
/home/luca/pgagroal/src/libpgagroal/management.c:1919:20: error: array subscript 5 is outside array bounds of ‘char[5]’ [-Werror=array-bounds]
 1919 |    version_name[5] = '\0';
      |    ~~~~~~~~~~~~~~~~^~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:1220:38: note: referencing an object of size 5 allocated by ‘malloc’
 1220 |           char* version_buf = (char*)malloc(5 * sizeof(char));
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c: In function ‘pgagroal_management_read_config_get’:
/home/luca/pgagroal/src/libpgagroal/management.c:1919:20: error: array subscript 5 is outside array bounds of ‘char[5]’ [-Werror=array-bounds]
 1919 |    version_name[5] = '\0';
      |    ~~~~~~~~~~~~~~~~^~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2134:31: note: referencing an object of size 5 allocated by ‘malloc’
 2134 |    char* version_buf = (char*)malloc(5 * sizeof(char));
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c: In function ‘pgagroal_management_read_conf_ls’:
/home/luca/pgagroal/src/libpgagroal/management.c:1919:20: error: array subscript 5 is outside array bounds of ‘char[5]’ [-Werror=array-bounds]
 1919 |    version_name[5] = '\0';
      |    ~~~~~~~~~~~~~~~~^~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2751:31: note: referencing an object of size 5 allocated by ‘malloc’
 2751 |    char* version_buf = (char*)malloc(5 * sizeof(char));
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2857:4: error: ‘buffer’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 2857 |    free(buffer);
      |    ^~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2735:10: note: ‘buffer’ was declared here
 2735 |    char* buffer;
      |          ^~~~~~
cc1: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/pgagroal.dir/build.make:118: src/CMakeFiles/pgagroal.dir/libpgagroal/management.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:254: src/CMakeFiles/pgagroal.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```


Last, there is the need to uncrustify the code: there are places where the code is not aligned and other lines that present only  a tab character.

Can you please have a look at all the above?

Thanls

EuGig:
Hi @fluca1978!

>  but we already have pgagroal_json_create_new_command_object that accepts the executable name, so I think it would be better to extend such prototype to include also the application version, so you need to extract such information first from the buffer and then create the JSON object with all the stuff altready packed. Note that pgagroal_json_create_new_command_object already has the "application" and "version" entries, so you need only to customize their values.

The problem with `pgagroal_json_create_new_command_object` is that it comes not only with `application` and `version` objects, but also with the `command` and the `output` object. So, if I add another json object through `pgagroal_json_create_new_command_object` I got something like this:

```
{
	"command":	{
		"name":	"conf ls",
		"status":	"OK",
		"error":	0,
		"exit-status":	0,
		"output":	{
			"application":	{
				"command":	{
					"name":	"conf ls",
					"status":	"OK",
					"error":	0,
					"exit-status":	0,
					"output":	{
					}
				},
				"application":	{
					"name":	"pgagroal",
					"major":	1,
					"minor":	7,
					"patch":	0,
					"version":	"1.7.0"
				}
			},
			"files":	{
				"list":	[{
						"description":	"Main Configuration file",
						"path":	"pgagroal.conf"
					}, {
						"description":	"HBA File",
						"path":	"pgagroal_hba.conf"
					}, {
						"description":	"Limit file",
						"path":	""
					}, {
						"description":	"Frontend users file",
						"path":	""
					}, {
						"description":	"Admins file",
						"path":	""
					}, {
						"description":	"Superuser file",
						"path":	""
					}, {
						"description":	"Users file",
						"path":	""
					}]
			}
		}
	},
	"application":	{
		"name":	"pgagroal-cli",
		"major":	1,
		"minor":	7,
		"patch":	0,
		"version":	"1.7.0"
	}
}
```

(For simplicity I put it into the `output` entry, since it is exposed with the `pgagroal_json_extract_command_output_object` function). So, we should expand this prototype to include the application-version, and also teach it to not print the `command` entry in case we don't want to repeat it.

fluca1978:
I meant to use `pgagroal_json_create_new_command_object` **after** you dereference the application and version.
For example, in `pgagroal_management_json_read_conf_ls` you have:

```
   cJSON* json = pgagroal_json_create_new_command_object("conf ls", true, "pgagroal-cli");
   cJSON* output = pgagroal_json_extract_command_output_object(json);
   
   if (read_complete(ssl, socket, &buf_info[0], sizeof(buf_info)))
   {
      goto error;
   }

   application = pgagroal_read_int32(&(buf_info[2]));
   version = pgagroal_read_int32(&(buf_info[9]));

   char* version_buf = (char*)malloc(5 * sizeof(char));
   if (version_buf != NULL)
   {  ... }
```

that should be like:

```
   
   if (read_complete(ssl, socket, &buf_info[0], sizeof(buf_info)))
   {
      goto error;
   }

   application = pgagroal_read_int32(&(buf_info[2]));
   version = pgagroal_read_int32(&(buf_info[9]));

   cJSON* json = pgagroal_json_create_new_command_object("conf ls", true, application, version);  // <- HERE !
   cJSON* output = pgagroal_json_extract_command_output_object(json);

```

so you need to add the version parameter to the `pgagroal_json_create_command_object` and place the appropriate arguments.
What do you think?

EuGig:
Oh I see! Sounds good.

fluca1978:
I still cannot compile:

```
% make clean && cmake .. && make && sudo make install
-- pgagroal 1.7.0
-- Build type is Release
-- System is Linux
-- libev found
-- OpenSSL found
-- rst2man found
-- cJSON found version 1.7.17
-- libatomic found
-- systemd found
-- Configuring done
-- Generating done
-- Build files have been written to: /home/luca/pgagroal/build
[  0%] Built target man
[  3%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/configuration.c.o
[  7%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/json.c.o
[ 10%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/logging.c.o
[ 14%] Building C object src/CMakeFiles/pgagroal.dir/libpgagroal/management.c.o
/home/luca/pgagroal/src/libpgagroal/management.c: In function ‘pgagroal_management_json_read_status_details’:
/home/luca/pgagroal/src/libpgagroal/management.c:935:4: error: ‘json’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  935 |    pgagroal_json_set_command_object_faulty(json, strerror(errno), errno);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c: In function ‘pgagroal_management_read_conf_ls’:
/home/luca/pgagroal/src/libpgagroal/management.c:2825:4: error: ‘json’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 2825 |    pgagroal_json_set_command_object_faulty(json, strerror(errno), errno);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2730:11: note: ‘json’ was declared here
 2730 |    cJSON* json = pgagroal_json_create_new_command_object("conf ls", true, pgagroal_executable_name(application), version_buf);
      |           ^~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2828:4: error: ‘version_buf’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 2828 |    free(version_buf);
      |    ^~~~~~~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2721:10: note: ‘version_buf’ was declared here
 2721 |    char* version_buf = (char*)malloc(6 * sizeof(char));
      |          ^~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2822:4: error: ‘buffer’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 2822 |    free(buffer);
      |    ^~~~~~~~~~~~
/home/luca/pgagroal/src/libpgagroal/management.c:2708:10: note: ‘buffer’ was declared here
 2708 |    char* buffer;
      |          ^~~~~~
cc1: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/pgagroal.dir/build.make:118: src/CMakeFiles/pgagroal.dir/libpgagroal/management.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:254: src/CMakeFiles/pgagroal.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

EuGig:
Oh sorry, I fixed this error. Though, I wonder why my compiler didn't see this.

fluca1978:
Looks fine to me.
@jesperpedersen  PTAL commit d7c6409

jesperpedersen:
@EuGig Looks at the CI failures

fluca1978:
Does compile on gcc 11.4, interesting there are failures.
@EuGig it should suffice to keep the values as `long` when calling `strtol` and cast them as integers when needed at the last time.

jesperpedersen:
Merged.

PR #418, opened 2024-03-20
[#416] Improve `pgagroal-admin` warning messages

### LINKED ISSUE #416: Change `warnx` in `pgagroal-admin` to point at the right file
In `pgagroal-admin` there are two places (<https://github.com/agroal/pgagroal/blob/master/src/admin.c#L390> and <https://github.com/agroal/pgagroal/blob/master/src/admin.c#L370>) where a `warnx` message is printed pointing the user to `~/.pgagroal` file.
This is not consistent, since the file could be somewhere else, according to `pgagroal_get_home_directory()` and the value of `buf[0]` as correctly reported here <https://github.com/agroal/pgagroal/blob/master/src/admin.c#L443>).

fluca1978:
Improves some messages about the `master-key` command when dealing with wrong file permissions: it clearly indicates the directory or the file that has wrong permissions.

Close #416

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -285,7 +285,7 @@ main(int argc, char** argv)
    {
       if (master_key(password, generate_pwd, pwd_length))
       {
-         errx(1, "Error for master key");
+         errx(1, "Cannot generate master key");
       }
    }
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

Issue #422, opened 2024-03-25
Improve the `restart_server` function


fluca1978:
The `restart_server` function is responbile to understand if the changes in the configuration require for a restart.
There is a *TODO* item left for evaluating the TLS configuration <https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L2527> that need to be fixed.

PR #423, opened 2024-03-25
[#422] Improve the `restart_server` function

T1t4m1un:
### Functionality:

Implement `restart_server` function for daemon with conditional restarts and warnings for configuration changes, including TLS updates.

### Document:

```c
/**
 * Compares the TLS configurations of two servers to determine if they are identical.
 *
 * This function assesses whether the source and destination servers have matching TLS configurations, 
 * which is crucial for ensuring secure communication protocols are uniformly applied across server interactions.
 *
 * @param s1 Pointer to the first server structure to compare.
 * @param s2 Pointer to the second server structure to compare.
 * @return Returns true if both servers have identical TLS configurations, false if they differ.
 */
static bool is_same_tls(struct server* s1, struct server* s2);

/**
 * Checks if two boolean values are same. If not, prints a line in the log
 *
 * @param name The name of the configuration option.
 * @param e The existing boolean value.
 * @param n The new boolean value.
 * @return 1 if a restart is required, 0 otherwise.
 */
static int restart_bool(char* name, bool e, bool n);
```

### Test Case Verification:

1. Follow the documentation to configure the pgagroal daemon and pgagroal-cli using the provided demo 
2. Initiate the pgagroal 
3. Update the pgagroal.conf file under the [primary] section to enable TLS by setting tls = on
4. Execute pgagroal-ctl conf reload to apply the new 
5. Confirm the restart requirement in /tmp/pgagroal.log, which should log an entry similar to:

```
2024-03-25 16:55:04 INFO  configuration.c:2479 Restart required for Server <primary>, parameter <tls> - Existing false New true
```

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -79,12 +79,14 @@ static void copy_server(struct server* dst, struct server* src);
 static void copy_hba(struct hba* dst, struct hba* src);
 static void copy_user(struct user* dst, struct user* src);
 static int restart_int(char* name, int e, int n);
+static int restart_bool(char* name, bool e, bool n);
 static int restart_string(char* name, char* e, char* n, bool skip_non_existing);
 static int restart_limit(char* name, struct configuration* config, struct configuration* reload);
 static int restart_server(struct server* src, struct server* dst);
 
 static bool is_empty_string(char* s);
 static bool is_same_server(struct server* s1, struct server* s2);
+static bool is_same_tls(struct server* s1, struct server* s2);
 
 static bool key_in_section(char* wanted, char* section, char* key, bool global, bool* unknown);
 static bool is_comment_line(char* line);
@@ -2395,6 +2397,26 @@ is_same_server(struct server* s1, struct server* s2)
    }
 }
 
+/**
+ * Checks if TLS configurations are same.
+ * @return true if the TLS configurations are same
+*/
+static bool
+is_same_tls(struct server* src, struct server* dst)
+{
+   if (src->tls == dst->tls &&
+       !strncmp(src->tls_cert_file, dst->tls_cert_file, MISC_LENGTH) &&
+       !strncmp(src->tls_key_file, dst->tls_key_file, MISC_LENGTH) &&
+       !strncmp(src->tls_ca_file, dst->tls_ca_file, MISC_LENGTH))
+   {
+      return true;
+   }
+   else
+   {
+      return false;
+   }
+}
+
 static void
 copy_server(struct server* dst, struct server* src)
 {
@@ -2440,6 +2462,22 @@ restart_int(char* name, int e, int n)
    if (e != n)
    {
       pgagroal_log_info("Restart required for %s - Existing %d New %d", name, e, n);
+      return 1;
+   }
+
+   return 0;
+}
+
+/**
+ * Utility function prints a line in the log when a restart is required.
+ * @return 0 when parameter values are same, 1 when a restart required.
+*/
+static int
+restart_bool(char* name, bool e, bool n)
+{
+   if (e != n)
+   {
+      pgagroal_log_info("Restart required for %s - Existing %s New %s", name, e ? "true" : "false", n ? "true" : "false");
       return 1;
    }
 
@@ -2524,7 +2562,18 @@ restart_server(struct server* src, struct server* dst)
       restart_string(restart_message, dst->host, src->host, false);
       snprintf(restart_message, sizeof(restart_message), "Server <%s>, parameter <port>", src->name);
       restart_int(restart_message, dst->port, src->port);
-      /* TODO - TLS */
+      return 1;
+   } 
+   else if (!is_same_tls(src, dst))
+   {
+      snprintf(restart_message, sizeof(restart_message), "Server <%s>, parameter <tls>", src->name);
+      restart_bool(restart_message, dst->tls, src->tls);
+      snprintf(restart_message, sizeof(restart_message), "Server <%s>, parameter <tls_cert_file>", src->name);
+      restart_string(restart_message, dst->tls_cert_file, src->tls_cert_file, false);
+      snprintf(restart_message, sizeof(restart_message), "Server <%s>, parameter <tls_key_file>", src->name);
+      restart_string(restart_message, dst->tls_key_file, src->tls_key_file, false);
+      snprintf(restart_message, sizeof(restart_message), "Server <%s>, parameter <tls_ca_file>", src->name);
+      restart_string(restart_message, dst->tls_ca_file, src->tls_ca_file, false);
       return 1;
    }
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,3 +10,4 @@ Lawrence Wu <lawrence910426@gmail.com>
 Yongting You <2010youy01@gmail.com>
 Ashutosh Sharma <ash2003sharma@gmail.com>
 Henrique de Carvalho <decarv.henrique@gmail.com>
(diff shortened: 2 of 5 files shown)


src/libpgagroal/configuration.c
```diff
@@ -2395,6 +2397,29 @@ is_same_server(struct server* s1, struct server* s2)
...
+   {
+      return false;
+   }
+
+   if (!src->tls)
+   {
+      return true;
+   }
+
+   if (strcmp(src->tls_cert_file, dst->tls_cert_file) ||
```
jesperpedersen, line 2413: You should reverse this check

src/libpgagroal/configuration.c
```diff
@@ -2395,6 +2397,29 @@ is_same_server(struct server* s1, struct server* s2)
...
+       strcmp(src->tls_key_file, dst->tls_key_file) ||
+       strcmp(src->tls_ca_file, dst->tls_ca_file))
+   {
+      return false;
+   }
+
+   return true;
```
jesperpedersen, line 2420: `return false;` should be default

jesperpedersen (changes requested):

src/libpgagroal/configuration.c
T1t4m1un: Yes I think you are right. In order to make the code style more consistent, I changed the name and logic for `is_same_tls`. The original logic is the opposite, but I forgot to change this return value😂🥺

src/libpgagroal/configuration.c
T1t4m1un: I'm sorry I wrote unclear code (just for cleaness). I will modify here to improve readability.

But actually strcmp will return 0 only if two strings are same, and every nonzero value will be considered to be true🥺

Could you reconsider if my code is wrong?🥺 Thank you so much and apologize again for its confusion🥺

T1t4m1un: https://pastebin.ubuntu.com/p/Dgd55gBSZW/
I wrote a unittest for this function. The code behavior seems consistent with my expectations. Could you review my unittest to see is any bias on my understanding to this feature? Thank you very much!

Or if you are concerning about line 2410? In this line, I think, if we don't use `tls` at all, there's no need to compare the rest of the config. So we can directly return true.

Additionally I found in our project all of the strcmp is used like mine. So could you give me a little further suggestions?

fluca1978:
@T1t4m1un  the `is_same_tls` function has the right name, what Jeper was suggesting is to make the code more robust by reversing the condition checks: test explicitly if the tls settings are the same, then return true, otherwise return false.
Your implementation at the moment does the opposite: checks if the tls are not the same, otherwise returns (by default) true.

With regard to `strcmp`, I know we are using `strcmp` in many places, but we should start using `strncmp` for safety reasons.

T1t4m1un:
> @T1t4m1un the `is_same_tls` function has the right name, what Jeper was suggesting is to make the code more robust by reversing the condition checks: test explicitly if the tls settings are the same, then return true, otherwise return false. Your implementation at the moment does the opposite: checks if the tls are not the same, otherwise returns (by default) true.
> 
> With regard to `strcmp`, I know we are using `strcmp` in many places, but we should start using `strncmp` for safety reasons.

Thank you very much for your patient reply! Please take a look my latest commit🥺

fluca1978:
Please reverse the `true` and `false` value checks as Jesper asked, then add the note about the need for restart into the documentation and your name to the authors file.

T1t4m1un:
> Please reverse the `true` and `false` value checks as Jesper asked, then add the note about the need for restart into the documentation and your name to the authors file.

Thank you for your recognition of me!

I'm editing the documentation but have some specific features want to confirm with you.

In our current implementation, we only check the config in server section. We won't check whether the host name or TLS config is same as former one in the `[pgagroal] section`. Is this the feature we expected?

```bash
# provided config same as `getting started`
pgagroal-cli conf set tls on # success, we won't check the config not in server section
pgagroal-cli conf set primary.tls on # failed, we do check the config in server `primary` section
```

image

Thank you for your answer!

jesperpedersen:
The TLS in the `[pgagroal]` is to secure pgagroal itself, so "server side".

The TLS in the server sections are to support connecting to secure PostgreSQL instances.

T1t4m1un:
> The TLS in the `[pgagroal]` is to secure pgagroal itself, so "server side".
> 
> The TLS in the server sections are to support connecting to secure PostgreSQL instances.

Thank you for your answer! I emphasized it in my documentation. Looking forward to your review~

src/libpgagroal/configuration.c
```diff
@@ -2400,15 +2400,17 @@ is_same_server(struct server* s1, struct server* s2)
...
-   if (src->tls != dst->tls ||
-       strncmp(src->tls_cert_file, dst->tls_cert_file, MISC_LENGTH)||
-       strncmp(src->tls_key_file, dst->tls_key_file, MISC_LENGTH)||
-       strncmp(src->tls_ca_file, dst->tls_ca_file, MISC_LENGTH))
+   if (src->tls == dst->tls ||
+       !strncmp(src->tls_cert_file, dst->tls_cert_file, MISC_LENGTH)||
+       !strncmp(src->tls_key_file, dst->tls_key_file, MISC_LENGTH)||
+       !strncmp(src->tls_ca_file, dst->tls_ca_file, MISC_LENGTH))
+   {
+      return false;
```
fluca1978, line 2408: `return true;` ?

fluca1978 (commented):
Seems you have always `is_same_tls` returning `false`.

src/libpgagroal/configuration.c
T1t4m1un: 🥹🥹sorry I made a mistake

src/libpgagroal/configuration.c
```diff
@@ -2400,19 +2400,10 @@ is_same_server(struct server* s1, struct server* s2)
...
-      return true;
-   }
-
-   if (strcmp(src->tls_cert_file, dst->tls_cert_file) ||
-       strcmp(src->tls_key_file, dst->tls_key_file) ||
-       strcmp(src->tls_ca_file, dst->tls_ca_file))
+   if (src->tls != dst->tls ||
+       strncmp(src->tls_cert_file, dst->tls_cert_file, MISC_LENGTH)||
+       strncmp(src->tls_key_file, dst->tls_key_file, MISC_LENGTH)||
+       strncmp(src->tls_ca_file, dst->tls_ca_file, MISC_LENGTH))
```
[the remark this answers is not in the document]
T1t4m1un: SORRY it seems there's something wrong with my GitHub notification, I didn't see your review before 🥹🥹 I have no means to do it🥹

src/libpgagroal/configuration.c
```diff
@@ -2446,6 +2471,18 @@ restart_int(char* name, int e, int n)
    return 0;
 }
 
+static int
```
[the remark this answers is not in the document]
T1t4m1un: Because all the exsisting `restart_XXX` functions return `int`. It's for code style consistency.

What your opinion about it?

fluca1978:
> Sorry for such careless mistake. I've fixed it and add some short comments.
> 

No problem, but better test your code before pushing for review.

> I haven't change the return type of `restart_bool`, I'm looking forward for your opinion on it:
> 

The idea should be this: the `is_same_xxx` functions return a `bool`, the `restart_xxx` functions return and int.


Please, meld all the commits into a single one and force push (or push to another branch) so that we can get the last look at it.

fluca1978:
@jesperpedersen PTAL seems fine to me

jesperpedersen:
Rebased, and merged.

Issue #419, opened 2024-03-23
Allow `pgagroal-admin` to specify where to store the master key password file


fluca1978:
When `pgagroal-admin` is invoked to generate a master key, i.e., with the `master-key` command, it generates the file into `$HOME/.pgagroal/master,key`, and such location is not configurable.
Users could require to store the *private `pgagroal`* stuff into another location, for example `~/.config~.

There is therefore the need for `pgagroal-admin` to have a command line option for the `master-key` command to specifiy where to put (and hence, read from) the master key file.

jesperpedersen:
The management TLS files are also loaded from there...

fluca1978:
@palak-chaturvedi is working on this

palak-chaturvedi:
[v001-Adding_File_Name.patch](https://github.com/agroal/pgagroal/files/14758729/v001-Adding_File_Name.patch)
I worked on this and made the changes.
I just wanted to know if there is a specific code formatter that I have to run?

palak-chaturvedi:
@fluca1978 @jesperpedersen 
https://github.com/palak-chaturvedi/pgagroal/tree/addingFileName

This a branch in my github.

palak-chaturvedi:
Ok I will
Can I store the required location in a variable or we need an argument for it?
@fluca1978

fluca1978:
> Ok I will Can I store the required location in a variable or we need an argument for it? @fluca1978


You are free to propose the solution you think is the best, and can also improve it later after a review.
One optimal approach, but longer, could be to introduce a configuration tunable for the location to where store the files.

But let's go with calm and see your idea first.

palak-chaturvedi:
created a pull request.
@fluca1978 
https://github.com/agroal/pgagroal/pull/430

PR #430, opened 2024-03-27
Issue #419 Allow pgagroal-admin to specify where to store the master key password file

palak-chaturvedi:
Added the filename attribute to master-key generation.
To tackle the multiple uses of file in various location, stored the location of the file in a text file.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -55,25 +55,27 @@
 #define ACTION_REMOVE_USER 4
 #define ACTION_LIST_USERS  5
 
+
 static char CHARS[] = {'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
                        'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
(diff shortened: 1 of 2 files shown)


fluca1978:
First of all, it seems you need to rebase your code.
Second, I don't like the idea of having `master_key.txt` to hold the path of the master key file location.
Moreover, the master key file is used for other operations, so we need to be able to pass to other commands.
One possible solution could be to use  a command line flag `pgagroal-admin -m /path/to/blah/foo/master.key  master-key` so that it is also possible then  `pgagroal-admin -m /path/to/blah/foo/master.key  used add`.
This would be only a partial solution, in any case, since the master key file is used also in other source files, so the flag has to be then added to (at least I think), `main.c`.
Another idea, could be to allow for a tunable in `pgagroal.conf` that indicates a possible alternative directory for the master key file, so that the user is not asked to type the location but the application knows where to find the master key outside of `~/.pgagroal`.

palak-chaturvedi:
The master.key filed is called in only two files.  I changed the "master_key" function as well as the "pgagroal_get_master_key" function in security.c. 
I can add arguments but I thought if the user does not write the key in arguments but it has been created in a location other than default then if the location is not stored somewhere the function would throw error.
I also thought txt file is not the best option (adding it to conf file is an option but the conf file cannot be edited while creating the master_key to store the location. The location would be predefined by the user).
Instead of it arguments options is much better I think. 
What are your views on this?
@fluca1978

fluca1978:
Sorry, I don't like the idea of having a text file storing a "pointer" to the master key file.
I think we should either add a command line flag (e.g., `-m`) to all the executables that can use the master key file or add a tunable in the configuration file `pgagroal.conf`. Clearly, the latter has to be in place **before** `pgagroal-admin master-key` is run, in other words, it is not `pgagroal-admin` supposed to change the configuration by itself.

PR #431, opened 2024-03-28
Fix exit status when parsing error occurs in pgagroal-cli

decarv:
While experimenting in https://github.com/agroal/pgagroal/pull/427, I found out that `pgagroal-cli` returns an exit status of 0, indicating success to the caller, if a command parsing error occurs.

This commit resolves the issue by setting `exit_error = 1;` after a parsing error, ensuring it properly returns the correct error status to the caller.

Notice that the same does not occur with `pgagroal-admin` (it has a different way of handling errors -- `goto error`, while `pgagroal-cli` goes to the same label  --`done`).

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -686,6 +686,7 @@ main(int argc, char** argv)
    if (!parse_command(argc, argv, optind, &parsed, command_table, command_count))
    {
       usage();
+      exit_code = 1;
       goto done;
    }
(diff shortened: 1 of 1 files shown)


fluca1978:
Looks fine to me.
@jesperpedersen  PTAL

jesperpedersen:
Rebased, and merged.

PR #435, opened 2024-04-01
Adding filename to master key generation Issue #419

palak-chaturvedi:
Updated with respect to the new configuration.c changes.

- Can add -c `conf-file` to the master-key generation
`pgagroal-admin -c conf-file -g master-key`
- The pgagroal_get_master_key() is also updated to access the master_key_file_location from the conf file.

@jesperpedersen @fluca1978

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -31,6 +31,8 @@
 #include <logging.h>
 #include <security.h>
 #include <utils.h>
+#include <shmem.h>
+#include <configuration.h>
 
 /* system */
 #include <ctype.h>
@@ -52,9 +54,9 @@
 #define ACTION_REMOVE_USER 4
 #define ACTION_LIST_USERS  5
 
-static int master_key(char* password, bool generate_pwd, int pwd_length);
-static int add_user(char* users_path, char* username, char* password, bool generate_pwd, int pwd_length);
-static int update_user(char* users_path, char* username, char* password, bool generate_pwd, int pwd_length);
+static int master_key(char* password, bool generate_pwd, int pwd_length, char* filename);
+static int add_user(char* users_path, char* username, char* password, bool generate_pwd, int pwd_length, char* filename);
+static int update_user(char* users_path, char* username, char* password, bool generate_pwd, int pwd_length, char* filename);
 static int remove_user(char* users_path, char* username);
 static int list_users(char* users_path);
 
@@ -164,6 +166,7 @@ usage(void)
    printf("  pgagroal-admin [ -f FILE ] [ COMMAND ] \n");
    printf("\n");
    printf("Options:\n");
+   printf("  -m, --master-key-file   MASTER_KEY_FILE Set the path to the master.key file\n");
    printf("  -f, --file FILE         Set the path to a user file\n");
    printf("                          Defaults to %s\n", PGAGROAL_DEFAULT_USERS_FILE);
    printf("  -U, --user USER         Set the user name\n");
@@ -189,6 +192,8 @@ int
 main(int argc, char** argv)
 {
    int c;
+   int error_code;
+   char* masterkey = NULL;
    char* username = NULL;
    char* password = NULL;
    char* file_path = NULL;
@@ -197,11 +202,13 @@ main(int argc, char** argv)
    int option_index = 0;
    size_t command_count = sizeof(command_table) / sizeof(struct pgagroal_command);
    struct pgagroal_parsed_command parsed = {.cmd = NULL, .args = {0}};
+   char* master_key_path = NULL;
 
    while (1)
    {
       static struct option long_options[] =
       {
+         {"master-key-file", required_argument, 0, 'm'},
          {"user", required_argument, 0, 'U'},
          {"password", required_argument, 0, 'P'},
          {"file", required_argument, 0, 'f'},
@@ -211,7 +218,7 @@ main(int argc, char** argv)
          {"help", no_argument, 0, '?'}
       };
 
-      c = getopt_long(argc, argv, "gV?f:U:P:l:",
+      c = getopt_long(argc, argv, "gV?f:m:U:P:l:",
                       long_options, &option_index);
 
       if (c == -1)
@@ -221,6 +228,9 @@ main(int argc, char** argv)
 
       switch (c)
       {
+         case 'm':
+            master_key_path = optarg;
+            break;
          case 'U':
             username = optarg;
             break;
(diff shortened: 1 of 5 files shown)


jesperpedersen:
Please, use the same pull request, and force push changes instead of opening new ones...

palak-chaturvedi:
Okk I will continue in this
There were some merge conflicts in it. I was not able to pull all the new commits from the master branch. So I thought I would make it much cleaner in this one.

src/admin.c
```diff
@@ -253,6 +267,85 @@ main(int argc, char** argv)
...
+      config = (struct main_configuration*)shmem;
+   }
+   else
+   {
+      ret = pgagroal_read_configuration(shmem, PGAGROAL_DEFAULT_CONF_FILE, false);
+      if (ret != PGAGROAL_CONFIGURATION_STATUS_OK)
+      {
+         if (!remote_connection)
+         {
+            errx(1, "Host (-h) and port (-p) must be specified to connect to the remote host");
```
fluca1978, line 324: Is `pgagroal-admin` supposed to be able to connect to the remote host? It doesn't even contain the arguments the message is claiming to use. I suspect there is extra copy and paste here.

src/admin.c
```diff
@@ -197,11 +201,18 @@ main(int argc, char** argv)
...
+   int ret;
+   size_t size;
+   bool remote_connection = false;
+   char* logfile = NULL;
 
    while (1)
    {
       static struct option long_options[] =
       {
+         {"config", required_argument, 0, 'c'},
```
fluca1978, line 215: isn't `m` less confusing than `c`?

src/admin.c
```diff
@@ -253,6 +267,85 @@ main(int argc, char** argv)
...
+         if (logfile)
+         {
+            config = (struct main_configuration*)shmem;
+
+            config->common.log_type = PGAGROAL_LOGGING_TYPE_FILE;
+            memset(&config->common.log_path[0], 0, MISC_LENGTH);
+            memcpy(&config->common.log_path[0], logfile, MIN(MISC_LENGTH - 1, strlen(logfile)));
+         }
+
+         if (pgagroal_start_logging())
```
fluca1978, line 340: is `pgagroal-admin` using the logging subsystem?

src/admin.c
```diff
@@ -318,77 +411,90 @@ main(int argc, char** argv)
...
+         goto error;
+      }
+
+      memset(&buf, 0, sizeof(buf));
+      snprintf(&buf[0], sizeof(buf), "%s/.pgagroal", pgagroal_get_home_directory());
+
+      if (stat(&buf[0], &st) == -1)
       {
-         warnx("No home directory for user \'%s\'", username);
+         mkdir(&buf[0], S_IRWXU);
```
fluca1978, line 444: should we catch a failure here?

fluca1978 (changes requested):
I need some time to test this, but at glance it seems there is extra work to setup the configuration in the `pgagroal-admin`.

src/admin.c
jesperpedersen: pgagroal-admin is local only, pgagroal-cli can do remote

src/admin.c
jesperpedersen: `c` is for main configuration file, otherwise it is another letter

src/admin.c
palak-chaturvedi: I haven't made any modifications to this part. This is already written in the master branch. I just indented it for those cases where filename is not provided.
I can look into it if there is some error?

src/admin.c
palak-chaturvedi: I have used main configuration file to add the master_key_file_location.

jesperpedersen: I think for `admin.c` we just need the explicit `-m` option e.g. no `-c`

src/admin.c
fluca1978: > I haven't made any modifications to this part. This is already written in the master branch. I just indented it for those cases where filename is not provided. I can look into it if there is some error?

Ok, if you will, please check for possible errors, or will postpone.

fluca1978:
@palak-chaturvedi please be patient, let's go back one step: add the `-m` flag to both `pagroal-` and `pgagroal-admin`, so that if specified, the executables can find out the correct location of the master key file and other related stuff. Then we will decide if and how to handle it within the main configuration file.

palak-chaturvedi:
If we add the -m flag then we will have to change lots of functions including reloading the configuration files and all the functions in configuration.c .
Is this better option than conf file changing?

jesperpedersen:
It is ok for the `pgagroal.conf` file, but the `pgagroal-admin` shouldn't depend on that file

palak-chaturvedi:
Ok so
If the conf file is not given then all the functions other than those that require master-key should not throw error.

I am working on the -m flag it would change all the configuration.c functions because master key is required in those codes.

fluca1978:
I'm wondering if this is rally worth, since it seems to be fairly complex (as of design).

However, mu opinion is that `pgagroal.conf` should have a setting for the master key file location, and `pgagroal-admin` should have a `-m` flag that reflects the same setting. In this way, the configuration of `main` should be straightforward, while the `pgagroal-admin` can use the same location specified on the command line.
However, this could cause some confusion, since `pgagroal-admin` does not ask for a master key location, and will write on the default path, so the user can end up with `pgagroal-admin` writing in a location and `pgagroal` looking into another without any error or warning. Is it worth?

jesperpedersen:
The `master_key_path` is in `pgagroal.conf`, so it will be used for `pgagroal`. `pgagroal-cli` also uses `pgagroal.conf`.

The only binary that you have to change are `pgagroal-admin` with the `-m` switch, and `pgagroal-vault` for the `pgagroal-vault.conf` configuration

palak-chaturvedi:
heyy @fluca1978 I made the changes can you please check once again?

fluca1978:
Looks fine. A few comments:

1) if I try to use a command that do not require a master key file with a wrong master key file, everything works fine. Maybe we should check for the `-m` usage early in `pgagroal-admin`:
```
% pgagroal-admin user ls --master-key-file /this/does/not/exists
luca
```

2) if `pgagroal` does not find the master key file, it warns with the wrong error message:
```
% pgagroal
pgagroal: Invalid master key file (file </etc/pgagroal/pgagroal_users.conf>)
```

3) I would remove `master_key_file_location` from the configuration and add the `-m` flag to the `pgagroal` tool itself.

fluca1978:
@palak-chaturvedi in commit 411a0fa you fixed the first problem, good!
Still issue two is there.
```
% pgagroal                                   
pgagroal: Invalid master key file (file </etc/pgagroal/pgagroal_users.conf>)

```
and this is related to issue three: I would remove the master key file from the configuration and add the `-m` option to `main.c`. Otherwise, if you believe it is better to keep such file in the configuration file, can you please argument your idea?

fluca1978:
Closing due to unfinished code. @palak-chaturvedi if you want to finish your job, please rebase and reopen.

PR #437, opened 2024-04-04
[#426] Fix a problem in `pgagroal_bind`

### LINKED ISSUE #426: Multiple instances can be run if the pid file is specified as a relative path
On 8a1d6416f02033a7a307439983ee6982513f5bc5 having `pid_file = relative_file.pid` makes `pgagroal` able to run multiple times from different directories.

Example: running the first instance:
```
```

Running the second instance from a different directory:

```
[luca@rachel]~% pgagroal
DEBUG network.c:648 server: bind: 127.0.0.1:54322 (Address already in use)
DEBUG network.c:648 server: bind: 10.0.2.15:54322 (Address already in use)
DEBUG network.c:648 server: bind: 192.168.222.50:54322 (Address already in use)
DEBUG network.c:648 server: bind: ::1:54322 (Address already in use)
DEBUG network.c:648 server: bind: fe80::a00:27ff:fee4:438d:54322 (Invalid argument)
```

I think we should either force an absolute `pid_file`, therefore aborting execution if `pid_file` is not absolute, or abort the execution if the `bind` fails. In any case, the fact that `bind` failure allows for continuation is suspicions.

fluca1978:
If the configuration `host` is set to `*`, meaning all the interfaces, the `pgagroal_bind` function was always returning a true value even if no available bind addresses were left.
In order to fix this, the commit tests if the `star_length` variable has been set to something non-zero, otherwise no more sockets are available (bound) and the function returns a `1` to indicate an error.

Close #426

### DIFF
--- a/src/libpgagroal/network.c
+++ b/src/libpgagroal/network.c
@@ -120,10 +120,17 @@ pgagroal_bind(const char* hostname, int port, int** fds, int* length)
          }
       }
 
+      freeifaddrs(ifaddr);
+
+      if (star_length == 0)
+      {
+         // cannot bind to anything!
+         return 1;
+      }
+
       *fds = star_fds;
       *length = star_length;
 
-      freeifaddrs(ifaddr);
       return 0;
    }
 


jesperpedersen:
Rebased and merged.

Issue #417, opened 2024-03-19
Make `pgagroal-admin` provide JSON output


fluca1978:
Since `pgagroal-cli` now has a complete JSON output support, it is time to provide also `pgagroal-admin` with JSON support.
The implementation should be simple, since no management protocol is involved, and the idea is that every command can directly output a JSON with all the details about the action.
The most notably use case is for `user ls` that has to provide a `list` with all the users.

T1t4m1un:
Could I take on this ticket?~

fluca1978:
@T1t4m1un yes, sure, thanks!

PR #440, opened 2024-04-17
[#417] Make pgagroal-admin provide JSON output

T1t4m1un:
Since the output data is very simple, I directly use printf to output in JSON format in the original function.

I think my implementation is the simplest and most reliable implementation considering the complexity of the data. Of course, if you have a better idea, please tell me.

If there are other use cases that need support, please let me know ~ I will complete the documentation after completing all use cases~

### DIFF
--- a/doc/ADMIN.md
+++ b/doc/ADMIN.md
@@ -22,6 +22,7 @@ Available options are the following ones:
   -P, --password PASSWORD Set the password for the user
   -g, --generate          Generate a password
   -l, --length            Password length
+  -F, --format  text|json Set the output format
   -V, --version           Display version information
   -?, --help              Display help
 
@@ -58,6 +59,16 @@ pgagroal-admin user del -U simon
 
 The `master-key` command allows the definition of a password to protect the vault of the users,
 that is the "container" for users' credentials.
+
+
+## JSON Output Format
+
+It is possible to obtain the output of a command in a JSON format by specyfing the `-F` (`--format`) option on the command line.
+Supported output formats are:
+- `text` (the default)
+- `json`
+
+For more details, see the corresponding section in the CLI documentation.
 
 
 ## Deprecated commands
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -223,6 +223,7 @@ Options:
   -P, --password PASSWORD Set the password for the user
   -g, --generate          Generate a password
   -l, --length            Password length
+  -F, --format  text|json Set the output format
   -V, --version           Display version information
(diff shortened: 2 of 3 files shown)


fluca1978:
@T1t4m1un thanks for the contribution, but I don't like the idea of using `printf` to print directly to the output.
I agree that the output format is quite simple, but the aim should be to have much more information in the JSON output, so the `list` array with the `count` and `users` subelements, the command name, version and all the other stuff we have for `pgagroal-cli` json output. In other words, please use the `json.c` stuff.

The aim is not to present output in brackets, it is to allow for automation, hence the output must be as complete as possible.

T1t4m1un:
> @T1t4m1un thanks for the contribution, but I don't like the idea of using `printf` to print directly to the output. I agree that the output format is quite simple, but the aim should be to have much more information in the JSON output, so the `list` array with the `count` and `users` subelements, the command name, version and all the other stuff we have for `pgagroal-cli` json output. In other words, please use the `json.c` stuff.

OK, I will refactor it~

> The aim is not to present output in brackets, it is to allow for automation, hence the output must be as complete as possible.

Could you give me some examples of the format? I don't clearly understand the format you envisioned😶‍🌫️

fluca1978:
> Could you give me some examples of the format? I don't clearly understand the format you envisioned😶‍🌫️

See <https://github.com/agroal/pgagroal/blob/master/doc/CLI.md#json-output-format>.
I think the command should give this kind of output:
```
$ pgagroal-admin user ls --format json

{
        "command":      {
                "name": "user ls",
                "status":       "OK",
                "error":        0,
                "exit-status":  0,
                "output":       {
                       "users":    {
                                        "count":        2,
                                        "list": [ "luca", "jesper" ]
                        }
                }
        },
        "application":  {
                "name": "pgagroal-admin",
                "major":        1,
                "minor":        7,
                "patch":        0,
                "version":      "1.7.0"
        }
}
```

or something simpler, but the above should be for free using the `json.c` stuff.

T1t4m1un:
image

@fluca1978 Hey Luca, I think this version is what you want~ However I'm not sure if these json manipulations should be encapsulated into a function, so I completed the functionality first. If you think encapsulation is necessary, please give me some suggestions on the name and where to place it~

I'll add this option into documentation if there's no other problems with this version~

fluca1978 (approved):
I think it is done, I would only change the assignment of subelements so to get them early in the `json` structure, as I commented.

@jesperpedersen  PTAL

fluca1978:
We don't use cJSON anymore, so this has to be rebased and adapted. I'm closing this request.

PR #454, opened 2024-07-24
Enable Vault Prometheus

ashu3103:
### About this commit
- Bind an HTTP server in `pgagroal-vault`  at port `metrics` in configuration to track server activities
- Added some fields like - `metrics`, `hugepage`, `metric_cache_max_age` and `metric_cache_max_size` in the common configuration
- Added an additional check `is_prometheus_enabled` on all the `prometheus` functions and code blocks which basically checks whether the metrics port is active or not.
- Based on the `metric` port value we allocate memory to `prometheus` structure and `prometheus_cache`

### Future work
- Segregate the `prometheus` structure for `main` and `vault` similar to `configuration` for optimal memory management.
- `prometheus` should be initiated only when `metric` port is active --> **Document it properly somewhere**.

@jesperpedersen @fluca1978  PTAL.

### DIFF
--- a/src/include/pgagroal.h
+++ b/src/include/pgagroal.h
@@ -371,10 +371,28 @@ struct prometheus_cache
 } __attribute__ ((aligned (64)));
 
 /** @struct
- * Defines the Prometheus metrics
+ * Defines the common Prometheus metrics
  */
 struct prometheus
 {
+   // logging
+   atomic_ulong logging_info;          /**< Logging: INFO */
+   atomic_ulong logging_warn;          /**< Logging: WARN */
+   atomic_ulong logging_error;         /**< Logging: ERROR */
+   atomic_ulong logging_fatal;         /**< Logging: FATAL */
+
+   // internal connections
+   atomic_int client_sockets;          /**< The number of sockets the client used */
+   atomic_int self_sockets;            /**< The number of sockets used by pgagroal itself */
+
+} __attribute__ ((aligned (64)));
+
+/** @struct
+ * Defines the Main Prometheus metrics
+ */
+struct main_prometheus
+{
+   struct prometheus prometheus_base;            /**< Common base class */
    atomic_ulong session_time[HISTOGRAM_BUCKETS]; /**< The histogram buckets */
    atomic_ulong session_time_sum;                /**< Total session time */
 
@@ -389,11 +407,6 @@ struct prometheus
    atomic_ulong connection_max_connection_age; /**< The number of max connection age calls */
    atomic_ulong connection_flush;              /**< The number of flush calls */
    atomic_ulong connection_success;            /**< The number of success calls */
-
-   atomic_ulong logging_info;  /**< Logging: INFO */
-   atomic_ulong logging_warn;  /**< Logging: WARN */
-   atomic_ulong logging_error; /**< Logging: ERROR */
-   atomic_ulong logging_fatal; /**< Logging: FATAL */
 
    /**< The number of connection awaiting due to `blocking_timeout` */
    atomic_ulong connections_awaiting[NUMBER_OF_LIMITS];
@@ -413,13 +426,18 @@ struct prometheus
    atomic_ullong network_sent;          /**< The bytes sent by clients */
    atomic_ullong network_received;      /**< The bytes received from servers */
 
-   atomic_int client_sockets;          /**< The number of sockets the client used */
-   atomic_int self_sockets;            /**< The number of sockets used by pgagroal itself */
-
    atomic_ulong server_error[NUMBER_OF_SERVERS]; /**< The number of errors for a server */
    atomic_ulong failed_servers;                  /**< The number of failed servers */
    struct prometheus_connection prometheus_connections[];  /**< The number of prometheus connections (FMA) */
 
+} __attribute__ ((aligned (64)));
+
+/** @struct
+ * Defines the Vault Prometheus metrics
+ */
+struct vault_prometheus
+{
+   struct prometheus prometheus_base;
 } __attribute__ ((aligned (64)));
 
 /** @struct
@@ -430,6 +448,7 @@ struct configuration
    char configuration_path[MAX_PATH]; /**< The configuration path */
    char host[MISC_LENGTH];            /**< The host */
    int port;                          /**< The port */
+   int authentication_timeout;        /**< The authentication timeout in seconds */
 
    // Logging
    int log_type;                       /**< The logging type */
@@ -443,6 +462,12 @@ struct configuration
    char log_line_prefix[MISC_LENGTH];  /**< The logging prefix */
    atomic_schar log_lock;              /**< The logging lock */
    char default_log_path[MISC_LENGTH]; /**< The default logging path */
+
+   // Prometheus
+   unsigned char hugepage;              /**< Huge page support */
+   int metrics;                         /**< The metrics port */
+   unsigned int metrics_cache_max_age;  /**< Number of seconds to cache the Prometheus response */
+   unsigned int metrics_cache_max_size; /**< Number of bytes max to cache the Prometheus response */
 };
 
 /** @struct
@@ -469,9 +494,6 @@ struct main_configuration
    char admins_path[MAX_PATH];        /**< The admins path */
    char superuser_path[MAX_PATH];     /**< The superuser path */
 
-   int metrics;            /**< The metrics port */
-   unsigned int metrics_cache_max_age;      /**< Number of seconds to cache the Prometheus response */
-   unsigned int metrics_cache_max_size;     /**< Number of bytes max to cache the Prometheus response */
    int management;         /**< The management port */
    bool gracefully;        /**< Is pgagroal in gracefully mode */
 
@@ -503,7 +525,6 @@ struct main_configuration
    int validation;               /**< Validation mode */
    int background_interval;      /**< Background validation timer in seconds */
    int max_retries;              /**< The maximum number of retries */
-   int authentication_timeout;   /**< The authentication timeout in seconds */
    int disconnect_client;        /**< Disconnect client if idle for more than the specified seconds */
    bool disconnect_client_force; /**< Force a disconnect client if active for more than the specified seconds */
    char pidfile[MAX_PATH];       /**< File containing the PID */
@@ -514,7 +535,6 @@ struct main_configuration
    bool nodelay;            /**< Use NODELAY */
    bool non_blocking;       /**< Use non blocking */
    int backlog;             /**< The backlog for listen */
-   unsigned char hugepage;  /**< Huge page support */
    bool tracker;            /**< Tracker support */
    bool track_prepared_statements; /**< Track prepared statements (transaction pooling) */
 
--- a/src/libpgagroal/prometheus.c
+++ b/src/libpgagroal/prometheus.c
@@ -71,10 +71,13 @@
 static int resolve_page(struct message* msg);
 static int unknown_page(int client_fd);
 static int home_page(int client_fd);
+static int home_vault_page(int client_fd);
 static int metrics_page(int client_fd);
+static int metrics_vault_page(int client_fd);
 static int bad_request(int client_fd);
 
 static void general_information(int client_fd);
+static void general_vault_information(int client_fd);
 static void connection_information(int client_fd);
 static void limit_information(int client_fd);
 static void session_information(int client_fd);
@@ -82,6 +85,7 @@ static void pool_information(int client_fd);
 static void auth_information(int client_fd);
 static void client_information(int client_fd);
 static void internal_information(int client_fd);
+static void internal_vault_information(int client_fd);
 static void connection_awaiting_information(int client_fd);
 
 static int send_chunk(int client_fd, char* data);
@@ -92,6 +96,7 @@ static bool metrics_cache_append(char* data);
 static bool metrics_cache_finalize(void);
 static size_t metrics_cache_size_to_alloc(void);
 static void metrics_cache_invalidate(void);
+static bool is_prometheus_enabled(void);
 
 void
 pgagroal_prometheus(int client_fd)
@@ -101,12 +106,17 @@ pgagroal_prometheus(int client_fd)
    struct message* msg = NULL;
    struct main_configuration* config;
 
+   if (!is_prometheus_enabled())
+   {
+      exit(1);
+   }
+
    pgagroal_start_logging();
    pgagroal_memory_init();
 
    config = (struct main_configuration*)shmem;
 
-   status = pgagroal_read_timeout_message(NULL, client_fd, config->authentication_timeout, &msg);
+   status = pgagroal_read_timeout_message(NULL, client_fd, config->common.authentication_timeout, &msg);
    if (status != MESSAGE_STATUS_OK)
    {
       goto error;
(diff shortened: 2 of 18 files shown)


src/libpgagroal/prometheus.c
```diff
@@ -92,10 +96,16 @@ static bool metrics_cache_append(char* data);
 static bool metrics_cache_finalize(void);
 static size_t metrics_cache_size_to_alloc(void);
 static void metrics_cache_invalidate(void);
+static bool is_prometheus_enabled(void);
 
 void
 pgagroal_prometheus(int client_fd)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 104: You should do this after the variable definitions

src/libpgagroal/prometheus.c
```diff
@@ -149,21 +159,79 @@ pgagroal_prometheus(int client_fd)
    exit(1);
 }
 
+void
+pgagroal_vault_prometheus(int client_fd)
+{
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 165: Same

src/libpgagroal/prometheus.c
```diff
@@ -244,6 +312,11 @@ pgagroal_init_prometheus(size_t* p_size, void** p_shmem)
 void
 pgagroal_prometheus_session_time(double time)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 315: Same

src/libpgagroal/prometheus.c
```diff
@@ -330,6 +403,11 @@ pgagroal_prometheus_session_time(double time)
 void
 pgagroal_prometheus_connection_error(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 406: Same

src/libpgagroal/prometheus.c
```diff
@@ -340,6 +418,11 @@ pgagroal_prometheus_connection_error(void)
 void
 pgagroal_prometheus_connection_kill(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 421: Same

src/libpgagroal/prometheus.c
```diff
@@ -350,6 +433,11 @@ pgagroal_prometheus_connection_kill(void)
 void
 pgagroal_prometheus_connection_remove(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 436: Same

src/libpgagroal/prometheus.c
```diff
@@ -360,6 +448,11 @@ pgagroal_prometheus_connection_remove(void)
 void
 pgagroal_prometheus_connection_timeout(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 451: Same

src/libpgagroal/prometheus.c
```diff
@@ -370,6 +463,11 @@ pgagroal_prometheus_connection_timeout(void)
 void
 pgagroal_prometheus_connection_return(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 466: Same

src/libpgagroal/prometheus.c
```diff
@@ -380,6 +478,11 @@ pgagroal_prometheus_connection_return(void)
 void
 pgagroal_prometheus_connection_invalid(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 481: Same

src/libpgagroal/prometheus.c
```diff
@@ -390,6 +493,11 @@ pgagroal_prometheus_connection_invalid(void)
 void
 pgagroal_prometheus_connection_get(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 496: Same

src/libpgagroal/prometheus.c
```diff
@@ -400,6 +508,11 @@ pgagroal_prometheus_connection_get(void)
 void
 pgagroal_prometheus_connection_idletimeout(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 511: Same

src/libpgagroal/prometheus.c
```diff
@@ -410,6 +523,11 @@ pgagroal_prometheus_connection_idletimeout(void)
 void
 pgagroal_prometheus_connection_max_connection_age(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 526: Same

src/libpgagroal/prometheus.c
```diff
@@ -420,6 +538,11 @@ pgagroal_prometheus_connection_max_connection_age(void)
 void
 pgagroal_prometheus_connection_awaiting(int limit_index)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 541: Same

src/libpgagroal/prometheus.c
```diff
@@ -435,6 +558,11 @@ pgagroal_prometheus_connection_awaiting(int limit_index)
 void
 pgagroal_prometheus_connection_unawaiting(int limit_index)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 561: Same

src/libpgagroal/prometheus.c
```diff
@@ -453,6 +581,11 @@ pgagroal_prometheus_connection_unawaiting(int limit_index)
 void
 pgagroal_prometheus_connection_flush(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 584: Same

src/libpgagroal/prometheus.c
```diff
@@ -463,6 +596,11 @@ pgagroal_prometheus_connection_flush(void)
 void
 pgagroal_prometheus_connection_success(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 599: Same

src/libpgagroal/prometheus.c
```diff
@@ -473,6 +611,11 @@ pgagroal_prometheus_connection_success(void)
 void
 pgagroal_prometheus_auth_user_success(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 614: Same

src/libpgagroal/prometheus.c
```diff
@@ -483,6 +626,11 @@ pgagroal_prometheus_auth_user_success(void)
 void
 pgagroal_prometheus_auth_user_bad_password(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 629: Same

src/libpgagroal/prometheus.c
```diff
@@ -493,6 +641,11 @@ pgagroal_prometheus_auth_user_bad_password(void)
 void
 pgagroal_prometheus_auth_user_error(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 644: Same

src/libpgagroal/prometheus.c
```diff
@@ -503,6 +656,11 @@ pgagroal_prometheus_auth_user_error(void)
 void
 pgagroal_prometheus_client_wait_add(void)
 {
+   if (!is_prometheus_enabled())
```
jesperpedersen, line 659: Same

src/libpgagroal/prometheus.c
```diff
@@ -2090,6 +2583,31 @@ internal_information(int client_fd)
...
+
+   prometheus = (struct prometheus*)prometheus_shmem;
+
+   data = pgagroal_append(data, "#HELP pgagroal_client_sockets Number of sockets the client used\n");
+   data = pgagroal_append(data, "#TYPE pgagroal_client_sockets gauge\n");
+   data = pgagroal_append(data, "pgagroal_client_sockets ");
+   data = pgagroal_append_int(data, atomic_load(&prometheus->client_sockets));
+   data = pgagroal_append(data, "\n\n");
+
+   data = pgagroal_append(data, "#HELP pgagroal_self_sockets Number of sockets used by pgagroal itself\n");
```
jesperpedersen, line 2600: pgagroal-vault

src/libpgagroal/prometheus.c
```diff
@@ -1241,6 +1484,108 @@ home_page(int client_fd)
...
+   data = pgagroal_append(data, "  <p>\n");
+   data = pgagroal_append(data, "  The number of FATAL logging statements\n");
+   data = pgagroal_append(data, "  </p>\n");
+   data = pgagroal_append(data, "  <h2>pgagroal_client_sockets</h2>\n");
+   data = pgagroal_append(data, "  <p>\n");
+   data = pgagroal_append(data, "   Number of sockets the client used\n");
+   data = pgagroal_append(data, "  </p>\n");
+   data = pgagroal_append(data, "  <h2>pgagroal_self_sockets</h2>\n");
+   data = pgagroal_append(data, "  <p>\n");
+   data = pgagroal_append(data, "   Number of sockets used by pgagroal itself\n");
```
jesperpedersen, line 1559: pgagroal-vault

src/libpgagroal/prometheus.c
jesperpedersen, line 659: Well, all of them...

src/libpgagroal/prometheus.c
```diff
@@ -1241,6 +1484,108 @@ home_page(int client_fd)
...
+   data = pgagroal_append(data, "<head>\n");
+   data = pgagroal_append(data, "  <title>pgagroal-vault exporter</title>\n");
+   data = pgagroal_append(data, "  <meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\"/>");
+   data = pgagroal_append(data, "</head>\n");
+   data = pgagroal_append(data, "<body>\n");
+   data = pgagroal_append(data, "  <h1>pgagroal-vault exporter</h1>\n");
+   data = pgagroal_append(data, "  <p>\n");
+   data = pgagroal_append(data, "   <a href=\"/metrics\">Metrics</a>\n");
+   data = pgagroal_append(data, "  </p>\n");
+   data = pgagroal_append(data, "  <h2>pgagroal_logging_info</h2>\n");
```
jesperpedersen, line 1537: These should be called `pgagroal_vault_XYZ`

src/libpgagroal/prometheus.c
```diff
@@ -1546,6 +2001,44 @@ general_information(int client_fd)
...
 
+static void
+general_vault_information(int client_fd)
+{
+   char* data = NULL;
+   struct prometheus* prometheus;
+
+   prometheus = (struct prometheus*)prometheus_shmem;
+
+   data = pgagroal_append(data, "#HELP pgagroal_logging_info The number of INFO logging statements\n");
```
jesperpedersen, line 2012: `pgagroal_vault_logging_info`

src/libpgagroal/prometheus.c
```diff
@@ -2090,6 +2583,31 @@ internal_information(int client_fd)
...
+internal_vault_information(int client_fd)
+{
+   char* data = NULL;
+   struct prometheus* prometheus;
+
+   prometheus = (struct prometheus*)prometheus_shmem;
+
+   data = pgagroal_append(data, "#HELP pgagroal_client_sockets Number of sockets the client used\n");
+   data = pgagroal_append(data, "#TYPE pgagroal_client_sockets gauge\n");
+   data = pgagroal_append(data, "pgagroal_client_sockets ");
```
jesperpedersen, line 2596: `pgagroal_vault_XYZ`

jesperpedersen (changes requested):

jesperpedersen:
Remember the tutorial, man page and manual as well

ashu3103:
> Remember the tutorial, man page and manual as well

Yeah working on it parallely

jesperpedersen:
Maybe enable metrics by default in the `pgagroal_vault.conf` as well

jesperpedersen:
I think that your "Future" section should be easy as well

jesperpedersen:
> prometheus should be initiated only when metric port is active --> Document it properly somewhere.

It is documented in the configuration files - users doesn't need to know about the internals

ashu3103:
> > prometheus should be initiated only when metric port is active --> Document it properly somewhere.
> 
> It is documented in the configuration files - users doesn't need to know about the internals

Yeah, but for new developers it might be required

jesperpedersen:
You can put it in

* https://github.com/agroal/pgagroal/blob/master/doc/ARCHITECTURE.md
* https://github.com/agroal/pgagroal/blob/master/doc/manual/dev-02-architecture.md

ashu3103:
@jesperpedersen The suggested changes have been integrated PTAL.

src/libpgagroal/prometheus.c
```diff
@@ -2267,6 +2824,7 @@ pgagroal_init_prometheus_cache(size_t* p_size, void** p_shmem)
 
 error:
    // disable caching
+
```
fluca1978, line 2827: please remove the empty line and leave the comment near the code

src/libpgagroal/prometheus.c
```diff
@@ -2402,6 +2960,13 @@ metrics_cache_finalize(void)
...
    now = time(NULL);
-   cache->valid_until = now + config->metrics_cache_max_age;
+   cache->valid_until = now + config->common.metrics_cache_max_age;
    return cache->valid_until > now;
 }
+
+static bool
+is_prometheus_enabled(void)
+{
+   struct configuration* config = (struct configuration*)shmem;
```
fluca1978, line 2970: add test against nulls?

src/main.c
```diff
@@ -742,20 +742,23 @@ main(int argc, char** argv)
       errx(1, "Failed to start logging");
    }
 
-   if (pgagroal_init_prometheus(&prometheus_shmem_size, &prometheus_shmem))
+   if (config->common.metrics > 0)
```
fluca1978, line 745: should we export `is_prometheus_active` function and use it so the code is consistent?

src/main.c
```diff
@@ -1044,14 +1047,14 @@ main(int argc, char** argv)
       ev_periodic_start (main_loop, &rotate_frontend_password);
    }
 
-   if (config->metrics > 0)
+   if (config->common.metrics > 0)
```
fluca1978, line 1050: same as above

src/vault.c
```diff
@@ -379,6 +435,25 @@ main(int argc, char** argv)
       errx(1, "Failed to start logging");
    }
 
+   if (config->common.metrics > 0)
```
fluca1978, line 438: same as above

src/vault.c
```diff
@@ -445,21 +529,64 @@ main(int argc, char** argv)
 
    start_vault_io();
 
+   if (config->common.metrics > 0)
```
fluca1978, line 532: same as above

src/include/pgagroal.h
```diff
@@ -371,10 +371,28 @@ struct prometheus_cache
...
+   // internal connections
+   atomic_int client_sockets;          /**< The number of sockets the client used */
+   atomic_int self_sockets;            /**< The number of sockets used by pgagroal itself */
+
+} __attribute__ ((aligned (64)));
+
+/** @struct
+ * Defines the Main Prometheus metrics
+ */
+struct main_prometheus
```
fluca1978, line 393: I don't get the point in splitting the `prometheus` structure into two parts, it seems to me it is useless: if `metrics > 0` prometheus is active, so there is no advantage in memory allocation. Also, while `vault_configuration` added fields around `configuration`, here you are narrowing the structure only to the fields you need.
While I don't dislike in total this approach, it seems counterintuitive to have `prometheus` not being the main metrics structure as I would at glange expect. I would rather make a `prometheus_logging` structure, a `prometheus_socket` one and have `prometheus` to wrap the above, so that `prometheus_vault` can handle only the former twos.

fluca1978 (changes requested):

src/libpgagroal/prometheus.c
ashu3103: Not required as we initiate `prometheus` memory allocation only if `config->metrics` > 0.

src/include/pgagroal.h
ashu3103: > if metrics > 0 prometheus is active, so there is no advantage in memory allocation.

True, the aim of splitting `prometheus` is to separate the memory space allocation for different programs (`main`/`vault`) as there as several fields which are not utilized by vault but are still allocated extra memory for unused fields.

> Also, while vault_configuration added fields around configuration, here you are narrowing the structure only to the  
> fields you need.

As of now I have added just the required fields and more fields can be added as and when required.

> While I don't dislike in total this approach, it seems counterintuitive to have prometheus not being the main metrics  
> structure as I would at glange expect. I would rather make a prometheus_logging structure, a prometheus_socket 
> one and have prometheus to wrap the above, so that prometheus_vault can handle only the former twos.

Can be done that way also. The thing i kept in my mind was that if in future if the project integrates other such service programs like `vault`, that will need `prometheus` then it can directly inherit the common fields (common for all services/servers) while keep on adding its specific fields in its `prometheus` structure.

src/main.c
ashu3103: Yeah can be done, but won't it waste some memory and computation as `config` is already there in the program it will again define `struct configuration` in the check function.

However, as far as consistency is concerned it can be done.

fluca1978: > Yeah can be done, but won't it waste some memory and computation as `config` is already there in the program it will again define `struct configuration` in the check function.

The only waste I can see is the function call, but I don't think we are optimizing at this level right now.
It could also be a macro.

src/libpgagroal/prometheus.c
fluca1978: > Not required as we initiate `prometheus` memory allocation only if `config->metrics` > 0.

Well, but if we export the function to somewhere else, we could need some extra (redundant) checks.

src/include/pgagroal.h
fluca1978: > Can be done that way also. The thing i kept in my mind was that if in future if the project integrates other such service programs like `vault`, that will need `prometheus` then it can directly inherit the common fields (common for all services/servers) while keep on adding its specific fields in its `prometheus` structure.

This is clear, and is a good approach, but I don't think that `common` is the right name for the structure. It seems, so far, you need only a subset of the structure for the vault, so I'm not sure if this effort in splitting the structure is worth. I suspect you will find the need to add more fields here and there from time to time, so at this stage I would not split the structure. Surely, I will not name it as `common`, since that makes me think at *common* fields, while so far there are only "basic" fields like logging counters.
@jesperpedersen what do you think?

src/libpgagroal/prometheus.c
ashu3103: Okay, will add the `NULL` check for `prometheus` to avoid any unlikely cases that may occur.

src/main.c
ashu3103: > The only waste I can see is the function call, but I don't think we are optimizing at this level right now.

Okay :+1: 

> It could also be a macro.

Not sure how a macro could solve the issue, can you please elaborate a little.

Also, if we add `NULL` check of `prometheus` in `is_prometheus_enabled()` function (as per the aforementioned change request) then we can't use it in this scenario as it will always fail the test (since `prometheus` is checked before it is being initialised), and the only test which should be checked is `metrics > 0`. So I believe the current setup is ok.

What do you think about it? @fluca1978.

fluca1978: > Also, if we add `NULL` check of `prometheus` in `is_prometheus_enabled()` function (as per the aforementioned change request) then we can't use it in this scenario as it will always fail the test (since `prometheus` is checked before it is being initialised), and the only test which should be checked is `metrics > 0`. So I believe the current setup is ok.

Indeed having the test against nulls will fail in this case, so better leave it off the `is_prometheus_enabled`. Sorry, I was running out of coffee.

src/include/pgagroal.h
jesperpedersen: `prometheus_base` is an alternative

jesperpedersen (changes requested):

jesperpedersen:
I think it looks good but `prometheus_common` or `prometheus_base` is probably better as a name

jesperpedersen:
Merged.

PR #455, opened 2024-08-04
Enable TLS in vault

ashu3103:
### About this commit

This commit is aimed to integrate the tls mode in `pgagroal-vault` similar to `pgagroal` for secure connections.

-  Add required fields in the `vault` configuration like `tls`, `tls_cert_file`, `tls_ca_file` and `tls_key_file`
- Before handling the `HTTP` request, the `vault` will check whether the client wants to initiate `TLS Handshake` with `ClientHello` by peeking the first few bytes of the data after the establishing the connection and checking it against the `ClientHello` bytes.
- **Case1:** If client initiates the TLS Handshake and `config->tls` is on, both server and client will do the TLS handshake and then server handles the upcoming `HTTP` request.
- **Case2:** If client doesn't initiates the TLS Handshake and `config->tls` is on, the server will bypass TLS handshake and instead server handles the incoming `HTTP` request.
- **Case3:** If client initiates the TLS Handshake and `config->tls` is off, the server will simply closes the connection and report appropriate errors.

@jesperpedersen PTAL


jesperpedersen:
We need to remove **Case 2** - if TLS is on then it is mandatory for the clients

ashu3103:
> We need to remove **Case 2** - if TLS is on then it is mandatory for the clients

Done.

If Case 2 is encountered, an HTTP message with 301 (Redirect Message) code will be sent.

@jesperpedersen PTAL.

jesperpedersen:
Merged.

PR #458, opened 2024-08-22
[#457] Inform the user of misconfigured `log_level` debug value

### LINKED ISSUE #457: Report misconfiguration of `log_level` when set to `debug`
So far, the `log_level` can be set to a `debug` value with an optional number from `1` to `5`.
Every numeric value outside the range 1-5 is adjusted accordingly, so `debug-5` is set to `debug1` and `debug100` is set to `debug5`.
However, the user should be informed that its configuration has been ignored and a different one has been applied.

fluca1978:
If the user set the `log_level = debugX` with X being a number outside of the valid range 1-5, the application adjusts the logging level to the upper or lower boundary, now informing the user that her configuration has been ignored.

Documentation updated.

Also adds `trace` as a synonim for `debug5`.

Close #457


jesperpedersen:
Merged.

Issue #459, opened 2024-08-27
Unknown value for `tls` field for server configuration in `pgagroal.conf`


ashu3103:
## Bug Description

For the configuration
```
[pgagroal]
host = localhost
port = 2345

log_type = console
log_level = info
log_path = 

max_connections = 100
idle_timeout = 600
validation = off
unix_socket_dir = /tmp/

[primary]
host = localhost
port = 5432
tls = on
tls_ca_file = </path/to/ca/root.crt>

```

Warning is emitted :- `"Unknown key <tls> with value <on> in section [primry] (line %d of file <17>)"`

## Possible reasons
In `_apply_main_configuration` function, `unknown` is well defined.
so for `primary` section

```
else if (key_in_section("tls", section, key, true, &unknown))
   {
      if (as_bool(value, &config->common.tls))
      {
         unknown = true;
      }
   }
   else if (key_in_section("tls", section, key, false, &unknown))
   {
      if (as_bool(value, &srv->tls))
      {
         unknown = true;
      }
   }
```
the first call to `key_in_section` will return false but will make `unknown = true` because in `key_in_section` function (for the first call) --
```
if (global && (!strncmp(section, PGAGROAL_MAIN_INI_SECTION, MISC_LENGTH) || !strncmp(section, PGAGROAL_VAULT_INI_SECTION, MISC_LENGTH)))
   {
      return true;
   }
   else if (!global && strlen(section) > 0)
   {
      return true;
   }
   else
   {
      if (unknown)
      {
         *unknown = true;
      }
      return false;
   }
```
global will be true so code will go to else block and since `unknown` will always be true (well-defined), `*unknown` will be set true.

## Solution
The last argument of `key_in_section` function should be NULL where `global` is true i.e. if the section is either `[pgagroal]` or `[pgagroal-vault]`

jesperpedersen:
Shouldn't the check be against vault ?

ashu3103:
> Shouldn't the check be against vault ?

Hmm, I am not sure how it is related to `vault`.

Cause I'm facing this issue with the main

ashu3103:
@jesperpedersen, Shall we close this issue?

I am not facing any problem with tls in server side

PR #460, opened 2024-08-27
[#459] Unknown tls value in server configuration fix

ashu3103:
## WIP: Issue #459 

Changed the value of `unknown` pointer in `key_in_section` to NULL where `global=true` (section is pgagroal/pgagroal-vault).

@jesperpedersen PTAL

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -4408,7 +4408,7 @@ pgagroal_apply_main_configuration(struct main_configuration* config,
          unknown = true;
       }
    }
-   else if (key_in_section("tls", section, key, true, &unknown))
+   else if (key_in_section("tls", section, key, true, NULL))
    {
       if (as_bool(value, &config->common.tls))
       {
@@ -4422,7 +4422,7 @@ pgagroal_apply_main_configuration(struct main_configuration* config,
          unknown = true;
       }
    }
-   else if (key_in_section("tls_ca_file", section, key, true, &unknown))
+   else if (key_in_section("tls_ca_file", section, key, true, NULL))
    {
       max = strlen(value);
       if (max > MISC_LENGTH - 1)
@@ -4440,7 +4440,7 @@ pgagroal_apply_main_configuration(struct main_configuration* config,
       }
       memcpy(srv->tls_ca_file, value, max);
    }
-   else if (key_in_section("tls_cert_file", section, key, true, &unknown))
+   else if (key_in_section("tls_cert_file", section, key, true, NULL))
    {
       max = strlen(value);
       if (max > MISC_LENGTH - 1)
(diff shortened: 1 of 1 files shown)


jesperpedersen:
`struct server`  has a `tls` value...

ashu3103:
> `struct server` has a `tls` value...

Yeah, but when trying to use `tls = on` in `[primary]` section, I am getting this warning.

jesperpedersen:
I'll leave this to @fluca1978 -- we have `.tls` value for `struct server` so we need to parse it

ashu3103:
> I'll leave this to @fluca1978 -- we have `.tls` value for `struct server` so we need to parse it

Okay, but why can't we parse it just like we parse `"host"` and `"port"` for server. Like -

```
else if (key_in_section("port", section, key, true, NULL))
   {
      if (as_int(value, &config->common.port))
      {
         unknown = true;
      }
   }
   else if (key_in_section("port", section, key, false, &unknown))
   {
      memcpy(&srv->name, section, strlen(section));
      if (as_int(value, &srv->port))
      {
         unknown = true;
      }
      atomic_store(&srv->state, SERVER_NOTINIT);
   }
```

Here, for `global=true`, we are passing `NULL` and for `global=false` we are passing &unknown

fluca1978:
Since the change is scoped within `pgagroal_apply_vault_configuration` (https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L4736), I think we can accept the proposed solution.

fluca1978 (commented):
I think it is fine, since it is scoped within the vault configuration.

ashu3103:
> Since the change is scoped within `pgagroal_apply_vault_configuration` (https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L4736), I think we can accept the proposed solution.

I have proposed the changes in `pgagroal_apply_main_configuration`, since I was facing the issue while trying to use `tls` for pgagroal--postgres server.

Kindly refer to this issue : https://github.com/agroal/pgagroal/issues/459

fluca1978:
> > Since the change is scoped within `pgagroal_apply_vault_configuration` (https://github.com/agroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L4736), I think we can accept the proposed solution.
> 
> I have proposed the changes in `pgagroal_apply_main_configuration`, since I was facing the issue while trying to use `tls` for pgagroal--postgres server.
> 
> Kindly refer to this issue : #459

Sorry, I was mislead by another comment.
I've checked out the pull request, and it works fine. Even setting `tls` in main section works fine, this can be easily tested with `pgagroal-cli conf get tls` or `pgagroal-cli conf get server.primary.tls`.

@jesperpedersen I think it is fine to merge.

jesperpedersen:
Merged.

Issue #461, opened 2024-09-10
Postgres is refusing tls connection from `pgagroal`


ashu3103:
## Background

Trying to connect to postgres using tls (disabling ssl from `client` - `pgagroal`), focusing on tls of `pgagroal` - `postgres`.

#### pgagroal.conf
```
[pgagroal]
host = localhost
port = 2345

log_type = console
log_level = debug5
log_path = 

max_connections = 100
idle_timeout = 600
validation = off
unix_socket_dir = /tmp/

[primary]
host = localhost
port = 5432
tls = on
tls_ca_file = </path/to/root.crt>
```
#### pgagroal_hba.conf
```
host    all      all   all      all
```

#### pg_hba.conf
```
hostssl all all all md5
```
All connection must do ssl!

#### postgresql.conf
```
...
ssl = on
ssl_cert_file = </path/to/server.cert>
ssl_key_file = </path/to/server.key>
...
```

## Issue 

While executing the command `PGSSLMODE=disable psql -h localhost -p 2345 -U <username> <databasename>`

- The `pgagroal` successfully do the SSL handshake.
- Then `pgagroal` sends the StartUpMessage to postgres
- Now in `server_passthrough` we are passing the authentication request messages between client and postgres.
- But while exchanging messages in `server_passthrough`, we are writing and reading from postgres file descriptor with ssl=NULL (**which is an issue since our ssl handshake was successfull and all subsequent messages after ssl handshake must be SSL-encrypted**) 
![Screenshot from 2024-09-11 01-05-33](https://github.com/user-attachments/assets/ca3b3a7a-e463-46e6-8262-3456bb2fe7b3)

So, on writing/reading anything from postgres file descriptor, the postgres server terminates/resets connection.
Logs of postgres (after pgagroal read/write to postgres server in `server_passthrough`)
```
2024-09-11 00:27:29.702 IST [187451] ashu3103@test LOG:  SSL error: wrong version number
2024-09-11 00:27:29.702 IST [187451] ashu3103@test LOG:  could not receive data from client: Connection reset by peer
```

The current `security.c` file (where the )
```
...
2218: status = pgagroal_write_message(NULL, server_fd, msg);
2219:      if (status != MESSAGE_STATUS_OK)
2220:      {
2221:         goto error;
2222:      }
2223:      pgagroal_free_message(msg);
2224:
2225:      status = pgagroal_read_block_message(NULL, server_fd, &msg);
...
```

## Proposed Solution
[...]

PR #462, opened 2024-09-10
[#461] fix ssl connection to postgres in server_passthrough

ashu3103:
### Work In Progress

Tried to resolve the issue #461 

@jesperpedersen @fluca1978 PTAL


jesperpedersen:
I would start with the TLS tutorial - describe a setup where PostgreSQL has a self-signed certificate - and, how to generate that - and pgagroal has a self-signed certificate - and, how to generate it. You could call that section "Transport Security Layer (TLS) only setup".

How does a client get a connection through pgagroal to PostgreSQL when only TLS is enabled the whole way ?

So, `c_ssl` and `s_ssl` are very different in each layer, so we need to make sure they are the correct ones

ashu3103:
> I would start with the TLS tutorial - describe a setup where PostgreSQL has a self-signed certificate - and, how to generate that - and pgagroal has a self-signed certificate - and, how to generate it. You could call that section "Transport Security Layer (TLS) only setup".
> 

Okay, I'll add that.

> How does a client get a connection through pgagroal to PostgreSQL when only TLS is enabled the whole way ?
> 

The connection that client is not cached its a new connection (unpooled_connection)

> So, `c_ssl` and `s_ssl` are very different in each layer, so we need to make sure they are the correct ones

Can you please elaborate what does (`each layer`) exactly represents.

jesperpedersen:
By layers I mean: Client <-> pgagroal <-> PostgreSQL

The root certificate is different - since it is self-signed - between pgagroal and PostgreSQL.

If there are issues you will find them when you work your way through the new section in the tutorial

ashu3103:
> By layers I mean: Client <-> pgagroal <-> PostgreSQL
> 
> The root certificate is different - since it is self-signed - between pgagroal and PostgreSQL.
> 
> If there are issues you will find them when you work your way through the new section in the tutorial

Done with the tutorial

Working alright with the current changes when a new connection is created using `unpooled_connection`.

doc/tutorial/08_tls_enforced.md
```diff
@@ -0,0 +1,91 @@
+## Preface 
+
+This tutorial will take you through the connection path from client to `pgagroal` to `postgres` server when TLS is enforced.
+
+## Setup
+
+To enforce tls along the whole path, we first need to create X509 certicates for client->pgagroal and pgagroal->postgres seperately.
```
jesperpedersen, line 7: tls -> TLS, client->pgagroal -> client to pgagroal, pgagroal->postgres -> pgagroal to PostgreSQL

doc/tutorial/08_tls_enforced.md
```diff
@@ -0,0 +1,91 @@
...
+ssl = on
+ssl_cert_file = </path/to/postgres.crt>
+ssl_key_file = </path/to/postgres.key>
+...
+```
+
+and make the contents of `pg_hba.conf` -
+
+```
+hostssl all all all md5
```
jesperpedersen, line 55: we should use scram-sha-256

doc/tutorial/08_tls_enforced.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+here we are choosing md5 for authenticating the requested user and database against postgres catalog
```
jesperpedersen, line 58: md5 -> scram-sha-256

jesperpedersen (changes requested):

jesperpedersen:
Can you add the 06, 07 and 08 tutorials to doc/GETTING_STARTED.md ?

jesperpedersen:
Maybe swap 07 and 08 around

doc/tutorial/08_tls_enforced.md
ashu3103: Will do

doc/tutorial/08_tls_enforced.md
ashu3103: The scram-sha-256 is giving `channal binding check failed` errors because we are switching ssl sessions between actual client and postgres server.

Is there some way to bypass it?

Or we will have to present same certificates on both sides (`c_ssl == s_ssl`)?

ashu3103:
> Can you add the 06, 07 and 08 tutorials to doc/GETTING_STARTED.md ?

Done!

ashu3103:
> Maybe swap 07 and 08 around

Done!, will commit all the changes all together

jesperpedersen:
We need to have `scram-sha-256` working in order for this feature to get merged.

You can start with the same certificate chain for the entire stack and go from there

ashu3103:
> We need to have `scram-sha-256` working in order for this feature to get merged.
> 
> You can start with the same certificate chain for the entire stack and go from there

This again is posing some problems as `postgres` certificate key should be owned by either postgres/root and we can't generalise this for all use cases.

We have to think of some other work around of this

jesperpedersen:
Yes, the certificates should be owned by the user running PostgreSQL, and the user running pgagroal.

So, we need a setup with a TLS chain that contains the certificates that we need

ashu3103:
#### Quick Questions
Channel binding is a feature of SCRAM-SHA-256-PLUS right?
Or does it happen in both SCRAM-SHA-256 and SCRAM-SHA-256-PLUS

#### Progress
Actually our client is picking SCRAM-SHA-256-PLUS which induce channel binding and cause the error - `channal binding check failed`.

**I have tried couple of things to mitigate it!!**

1) No matter what client is always picking SCRAM-SHA-256-PLUS as a mechanism even if `channel_binding`=`disable`  is enforced.

2) I have also tried to make a custom `AuthenticationSASL` message at `pgagroal` -
```
msg->kind = `R`
msg->length = 23
pgagroal_write_byte(msg->data, 'R')
pgagroal_write_int32(msg->data + 1, 23)
pgagroal_write_int32(msg->data+5, 10)
pgagroal_write_string(msg->data + 9, "SCRAM-SHA-256")
pgagroal_write_byte(msg->data + strlen("SCRAM-SHA-256") + 9, '\0')
```

But apparently client is not accepting it 🥲

Can anyone of you help me on this @jesperpedersen @fluca1978

jesperpedersen:
We don't support channel binding - don't know how hard it would be to extend the current implementation.

https://www.postgresql.org/docs/current/sasl-authentication.html
https://datatracker.ietf.org/doc/html/rfc5802#page-14
https://www.postgresql.org/docs/devel/protocol-message-formats.html
AuthenticationSASL (B)

jesperpedersen:
See 2), and pgmoneta_create_auth_scram256_response() - that should force the non-PLUS version... but apparently the server is trying to upgrade...

jesperpedersen:
https://github.com/agroal/pgagroal/blob/master/doc/DEVELOPERS.md#rebase

ashu3103:
I was actually doing #485, didn't notice it got affected, this issue is still in progress so I am thinking to deal with this after #485
I have taken all the changes in another local branch will make a separate PR from different branch with `scram-sha-256` capabilities working.

Issue #467, opened 2024-09-16
Remove deprecated CLI commands


jesperpedersen:

MohanadKh03:
Is this available to work on ? 
I have ran through the codebase and found out that the deprecated commands can be removed from the ```command_table``` without having any effect on the command parser

Obviously it should be added in the documentations as well that the deprecated commands have been replaced and not used at all starting the current release version (not sure if it will be released in 2.0.0 or not ?) and not used anymore

So is this how it should be done or there are more details that I am missing about this ?

jesperpedersen:
No, you can just remove everything - code and documentation.

Remember to add yourself to `AUTHORS`

PR #476, opened 2024-10-10
[#467] Remove deprecated CLI commands

MohanadKh03:
Removed the deprecated CLI commands from the commands table and from the related docs

Not sure if all the related docs have been modified though 
``pgagroal-cli`` in https://github.com/agroal/pgagroal/tree/master/doc/man probably needs to change because it is outdated but now that these are going to be deprecated so they must change also right ? I saw that it hasnt been changed for a few months now , Only ``manual`` has changed so im not sure if this has to change as well


jesperpedersen:
Merged.

Thanks for your contribution, and welcome to our community !

Issue #465, opened 2024-09-16
Replace cJSON with in-tree implementation


jesperpedersen:

jesperpedersen:
@MohanadKh03 Do you want to take this one ?

You can see how to use the in-tree JSON API in the https://github.com/pgmoneta/pgmoneta project

MohanadKh03:
Sure
Just to make sure I understand this though, everything that uses cJSON API should instead use the native in-tree parser API like how pgmoneta does. Right ?

jesperpedersen:
Correct, so you search for all usage of cJSON and replace it

PR #477, opened 2024-10-11
[#465] Replace cJSON dependancy with the native in-tree json API

MohanadKh03:
Replaced cJSON completely with the native json parser.
Tested manually using the following commands **from the docs user guide section 6.4 JSON Output Format**

``` pgagroal-cli conf ls --format json ```

``` pgagroal-cli conf set max_connections 1000 --format json ```

``` pgagroal-cli status --format json ```

``` pgagroal-cli ping --format json ```

Though I noticed that when using text format it gave a status of 1 and exits for some reason ? 
After debugging I think the problem is that some exit status is not what it should be I think ? I am kind of stuck in it cuz now it's only working with json formats not text somehow

### CODE REVIEW

**Comment by jesperpedersen** [comment]
Start by putting the definitions into `MANAGEMENT_ARGUMENT_XYZ` defines.

Also, look at the return values from the functions so f.ex. `.._json_create` should result into a `goto error:` if it fails.

But, you are close on this !

**Comment by MohanadKh03** [comment]
Ok so appearantly the issue was the way printing using text was coupled with cJSON which would have wasted a lot of time and effort to debug on it so instead I used the native API's way -->  ``pgagroal_json_print_and_free_json_object`` and renamed it cuz it will have both json and text **and possibly any other format in the future ?**

As for the definitions , not sure if I got this right but I tried moving the functions to the `management.h` file though they are static so do you mean to remove them being static or move some other functions ?

Added another commit , I know I should squash  but for the sake of readability since this has a lot of changes I let it like this till all is good then ill squash
WDYT ?

**Comment by jesperpedersen** [comment]
See CI - always init to `NULL` then assign after

**Comment by jesperpedersen** [comment]
@MohanadKh03 See https://github.com/agroal/pgagroal/issues/466 which is next, and how it is done in pgmoneta.

You can just follow `cli.c` -> `management.c` -> `main.c` -> ?

**Comment by jesperpedersen** [comment]
@MohanadKh03 You can do both at the same time if that is easier... and you want to

**Comment by MohanadKh03** [comment]
It's all good I will fix the memory leaks issue and continue with this one before getting to the 466

**Comment by jesperpedersen** [comment]
Maybe it depends on the path of the code in question, and likely the Mac builds are using much older software.

I use clang 18.x w/ Debug to catch most of them - valgrind is too slow

**Comment by jesperpedersen** [comment]
... and GitHub have seemed to update their platform yesterday so more reports are coming in now - I'm guessing the `ubuntu-latest` image is different now

File: src/libpgagroal/management.c [hunk 1]
```diff
@@ -1236,42 +1258,52 @@ pgagroal_management_read_isalive(SSL* ssl, int socket, int* status, char output_
...
+   }
+   else
+   {
+      pgagroal_json_put(output, "message", (uintptr_t)  "unknown", ValueString);
    }
 
+   // do I need to provide JSON output?
+   if (output_format == COMMAND_OUTPUT_FORMAT_JSON)
+   {
+      return pgagroal_print_and_free_object(json,FORMAT_JSON);
```
fluca1978, line 1290: I'm not sure this function should not have `json` in its name. While I agree that we could have more formats in the future, the function will always *print from a json object* and *free a json object*. The printing could be in YAML, as an example, but we start and free always a json struct, right?

src/libpgagroal/management.c
```diff
@@ -2195,24 +2226,19 @@ pgagroal_management_read_config_get(int socket, char* config_key, char* expected
...
-
-   // if here, print out in text format
-   cJSON* output = pgagroal_json_extract_command_output_object(json);
-   cJSON* value = cJSON_GetObjectItemCaseSensitive(output, "value");
-   cJSON* key = cJSON_GetObjectItemCaseSensitive(output, "key");
-   if (verbose)
+   else if(output_format == COMMAND_OUTPUT_FORMAT_TEXT)
    {
-      printf("%s = %s\n", key->valuestring, value->valuestring);
+      pgagroal_json_print(json,FORMAT_TEXT);
```
fluca1978, line 2235: This is much too information according to me.
If `verbose` is specified, the previous version of the application was printing the result and the exit status, e.g.:
```
% pgagroal-cli conf get max_connections --verbose
max_connections = 10
pgagroal-cli: Success (0)
```

Now it is printing out the whole json, that is somehow confusing and difficult to parse:

```
% pgagroal-cli conf get max_connections --verbose
application: 
  major: 2
  minor: 0
  name: pgagroal
  patch: 0
  version: 2.0.0
command: 
  error: 0
  exit-status: 0
  name: conf get
  output: 
    key: max_connections
    value: 10
  status: OK
pgagroal-cli: Success (0)

```

fluca1978 (changes requested):
I think the work is very good, but the main point I don't like at all is the printing out the whole json stuff in a text tree when the output is specified as text and the verbosity flag is turned on. While text output is not supposed to be stable, do we need to append such a bulk of information?

Second, I'm not sure the idea naming "possibly handling other formats in the future" functions without the json suffix really works: as far as I understand we will use json as the backbone format from now on, so make it clear a function works on json even when is dealing with another format.

src/libpgagroal/management.c
jesperpedersen: I think it about getting the `conf get | set` ported to pgmoneta, and then we can work from there

@Jubilee101 Cc:

jesperpedersen: https://github.com/pgmoneta/pgmoneta/issues/401

MohanadKh03: True I actually thought that by ``verbose`` it meant literally ``verbose`` to print out everything out but now I get it
The second commit is actually more of a trial-and-error tbh but ill rewrite the logic of text format again starting the first commit as a fresh one now that I understand most of the logic

jesperpedersen: @MohanadKh03 Maybe we should start with pgmoneta's 401 to figure out the best way to extract a single value, or set it...

Because the changes will likely be part of pgagroal's 466 anyway

fluca1978: > True I actually thought that by `verbose` it meant literally `verbose` to print out everything out but now I get it The second commit is actually more of a trial-and-error tbh but ill rewrite the logic of text format again starting the first commit as a fresh one now that I understand most of the logic Though in `status details` for example isnt it normal to have all the details so the whole object should be printed out right ?

Well, `details` is a very rich command, and it is text printed in a kind of tabular data. We could simplify the output by simply printing out the json object, but I would not change the current text format too much in order to not "scary" users that could have been scripting on the text output.

That's also why `--verbose` should not be the implementation of a `dump` command, according to me.

MohanadKh03: @jesperpedersen After reading pgmoneta's way imo it's much easier to read and maintain tbh **compared to the current pgagroal's way**

So do you suggest holding this PR off until 401 in pgmoneta is done and then we can standardize the IO,Management in both pgmoneta and pgagroal ?

jesperpedersen: @MohanadKh03 Yes, the management part should be unified across pgagroal, pgmoneta and pgexporter.

@decarv is taking care of the I/O part

PR #483, opened 2024-11-04
[#481] Add better error message about `Invalid master key file`

### LINKED ISSUE #481: Specify the file which cannot be decrypted when `invalid master key file` error is raised
When `pgagroal` starts, if it cannot decrypt a `xxx_users.conf` file information, it raises the `Invalid master key file` error.
This does not help in understand which file cannot be decrypted, hence it does not help in understanding which file has to be "adjusted" (via for example `pgagroal-admin`).
On the other hand, if one `xxx_users.conf` file is not found, the error message clearly shows which file is missing.

I propose to uniform the error messages, so that if a file cannot be decrypted, the error message (and optionally the systemd status message) includes the target file.

fluca1978:
When a users file, (frontend, admins or users files) cannot be decrypted, a generic `Invalid master key file` was raised. This commit adds the target file that cannot be decrypted, so that the user can understand which file has to be adjusted (e.g., `pgagroal-admin` usage on such file).

Makes the error messages consistent with those about a missing file.

Close #481

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -525,7 +525,7 @@ read_users_path:
 
          snprintf(message, MISC_LENGTH, "Invalid master key file");
 #ifdef HAVE_LINUX
-         sd_notifyf(0, "STATUS=%s: %s", message, users_path);
+         sd_notifyf(0, "STATUS=%s: <%s>", message, users_path);
 #endif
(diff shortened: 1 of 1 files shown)


jesperpedersen (approved):

jesperpedersen:
:+1:

Issue #463, opened 2024-09-14
Core APIs migration


Jubilee101:
Introduce value system, migrate JSON, deque and ART from [pgmoneta](https://github.com/pgmoneta/pgmoneta).

Jubilee101:
Deque, ART and json are imported. I didn't replace cJSON in management.c to avoid breaking stuff. But it can be removed once management is refactored. What were once in json.c/h are move to management.c as static functions since they are only used there. I'll keep this issue open, so any questions just let me know.

Jubilee101:
Note that we found a small bug about indentation when printing in text format. It's fixed but I want to watch for a while in case something else comes up. If you are interested, check out [pgmoneta#380](https://github.com/pgmoneta/pgmoneta/issues/380). We try to make the text format yaml-like but it turns out to be a little harder than json.

PR #486, opened 2024-12-01
[#463] New features for values, deque and ART

Jubilee101:
* minor behavior adjustments for noop_to_string_cb
* minor fix when adding a tagged node to deque
* ValueRef now prints out pointer address
* deque_exists: if a node with certain tag exists
* deque_remove: remove all nodes with a certain tag
* deque_sort: sort the deque based on tag
* add_with_config: allow customized to_string and destroy callback when adding a node to deque/ART

### DIFF
--- a/src/include/art.h
+++ b/src/include/art.h
@@ -74,7 +74,7 @@ int
 pgagroal_art_create(struct art** tree);
 
 /**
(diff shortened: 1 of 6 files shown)


jesperpedersen:
Merged.

PR #494, opened 2025-01-17
[#417] Add -F/--format flag for pgagroal-admin

shikharish:
cc: @jesperpedersen

### DIFF
--- a/src/include/management.h
+++ b/src/include/management.h
@@ -66,6 +66,7 @@ extern "C" {
 /**
  * Management commands
  */
+#define MANAGEMENT_UNKNOWN             0
 #define MANAGEMENT_CANCEL_SHUTDOWN     1
 #define MANAGEMENT_CONFIG_LS           2
 #define MANAGEMENT_CONFIG_GET          3
@@ -84,6 +85,11 @@ extern "C" {
 #define MANAGEMENT_STATUS             16
 #define MANAGEMENT_SWITCH_TO          17
 
+#define MANAGEMENT_MASTER_KEY         18
+#define MANAGEMENT_ADD_USER           19
+#define MANAGEMENT_UPDATE_USER        20
+#define MANAGEMENT_REMOVE_USER        21
+#define MANAGEMENT_LIST_USERS         22
 /**
  * Management arguments
  */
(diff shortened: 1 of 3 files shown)


src/include/management.h
```diff
@@ -84,6 +85,11 @@ extern "C" {
 #define MANAGEMENT_STATUS             16
 #define MANAGEMENT_SWITCH_TO          17
 
+#define MANAGEMENT_MASTER_KEY         18
+#define MANAGEMENT_ADD_USER           19
+#define MANAGEMENT_UPDATE_USER        20
+#define MANAGEMENT_REMOVE_USER        21
+#define MANAGEMENT_LIST_USERS         22
```
fluca1978, line 92: A matter of taste, but isn't it better to rename these as `MANAGEMENT_USER_XXX` such as `MANAGEMENT_USER_ADD`? It would be a nice grouping of related options.

src/libpgagroal/management.c
```diff
@@ -78,12 +73,12 @@ pgagroal_management_request_flush(SSL* ssl, int socket, int32_t mode, char* data
    struct json* j = NULL;
    struct json* request = NULL;
 
-   if (create_header(MANAGEMENT_FLUSH, compression, encryption, output_format, &j))
+   if (pgagroal_management_create_header(MANAGEMENT_FLUSH, compression, encryption, output_format, &j))
```
fluca1978, line 76: Is this related to this patch? Should be into another PR?
Also true for the following ones that are not related to user management,.

src/libpgagroal/management.c
```diff
@@ -632,8 +627,8 @@ pgagroal_management_request_conf_set(SSL* ssl, int socket, char* config_key, cha
    return 1;
 }
 
-static int
-create_header(int32_t command, uint8_t compression, uint8_t encryption, int32_t output_format, struct json** json)
+int
+pgagroal_management_create_header(int32_t command, uint8_t compression, uint8_t encryption, int32_t output_format, struct json** json)
```
fluca1978, line 631: As commented before, this rename should be into another PR according to me.

src/libpgagroal/management.c
```diff
@@ -680,8 +675,8 @@ create_header(int32_t command, uint8_t compression, uint8_t encryption, int32_t
    return 1;
 }
 
-static int
-create_request(struct json* json, struct json** request)
+int
+pgagroal_management_create_request(struct json* json, struct json** request)
```
fluca1978, line 679: ditto

src/libpgagroal/management.c
```diff
@@ -705,8 +700,8 @@ create_request(struct json* json, struct json** request)
    return 1;
 }
 
-static int
-create_outcome_success(struct json* json, time_t start_time, time_t end_time, struct json** outcome)
+int
+pgagroal_management_create_outcome_success(struct json* json, time_t start_time, time_t end_time, struct json** outcome)
```
fluca1978, line 704: ditto

src/libpgagroal/management.c
```diff
@@ -741,8 +736,8 @@ create_outcome_success(struct json* json, time_t start_time, time_t end_time, st
    return 1;
 }
 
-static int
-create_outcome_failure(struct json* json, int32_t error, struct json** outcome)
+int
+pgagroal_management_create_outcome_failure(struct json* json, int32_t error, struct json** outcome)
```
fluca1978, line 740: ditto

fluca1978 (changes requested):
I think the renaming of management functions, while fine, should not be placed into this PR.

Moreover, should we update the documentation about the new -F flag for `admin`?

jesperpedersen:
I'm ok with having these changes together

jesperpedersen:
Rebased, and merged.

PR #495, opened 2025-01-18
libpgagroal: utils.c: fix UB in pgagroal_read_int32()

decarv:
In `pgagroal_read_int32`, the byte[i] value is promoted to int32. According to the C spec, if you cannot represent the result in the result type, then that behavior is undefined (6.5.7).

The compiler was able to catch this after recent additions to cmake file, producing the runtime error:

libpgagroal/utils.c:319:38: runtime error: left shift of 166 by 24 places cannot be represented in type 'int'

This issue is fixed by explicitly casting the bytes[i] values to uint32_t.


jesperpedersen:
Merged.

Issue #496, opened 2025-01-18
Build type 'Release' is failing


decarv:
**Describe the bug**

The build is failing with 'Release' type.

**To Reproduce**

```bash
 $ git log --oneline -1
92ac986 (HEAD -> origin/master) [#417] Add -F/--format flag for pgagroal-admin
 ~/pgagroal/build ((HEAD detached at origin/master))
 $ cmake .. -DCMAKE_BUILD_TYPE=Release
-- The C compiler identification is GNU 14.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- pgagroal 2.0.0
-- Build type is Release
-- System is Linux
-- Performing Test COMPILER_SUPPORTS_C17
-- Performing Test COMPILER_SUPPORTS_C17 - Success
-- Found ZLIB: /usr/lib64/libz.so (found version "1.3.1")
-- zlib found
-- Found BZip2: /usr/lib64/libbz2.so (found version "1.0.8")
-- Looking for BZ2_bzCompressInit
m-- Looking for BZ2_bzCompressInit - found
-- bzip2 found
-- Found Zstd: /usr/lib64/libzstd.so
-- zstd found
-- Found Lz4: /usr/lib64/liblz4.so
-- lz4 found
-- Found Libev: /usr/lib64/libev.so (found suitable version "4.33", minimum required is "4.11")
-- libev found
-- Found OpenSSL: /usr/lib64/libcrypto.so (found version "3.2.3")
[...]
-- Generating done (0.0s)
-- Build files have been written to: /home/hc/pgagroal/build
 ~/pgagroal/build ((HEAD detached at origin/master))
 $ make
[  0%] Built target man
Generating API documentation
[  0%] Built target api
[...]
[ 92%] Building C object src/CMakeFiles/pgagroal-admin-bin.dir/admin.c.o
In function ‘list_users’,
    inlined from ‘main’ at /home/hc/pgagroal/src/admin.c:279:11:
/home/hc/pgagroal/src/admin.c:1162:7: error: pointer ‘users_file’ may be used after ‘fclose’ [-Werror=use-after-free]
 1162 |       fclose(users_file);
      |       ^~~~~~~~~~~~~~~~~~
/home/hc/pgagroal/src/admin.c:1131:4: note: call to ‘fclose’ here
 1131 |    fclose(users_file);
[...]

PR #497, opened 2025-01-18
[#496] admin.c: fix UAF in list_users()

decarv:
The fclose(users_file) in the error path can still be called after the first fclose(users_file).

Fix this by setting users_file to NULL.

This makes the build with type Release fail.


jesperpedersen:
Merged.

PR #507, opened 2025-03-04
network.c: cleanup in case connect() fails in pgagroal_connect_unix_s…

decarv:
…ocket()

The fd never gets closed when connect() fails inside pgagroal_connect_unix_socket(). Fix this by calling pgagroal_disconnect(fd).


fluca1978:
Looks good to me, but we have failures (that seems not related to this) in the CI.

jesperpedersen:
Rebased, and merged.

PR #509, opened 2025-03-05
ci.yml: Fix Linux CI

decarv:
Linux CI started to break due to new postgres version being installed with apt. Fix this by correcting the postgres data directory. Also download ca-certificates and install pgdg repository keys.

*This does not fix macOS CI* because CI is failing on secure_getenv() function. This function is present on FreeBSD 14+ but apparently not on macOS: https://man.freebsd.org/cgi/man.cgi?getenv(3).

Our options are to either disable CI, use HAVE_LINUX macros and getenv() function for macOS (which I think is bad unless we are supporting macOS) or we use this https://github.com/marketplace/actions/freebsd-vm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,16 @@ jobs:
       run: sudo apt install -y postgresql
     - name: Start postgres
       run: |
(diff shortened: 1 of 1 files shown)


decarv:
I decided to merge from another branch other than my master.

Issue #512, opened 2025-03-08
pgagraol_linux_version


jesperpedersen:
Create a `pgagroal_linux_version` function that extracts the kernel version, as `int`s (`major`, `minor` and `patch`).

If the function is called on other platforms it should return `0`s

Userfrom1995:
Hello, I am interested in working on this issue. Please assign it to me. Looking forward to contributing. Thank you!

fluca1978:
I would suggest to create a `pgagroal_os_kernel_version` that accepts also a `char*` to return if the platform is running Linux or not (BSD for example).

PR #513, opened 2025-03-09
[#512] Added pgagroal_linux_version function to extract Linux kernel Version.

Userfrom1995:
This PR introduces the function pgagroal_os_kernel_version() to retrieve the operating system name and kernel version as integers (major, minor, patch).

Key Features:

Utilizes uname() to extract system information.

Supports Linux, FreeBSD, OpenBSD, and macOS (Darwin).

Returns "Unknown" for unsupported platforms.

Logs the OS name and kernel version if available.

Handles errors gracefully and logs debug messages if retrieval fails.

Let me know if any refinements are required!

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -586,6 +586,19 @@ pgagroal_server_state_as_string(signed char state);
 char*
 pgagroal_connection_state_as_string(signed char state);
 
+/**
+ * Get the OS name and kernel version.
+ *
+ * @param os            Pointer to store the OS name (e.g., "Linux", "FreeBSD", "OpenBSD").
+ *                      Memory will be allocated internally and should be freed by the caller.
+ * @param kernel_major  Pointer to store the kernel major version.
+ * @param kernel_minor  Pointer to store the kernel minor version.
+ * @param kernel_patch  Pointer to store the kernel patch version.
+ * @return              0 on success, 1 on error.
+ */
+int
+pgagroal_os_kernel_version(char** os, int* kernel_major, int* kernel_minor, int* kernel_patch);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -46,6 +46,7 @@
 #include <openssl/pem.h>
 #include <sys/types.h>
 #include <err.h>
+#include <sys/utsname.h>
 
 #ifndef EVBACKEND_LINUXAIO
 #define EVBACKEND_LINUXAIO 0x00000040U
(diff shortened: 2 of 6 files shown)


src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,34 @@ pgagroal_escape_string(char* str)
...
+   *minor = 0;
+   *patch = 0;
+
+#if defined(HAVE_LINUX)
+   struct utsname buffer;
+
+   if (uname(&buffer) != 0)
+   {
+      pgagroal_log_warn("Failed to retrieve Linux kernel version.");
+      return -1;   // Error retrieving system info
```
jesperpedersen, line 1320: Error is `1` - not `-1`

src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,34 @@ pgagroal_escape_string(char* str)
...
+   }
+
+   if (sscanf(buffer.release, "%d.%d.%d", major, minor, patch) < 2)
+   {
+      pgagroal_log_warn("Failed to parse Linux kernel version.");
+      return -1;   // Parsing error
```
jesperpedersen, line 1326: Do an `error` label

src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,34 @@ pgagroal_escape_string(char* str)
...
+   }
+
+   pgagroal_log_info("Linux Kernel version: %d.%d.%d", *major, *minor, *patch);
```
jesperpedersen, line 1329: debug level

src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,34 @@ pgagroal_escape_string(char* str)
...
+
+#else
+   pgagroal_log_info("Kernel version not available (non-Linux system)");
```
jesperpedersen, line 1332: debug level

jesperpedersen (changes requested):

jesperpedersen:
See if you can get system information from *BSD, and if we can do something similar.

This can be a separate pull request though

jesperpedersen:
@decarv Cc: as you need this function to restrict to Linux 6.1+

src/main.c
```diff
@@ -918,6 +919,8 @@ main(int argc, char** argv)
 
    pgagroal_set_proc_title(argc, argv, "main", NULL);
 
+   pgagroal_linux_version(&major, &minor, &patch);
```
fluca1978, line 922: What is the aim of this usage? Only reporting on the logs?

src/include/utils.h
```diff
@@ -586,6 +586,16 @@ pgagroal_server_state_as_string(signed char state);
...
 
+/**
+ * Get the Linux kernel version
+ * @param major Pointer to store major version
+ * @param minor Pointer to store minor version
+ * @param patch Pointer to store patch version
+ * @return 0 if success, -1 if error
+ */
+int
+pgagroal_linux_version(int* major, int* minor, int* patch);
```
fluca1978, line 597: Why not a `pgagroal_os_lernel_version( char* os, int* major, int *minor, int* patch)` and then implement the function so that it can detect also if the system is a BSD one and put it into `os`, filling the other values accordingly?
FreeBSD and OpenBSD us only two digits, with differnt meanings, so `patch` would be always zero or null.

fluca1978 (changes requested):
I would try to extract the operating system and the kernel version. Besides, this could be useful to be included into the command protocol to know from which machine a message is coming from.

src/main.c
unconventional-guy: Yes, that was the only purpose.

Userfrom1995: Yes, that was the only purpose.

src/include/utils.h
```diff
@@ -586,6 +586,17 @@ pgagroal_server_state_as_string(signed char state);
...
+/**
+ * Get the OS name and kernel version
+ * @param os Pointer to store the OS name (e.g., "Linux", "FreeBSD", "OpenBSD")
+ * @param kernel_major Pointer to store the kernel major version
+ * @param kernel_minor Pointer to store the kernel minor version
+ * @param kernel_patch Pointer to store the kernel patch version
+ * @return 0 if success, 1 if error
+ */
+int
+pgagroal_os_kernel_version(char* os, int* kernel_major, int* kernel_minor, int* kernel_patch);
```
jesperpedersen, line 598: Should be `char** os`

jesperpedersen (changes requested):

src/include/utils.h
fluca1978: Yes, `char **`

src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,66 @@ pgagroal_escape_string(char* str)
...
+   return 0;
+
+error:
+   snprintf(os, 64, "Unknown");
+
+   pgagroal_log_debug("Unable to retrieve OS and kernel version.\n");
+
+   *kernel_major = 0;
+   *kernel_minor = 0;
+   *kernel_patch = 0;
```
fluca1978, line 1359: any chance one piece has been set and the others have not? why clearing them out again?

src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,66 @@ pgagroal_escape_string(char* str)
...
+   return 1;
+
+#else
+   snprintf(os, 64, "Unknown");
```
fluca1978, line 1363: `pgagroal_append` if used before

src/main.c
```diff
@@ -344,6 +344,8 @@ main(int argc, char** argv)
    struct main_configuration* config = NULL;
    int ret;
    int c;
+   char os[64];
```
fluca1978, line 347: should we use a `#define` for this repeasted `64`? and `malloc` it as a `char *` instead of an array?

fluca1978 (changes requested):
Nice, I think it is almost done

src/libpgagroal/utils.c
Userfrom1995: if (sscanf(buffer.release, "%d.%d.%d", kernel_major, kernel_minor, kernel_patch) < 2) this is the reason I am setting them. As the situation can happen rarely, the major version might get set while minor and patch remain uninitialized. Now this ensures that we always return a well-defined value, as almost all the Linux, BSD, and Darwin systems provide at least major and minor versions, but maybe a parsing issue can cause an unexpected format to be encountered during parsing that can cause some inconsistencies. To prevent them or to deal with them, I am doing that.

src/libpgagroal/utils.c
```diff
@@ -1302,3 +1303,66 @@ pgagroal_escape_string(char* str)
...
+
+   pgagroal_log_debug("OS: %s | Kernel Version: %d.%d.%d\n", os, *kernel_major, *kernel_minor, *kernel_patch);
+   return 0;
+
+error:
+   snprintf(os, 64, "Unknown");
+
+   pgagroal_log_debug("Unable to retrieve OS and kernel version.\n");
+
+   *kernel_major = 0;
```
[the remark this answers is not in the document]
Userfrom1995: yes , just to handle the case where major gets set to something due to a parsing issue or something. As all the systems we are writing this function for has atleast major and minor.

jesperpedersen:
Remember to squash

jesperpedersen:
I think you did something wrong

jesperpedersen:
Merged.

jesperpedersen:
@decarv Its ready

@Userfrom1995 Can you port this patch to pgmoneta and pgexporter, please ?

Userfrom1995:
> pgmoneta

Yes, I’ll work on porting the patch to pgmoneta and pgexporter. I’ll make the necessary adjustments and raise separate PRs for both projects.

jesperpedersen:
@Userfrom1995 Yes, issues and PRs

jesperpedersen:
@Userfrom1995 Think about what else you can do with this function... Expose in Prometheus ?

Userfrom1995:
> @Userfrom1995 Think about what else you can do with this function... Expose in Prometheus ?

Yes, I think we can expose this as a metric in Prometheus, which would make it a useful metric to track. We can also use integers to represent the OS, such as 0 for unknown, 1 for Linux, 2 for Mac, 3 for FreeBSD, and 4 for OpenBSD. This will make it easier to track and identify the types of machines we are using.

jesperpedersen:
@Userfrom1995 Just use the OS string directly

Userfrom1995:
> @Userfrom1995 Just use the OS string directly
"Okay, understood, should i  start working on this ?

jesperpedersen:
@Userfrom1995 Just create an issue for it, and then you can start

PR #520, opened 2025-03-17
Use secure_getenv or getenv based on OS for compatibility

Annupamaa:
Implemented a macro that automatically chooses getenv on Apple platforms and secure_getenv on other systems to ensure macOS compatibility

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -384,6 +384,12 @@ master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_for
       warnx("Could not write to master key file <%s>", &buf[0]);
       goto error;
    }
+   
+   #if defined(HAVE_OSX)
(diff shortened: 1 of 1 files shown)


src/admin.c
```diff
@@ -384,6 +384,11 @@ master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_for
       warnx("Could not write to master key file <%s>", &buf[0]);
       goto error;
    }
+      #if defined(HAVE_OSX)
+      #define GET_ENV(name) getenv(name)
```
fluca1978, line 388: Please run `uncrustify.sh` before committing.

Also, should we rename the macro to `PGAGROAL_GETENV`?

fluca1978 (changes requested):

jesperpedersen:
Merged - remember to reference the issue number in the commit message.

Issue #521, opened 2025-03-20
Add a directory configuration to handle the position of all the files


fluca1978:
So far, the user can specify every single file on the command line by means of individual switches, like `-c`, `-a`, `-l` and so on.
It would be nice to have a single switch to specify a directory, like `-D` and let `pgagroal` to find the files by its own within such directory.
So for example:
```
pgagroal -d -D /etc/pgagroal
```

would let to `-c /etc/pgagroal/pgagroal.conf` and so on. On the other hand

```
pgagroal -d -D /etc/pgagroal -c /usr/local/etc/pgagroal.conf
````

will use the specified configuration file and let `pgagroal` to "guess" the others in the `-D` directory. In other words, a single file switch overrides the directory one.

**Comment by jesperpedersen** [comment]
👍 

Since `/etc/pgagroal/` is default should we do a `WARN` in case `-D /etc/pgagroal/` is specified ?

**Comment by fluca1978** [comment]
> Since `/etc/pgagroal/` is default should we do a `WARN` in case `-D /etc/pgagroal/` is specified ?

Good idea! We should take care of the default directory set on the operating system and warn the user so that she could omit it.

**Comment by Numbers0689** [comment]
@fluca1978 Hi, I would like to work on this issue as part of my gsoc 2025 pre-project for pgagroal. please assign it to me.

**Comment by jesperpedersen** [comment]
Related to #487

**Comment by Numbers0689** [comment]
I'm almost done but while running `uncrustify.sh`, I get `unknown option` errors. At first I was using uncrustify 0.80, then I switched to 0.74 like mentioned in `uncrustify.cfg` but still results in `unknown option` errors with a different option.

```
uncrustify.cfg:2237: unknown option 'nl_var_def_blk_end_func_top'
uncrustify.cfg:3180: unknown option 'pp_space_after'

```
I'd like any pointers on this, whether I should be using a different version, commnt those options out or anything.

jesperpedersen:
Using 0.78.1 here - or use the tool, and verify your patch

PR #523, opened 2025-03-22
[#521] Add directory configuration option

Numbers0689:
Changes: 

- Implements 
#521, 
#487 
- Support for setting configuration directory using -D flag, while configuration files passed with individual flags take precedence over directory configuration files.
- Adds fallback configuration path.
- Add `WARN` in case default directory is passed as argument. [#521(comment)](https://github.com/agroal/pgagroal/issues/521#issuecomment-2739689989)
- Update relevant docs and CLI help messages.

also a couple files that changed after running uncrustify have been added.
 
Let me know if any changes are required. 

an `INFO` to log the precedence of flags wasn't added. [#521(comment)](https://github.com/agroal/pgagroal/issues/521#issuecomment-2739803410)

### DIFF
--- a/src/include/configuration.h
+++ b/src/include/configuration.h
@@ -132,6 +132,16 @@ int
 pgagroal_init_configuration(void* shmem);
 
 /**
+ * Check and set directory path
+ * @param directory_path Directory to search for path
+ * @param default_path Default path to file
+ * @param path Pointer to the path to be set
+ * @return 0 upon success, otherwise 1
+ */
+int
+pgagroal_normalize_path(char* directory_path, char* default_path, char** path);
+
+/**
  * Initialize the vault configuration structure
  * @param shmem The shared memory segment
  * @return 0 upon success, otherwise 1
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -177,6 +177,59 @@ pgagroal_init_configuration(void* shm)
    return 0;
 }
 
+int
+pgagroal_normalize_path(char* directory_path, char* default_path, char** path)
+{
+   char* dir_copy = NULL;
+   char* temp_path = NULL;
+   char* local_path = NULL;
+
+   if (*path != NULL || default_path == NULL)
+   {
+      return 0;
+   }
+
+   dir_copy = strdup(directory_path);
+   if (dir_copy == NULL)
+   {
+      goto error;
+   }
+
+   temp_path = pgagroal_append(dir_copy, default_path + strlen(PGAGROAL_DEFAULT_CONFIGURATION_PATH));
+   if (temp_path == NULL)
+   {
+      goto error;
+   }
+
+   if (access(temp_path, F_OK) == 0)
+   {
+      local_path = strdup(temp_path);
+
+      if (local_path == NULL)
+      {
+         goto error;
+      }
+
+      *path = local_path;
+   }
+   else
+   {
+      pgagroal_log_warn("pgagroal: %s not found", temp_path);
+   }
+
+   free(temp_path);
+   return 0;
+
+error:
+   if (temp_path == NULL)
+   {
+      free(dir_copy);
+   }
+   free(temp_path);
+   free(local_path);
+   return 1;
+}
+
 /**
  * This struct is going to store the metadata
  * about which sections have been parsed during
--- a/src/main.c
+++ b/src/main.c
@@ -303,6 +303,8 @@ usage(void)
    printf("                                     Default: %s\n", PGAGROAL_DEFAULT_ADMINS_FILE);
    printf("  -S, --superuser SUPERUSER_FILE     Set the path to the pgagroal_superuser.conf file\n");
    printf("                                     Default: %s\n", PGAGROAL_DEFAULT_SUPERUSER_FILE);
+   printf("  -D, --directory DIRECTORY_PATH     Set the directory path to load configuration files\n");
+   printf("                                     Default: %s\n", PGAGROAL_DEFAULT_CONFIGURATION_PATH);
    printf("  -d, --daemon                       Run as a daemon\n");
    printf("  -V, --version                      Display version information\n");
    printf("  -?, --help                         Display help\n");
@@ -321,6 +323,7 @@ main(int argc, char** argv)
    char* frontend_users_path = NULL;
    char* admins_path = NULL;
    char* superuser_path = NULL;
+   char* directory_path = NULL;
    bool daemon = false;
    pid_t pid, sid;
 #ifdef HAVE_LINUX
@@ -351,6 +354,25 @@ main(int argc, char** argv)
    char message[MISC_LENGTH]; // a generic message used for errors
    argv_ptr = argv;
 
+   struct stat path_stat = {0};
+   char* adjusted_dir_path = NULL;
+
+   char** config_paths[] = {
+      &configuration_path, &hba_path, &limit_path,
+      &users_path, &frontend_users_path, &admins_path, &superuser_path
+   };
+
+   char* default_files[] = {
+      PGAGROAL_DEFAULT_CONF_FILE, PGAGROAL_DEFAULT_HBA_FILE, PGAGROAL_DEFAULT_LIMIT_FILE,
+      PGAGROAL_DEFAULT_USERS_FILE, PGAGROAL_DEFAULT_FRONTEND_USERS_FILE,
+      PGAGROAL_DEFAULT_ADMINS_FILE, PGAGROAL_DEFAULT_SUPERUSER_FILE
+   };
+
+   const char* config_names[] = {
+      "configuration_path", "hba_path", "limit_path",
+      "users_path", "frontend_users_path", "admins_path", "superuser_path"
+   };
+
    while (1)
    {
       static struct option long_options[] =
@@ -362,13 +384,14 @@ main(int argc, char** argv)
          {"frontend", required_argument, 0, 'F'},
          {"admins", required_argument, 0, 'A'},
          {"superuser", required_argument, 0, 'S'},
+         {"directory", required_argument, 0, 'D'},
          {"daemon", no_argument, 0, 'd'},
          {"version", no_argument, 0, 'V'},
          {"help", no_argument, 0, '?'}
       };
       int option_index = 0;
 
-      c = getopt_long (argc, argv, "dV?a:c:l:u:F:A:S:",
+      c = getopt_long (argc, argv, "dV?a:c:l:u:F:A:S:D:",
                        long_options, &option_index);
 
       if (c == -1)
@@ -398,6 +421,9 @@ main(int argc, char** argv)
             break;
          case 'S':
             superuser_path = optarg;
+            break;
+         case 'D':
+            directory_path = optarg;
             break;
          case 'd':
             daemon = true;
@@ -436,6 +462,67 @@ main(int argc, char** argv)
 
    memset(&known_fds, 0, sizeof(known_fds));
    memset(message, 0, MISC_LENGTH);
+
+   if (directory_path != NULL)
+   {
+      if (!strcmp(directory_path, PGAGROAL_DEFAULT_CONFIGURATION_PATH))
+      {
+         pgagroal_log_warn("Using the default configuration directory %s, -D can be omitted.", directory_path);
+      }
+      if (access(directory_path, F_OK) != 0)
+      {
+#ifdef HAVE_LINUX
+         sd_notifyf(0, "STATUS=Configuration directory not found: %s", directory_path);
+#endif
+         pgagroal_log_error("Configuration directory not found: %s", directory_path);
+         exit(1);
+      }
+      if (stat(directory_path, &path_stat) == 0)
+      {
+         if (!S_ISDIR(path_stat.st_mode))
+         {
+#ifdef HAVE_LINUX
+            sd_notifyf(0, "STATUS=Path is not a directory: %s", directory_path);
+#endif
+            pgagroal_log_error("Path is not a directory: %s", directory_path);
+            exit(1);
+         }
+      }
+      if (access(directory_path, R_OK | X_OK) != 0)
+      {
+#ifdef HAVE_LINUX
+         sd_notifyf(0, "STATUS=Insufficient permissions for directory: %s", directory_path);
+#endif
+         pgagroal_log_error("Insufficient permissions for directory: %s", directory_path);
+         exit(1);
+      }
+      if (directory_path[strlen(directory_path) - 1] != '/')
+      {
+         adjusted_dir_path = pgagroal_append(strdup(directory_path), "/");
+      }
+      else
+      {
+         adjusted_dir_path = strdup(directory_path);
+      }
+      if (adjusted_dir_path == NULL)
+      {
+         pgagroal_log_error("Memory allocation failed while copying directory path.");
+         exit(1);
+      }
+
+      for (int i = 0; i < 7; i++)
+      {
+         if (pgagroal_normalize_path(adjusted_dir_path, default_files[i], config_paths[i]) != 0)
+         {
+#ifdef HAVE_LINUX
+            sd_notifyf(0, "STATUS=Failed to normalize path: %s%s", adjusted_dir_path, config_names[i]);
+#endif
+            pgagroal_log_error("Failed to normalize path: %s%s", adjusted_dir_path, config_names[i]);
+            exit(1);
+         }
+      }
+      free(adjusted_dir_path);
+   }
 
    // the main configuration file is mandatory!
    configuration_path = configuration_path != NULL ? configuration_path : PGAGROAL_DEFAULT_CONF_FILE;
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -18,6 +18,7 @@ Options:
   -F, --frontend FRONTEND_USERS_FILE Set the path to the pgagroal_frontend_users.conf file
   -A, --admins ADMINS_FILE           Set the path to the pgagroal_admins.conf file
   -S, --superuser SUPERUSER_FILE     Set the path to the pgagroal_superuser.conf file
+  -D, --directory DIRECTORY_PATH     Set the path to load configuration files
   -d, --daemon                       Run as a daemon
   -V, --version                      Display version information
   -?, --help                         Display help
--- a/doc/man/pgagroal.1.rst
+++ b/doc/man/pgagroal.1.rst
@@ -39,6 +39,9 @@ OPTIONS
 -S, --superuser SUPERUSER_FILE
   Set the path to the pgagroal_superuser.conf file
 
+-D, --directory DIRECTORY_PATH
+  Set the path to load configuration files
+
 -d, --daemon
   Run as a daemon
 
--- a/doc/manual/user-01-quickstart.md
+++ b/doc/manual/user-01-quickstart.md
@@ -19,6 +19,7 @@ Options:
   -F, --frontend FRONTEND_USERS_FILE Set the path to the pgagroal_frontend_users.conf file
   -A, --admins ADMINS_FILE           Set the path to the pgagroal_admins.conf file
   -S, --superuser SUPERUSER_FILE     Set the path to the pgagroal_superuser.conf file
+  -D, --directory DIRECTORY_PATH     Set the path to load configuration files
   -d, --daemon                       Run as a daemon
   -V, --version                      Display version information
   -?, --help                         Display help


Numbers0689:
I think the `pgagroal.h` macros for default and fallback paths can be changed a bit, where the file name strings can be separate macros but I wasn't sure of the preferred way

jesperpedersen:
@fluca1978 Can you take this one ?

fluca1978:
> @fluca1978 Can you take this one ?

I'm going to review this today.

src/include/configuration.h
```diff
@@ -131,6 +131,26 @@ extern "C" {
...
+ * @param path Pointer to the path to be set
+ * @param directory_path Directory to search for path
+ * @param default_path Default path
+ * @param fallback_path Fallback path if default not found
+ */
+void
+pgagroal_check_directory_path(char** path, const char* directory_path, const char* default_path, const char* fallback_path);
+
+/**
+ * Check and set directory path
```
fluca1978, line 145: The comment is the same as the previous function, it should be more clear in which way the two are different.

src/include/pgagroal.h
```diff
@@ -52,8 +52,10 @@ extern "C" {
 
 #ifdef HAVE_FREEBSD
     #define PGAGROAL_DEFAULT_CONFIGURATION_PATH "/usr/local/etc/pgagroal/"
+    #define PGAGROAL_FALLBACK_CONFIGURATION_PATH "/etc/pgagroal/"
```
fluca1978, line 55: Isn't the fallback path supposed to be the default one?

src/libpgagroal/configuration.c
```diff
@@ -178,6 +178,46 @@ pgagroal_init_configuration(void* shm)
    return 0;
 }
 
+/**
+ *
```
fluca1978, line 182: Remove comments, leave only in the header file or when a static function is defined.

src/libpgagroal/configuration.c
```diff
@@ -178,6 +178,46 @@ pgagroal_init_configuration(void* shm)
...
+ */
+void
+pgagroal_check_directory_path(char** path, const char* directory_path, const char* default_path, const char* fallback_path)
+{
+   char temp_path[MAX_PATH];
+
+   snprintf(temp_path, MAX_PATH, "%s/%s", directory_path, default_path + strlen(PGAGROAL_DEFAULT_CONFIGURATION_PATH));
+   if (*path == NULL && access(temp_path, F_OK) == 0)
```
fluca1978, line 190: I would first check if `path` is 'NULL` or not, then in case it is, do the `snprintf` and other checks.

src/libpgagroal/configuration.c
```diff
@@ -178,6 +178,46 @@ pgagroal_init_configuration(void* shm)
...
+
+/**
+ *
+ */
+char*
+pgagroal_check_path(char* path, char* default_path, char* fallback_path)
+{
+   if (path == NULL)
+   {
+      if (access(default_path, F_OK) == 0)
```
fluca1978, line 211: Should also check if `default_path` is null?

src/libpgagroal/configuration.c
```diff
@@ -178,6 +178,46 @@ pgagroal_init_configuration(void* shm)
...
+      *path = strdup(temp_path);
+   }
+}
+
+/**
+ *
+ */
+char*
+pgagroal_check_path(char* path, char* default_path, char* fallback_path)
+{
```
fluca1978, line 208: Probably this function should log which path is chosen.

src/main.c
```diff
@@ -437,8 +444,40 @@ main(int argc, char** argv)
...
+      if (access(directory_path, R_OK | X_OK) != 0)
+      {
+         snprintf(message, MISC_LENGTH, "Insufficient permissions for directory");
+#ifdef HAVE_LINUX
+         sd_notifyf(0, "STATUS=%s: %s", message, directory_path);
+#endif
+         errx(1, "%s: %s", message, directory_path);
+      }
+
+      pgagroal_check_directory_path(&configuration_path, directory_path, PGAGROAL_DEFAULT_CONF_FILE, PGAGROAL_FALLBACK_CONF_FILE);
```
fluca1978, line 470: These functions are actually choosing a path and overriding a variable, right? Probably the name of the functions should reflect it (e.g., `pgagroal_normalize_path` or something).

src/main.c
```diff
@@ -437,8 +444,40 @@ main(int argc, char** argv)
...
+      pgagroal_check_directory_path(&limit_path, directory_path, PGAGROAL_DEFAULT_LIMIT_FILE, PGAGROAL_FALLBACK_LIMIT_FILE);
+      pgagroal_check_directory_path(&users_path, directory_path, PGAGROAL_DEFAULT_USERS_FILE, PGAGROAL_FALLBACK_USERS_FILE);
+      pgagroal_check_directory_path(&frontend_users_path, directory_path, PGAGROAL_DEFAULT_FRONTEND_USERS_FILE, PGAGROAL_FALLBACK_FRONTEND_USERS_FILE);
+      pgagroal_check_directory_path(&admins_path, directory_path, PGAGROAL_DEFAULT_ADMINS_FILE, PGAGROAL_FALLBACK_ADMINS_FILE);
+      pgagroal_check_directory_path(&superuser_path, directory_path, PGAGROAL_DEFAULT_SUPERUSER_FILE, PGAGROAL_FALLBACK_SUPERUSER_FILE);
+   }
+
    // the main configuration file is mandatory!
-   configuration_path = configuration_path != NULL ? configuration_path : PGAGROAL_DEFAULT_CONF_FILE;
+   configuration_path = pgagroal_check_path(configuration_path, PGAGROAL_DEFAULT_CONF_FILE, PGAGROAL_FALLBACK_CONF_FILE);
```
fluca1978, line 480: This is not clear to me: if there is no path, the default is used, then a fallback one is used.
Why not simply the default one?

fluca1978 (changes requested):
Looks good, even if I don't understand well the usage of the fallback path, that seems like a last resort to the default path missing. I think having the default path could be enough, but also if we want to keep the fallback use something like `/etc/pgagroal.d` on Linux and `/usr/local/etc/pgagroal.d` on FreeBSD. Note that on OpenBSD things are placed similarly to Linux.

Moreover, if the directory path does not exist at all, no warning is issued and the user does not get a change to understand the application is starting from the default directory:


```
% pgagroal -D /tmp 

2025-03-24 11:43:10 DEBUG configuration.c:3114 PID file automatically set to: [/tmp/pgagroal.54333.pid]
2025-03-24 11:43:10 DEBUG utils.c:1359 OS: Linux | Kernel Version: 5.14.0
2025-03-24 11:43:10 INFO  main.c:1190 pgagroal: 2.0.0 started on localhost:54333
2025-03-24 11:43:10 DEBUG main.c:1193 Socket: 7
2025-03-24 11:43:10 DEBUG main.c:1193 Socket: 8
2025-03-24 11:43:10 DEBUG main.c:1195 Unix Domain Socket: 5
2025-03-24 11:43:10 DEBUG main.c:1196 Management: 3
2025-03-24 11:43:10 DEBUG main.c:1197 Transfer: 4
2025-03-24 11:43:10 DEBUG main.c:1200 Metrics: 12
2025-03-24 11:43:10 DEBUG main.c:1200 Metrics: 13
2025-03-24 11:43:10 DEBUG utils.c:466 libev available: select
2025-03-24 11:43:10 DEBUG utils.c:470 libev available: poll
2025-03-24 11:43:10 DEBUG utils.c:474 libev available: epoll
2025-03-24 11:43:10 DEBUG utils.c:482 libev available: iouring
2025-03-24 11:43:10 DEBUG main.c:1207 libev engine: epoll
2025-03-24 11:43:10 DEBUG main.c:1208 Pipeline: 0
2025-03-24 11:43:10 DEBUG main.c:1209 Pipeline size: 0
2025-03-24 11:43:10 DEBUG main.c:1210 OpenSSL 3.2.2 4 Jun 2024
2025-03-24 11:43:10 DEBUG main.c:1211 Configuration size: 418368
2025-03-24 11:43:10 DEBUG main.c:1212 Max connections: 20
2025-03-24 11:43:10 DEBUG main.c:1213 Known users: 3
2025-03-24 11:43:10 DEBUG main.c:1214 Known frontend users: 1
2025-03-24 11:43:10 DEBUG main.c:1215 Known admins: 1
2025-03-24 11:43:10 DEBUG main.c:1216 Known superuser: No
2025-03-24 11:43:10 DEBUG pool.c:1056 pgagroal_prefill
2025-03-24 11:43:10 DEBUG pool.c:174 connect: server 0
2025-03-24 11:43:10 DEBUG pool.c:213 connect: 127.0.0.1:5432 using slot 0 fd 7
```

src/main.c
Numbers0689: alright, I'll remove the fallback path.

jesperpedersen: yeah, we can just use the default - no need for fallback with `-D`

src/libpgagroal/configuration.c
Numbers0689: It will always be the default path macro, so can't be null right? or I could just use the macro directly inside the function and remove the `default_path` argument

jesperpedersen: Correct - you can use the constant directly

fluca1978: If the argument remains, I do prefer having the test. If the argument is removed and replaced with the macro itself, we can remove the test (and document that it is using the macro).

src/include/art.h
```diff
@@ -40,9 +40,9 @@ extern "C" {
 
 #define MAX_PREFIX_LEN 10
 
-typedef int (*art_callback)(void* data, const unsigned char* key, uint32_t key_len, struct value* value);
+typedef int (* art_callback)(void* data, const unsigned char* key, uint32_t key_len, struct value* value);
```
jesperpedersen, line 43: Watch out for unneeded changes

src/include/configuration.h
```diff
@@ -131,6 +131,15 @@ extern "C" {
 int
 pgagroal_init_configuration(void* shmem);
 
+/**
+ * Check and set directory path
+ * @param path Pointer to the path to be set
+ * @param directory_path Directory to search for path
+ * @param default_path Default path to file
+ */
+void
```
jesperpedersen, line 140: int

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,21 @@ pgagroal_init_configuration(void* shm)
    return 0;
 }
 
+void
+pgagroal_normalize_path(char** path, const char* directory_path, const char* default_path)
+{
+   if (*path == NULL && default_path != NULL)
```
jesperpedersen, line 183: You need an error path

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,21 @@ pgagroal_init_configuration(void* shm)
...
+   {
+      char temp_path[MAX_PATH];
+
+      snprintf(temp_path, MAX_PATH, "%s/%s", directory_path, default_path + strlen(PGAGROAL_DEFAULT_CONFIGURATION_PATH));
```
jesperpedersen, line 187: Use pgagroal_append

src/main.c
```diff
@@ -437,6 +444,38 @@ main(int argc, char** argv)
    memset(&known_fds, 0, sizeof(known_fds));
    memset(message, 0, MISC_LENGTH);
 
+   if (directory_path != NULL)
+   {
+      if (strcmp(directory_path, PGAGROAL_DEFAULT_CONFIGURATION_PATH) == 0)
```
jesperpedersen, line 449: !strcmp

jesperpedersen (changes requested):

src/libpgagroal/configuration.c
jesperpedersen: Output parameters should be last

jesperpedersen: Use a local variable to do the work with

jesperpedersen: So init first

`*path = NULL;`

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,38 @@ pgagroal_init_configuration(void* shm)
...
+            return 1;
+         }
+      }
+      else
+      {
+         pgagroal_log_debug("pgagroal: %s not found", temp_path);
+      }
+      free(temp_path);
+   }
+   return 0;
```
jesperpedersen, line 209: error label

src/main.c
```diff
@@ -437,8 +444,61 @@ main(int argc, char** argv)
...
+      {
+         pgagroal_log_warn("Using the default configuration directory %s, -D can be omitted.", directory_path);
+      }
+      if (access(directory_path, F_OK) != 0)
+      {
+         snprintf(message, MISC_LENGTH, "Configuration directory not found");
+#ifdef HAVE_LINUX
+         sd_notifyf(0, "STATUS=%s: %s", message, directory_path);
+#endif
+         errx(1, "%s: %s", message, directory_path);
```
jesperpedersen, line 459: We need to error out too

src/main.c
```diff
@@ -437,8 +444,61 @@ main(int argc, char** argv)
...
+      }
+      if (access(directory_path, R_OK | X_OK) != 0)
+      {
+         snprintf(message, MISC_LENGTH, "Insufficient permissions for directory");
+#ifdef HAVE_LINUX
+         sd_notifyf(0, "STATUS=%s: %s", message, directory_path);
+#endif
+         errx(1, "%s: %s", message, directory_path);
```
jesperpedersen, line 467: Same

src/main.c
```diff
@@ -437,8 +444,61 @@ main(int argc, char** argv)
...
+      }
+
+      char** config_paths[] = {
+         &configuration_path, &hba_path, &limit_path,
+         &users_path, &frontend_users_path, &admins_path, &superuser_path
+      };
+
+      char* default_files[] = {
```
jesperpedersen, line 475: Variable definitions at the top

src/main.c
```diff
@@ -437,8 +444,61 @@ main(int argc, char** argv)
...
+
+      for (int i = 0; i < 7; i++)
+      {
+         if (pgagroal_normalize_path(config_paths[i], directory_path, default_files[i]) != 0)
+         {
+            snprintf(message, MISC_LENGTH, "Failed to normalize path");
+#ifdef HAVE_LINUX
+            sd_notifyf(0, "STATUS=%s: %s", message, config_names[i]);
+#endif
+            errx(1, "%s: %s", message, config_names[i]);
```
jesperpedersen, line 494: Same

src/main.c
```diff
@@ -437,8 +444,61 @@ main(int argc, char** argv)
...
+         }
+      }
+   }
+
    // the main configuration file is mandatory!
    configuration_path = configuration_path != NULL ? configuration_path : PGAGROAL_DEFAULT_CONF_FILE;
+   memset(message, 0, MISC_LENGTH);
```
jesperpedersen, line 501: What is that ?

src/main.c
jesperpedersen: Log the message directly

jesperpedersen (changes requested):

jesperpedersen:
First round - apply across the patch

src/main.c
fluca1978: We need also to specify what type of path is not normalized, or the user will be confused.

fluca1978:
I've added a few comments on b9f40e0987f72a30c6f0160c19805a690c78f7c3.
Besides, it seems to me again it is not warning in case the specified path does not contain configuration:

```
% pgagroal -D /tmp        
2025-03-26 14:05:14 DEBUG configuration.c:3085 PID file automatically set to: [/tmp/pgagroal.54333.pid]
2025-03-26 14:05:14 DEBUG utils.c:1359 OS: Linux | Kernel Version: 5.14.0
2025-03-26 14:05:14 INFO  main.c:1211 pgagroal: 2.0.0 started on localhost:54333
...
```

Also, specifying a non directory has a wrong error message:

```
% pgagroal -D /tmp/err.log
pgagroal: Insufficient permissions for directory: /tmp/err.log
```

Numbers0689:
I'll work the changes :+1:

src/main.c
Numbers0689: should i remove the sd_notify in these cases or keep it? and log directly with just errx()? or use pgagroal_log_error?

jesperpedersen: Just go with `sd_notify` for now

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,59 @@ pgagroal_init_configuration(void* shm)
...
+   char* temp_path = NULL;
+   char* local_path = NULL;
+
+   if (*path != NULL || default_path == NULL)
+   {
+      return 0;
+   }
+
+   dir_copy = strdup(directory_path);
+   if (dir_copy == NULL)
```
jesperpedersen, line 193: Remove

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,59 @@ pgagroal_init_configuration(void* shm)
...
+      if (local_path == NULL)
+      {
+         goto error;
+      }
+
+      *path = local_path;
+   }
+   else
+   {
+      pgagroal_log_warn("pgagroal: %s not found", temp_path);
```
jesperpedersen, line 217: This is an error

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,59 @@ pgagroal_init_configuration(void* shm)
...
+   }
+
+   free(temp_path);
+   return 0;
+
+error:
+   if (temp_path == NULL)
```
jesperpedersen, line 224: This is confusing

src/main.c
```diff
@@ -437,6 +463,67 @@ main(int argc, char** argv)
...
 
+   if (directory_path != NULL)
+   {
+      if (!strcmp(directory_path, PGAGROAL_DEFAULT_CONFIGURATION_PATH))
+      {
+         pgagroal_log_warn("Using the default configuration directory %s, -D can be omitted.", directory_path);
+      }
+      if (access(directory_path, F_OK) != 0)
+      {
+#ifdef HAVE_LINUX
```
jesperpedersen, line 474: HAVE_SYSTEMD

jesperpedersen (changes requested):

src/libpgagroal/configuration.c
Numbers0689: I'll remove it, since the directory path is now dynamically allocated we can use `pgagroal_append` directly, my bad.

src/libpgagroal/configuration.c
Numbers0689: some of the configuration files are optional right? ahould I log as an error?

src/libpgagroal/configuration.c
```diff
@@ -177,6 +177,59 @@ pgagroal_init_configuration(void* shm)
...
+      return 0;
+   }
+
+   dir_copy = strdup(directory_path);
+   if (dir_copy == NULL)
+   {
+      goto error;
+   }
+
+   temp_path = pgagroal_append(dir_copy, default_path + strlen(PGAGROAL_DEFAULT_CONFIGURATION_PATH));
```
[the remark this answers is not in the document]
Numbers0689: we need the configuration file names. I can declare those names as separate macros to use here directly and update `PGAGROAL_DEFAULT_*_FILE` to use them

src/libpgagroal/configuration.c
jesperpedersen: No, you can log it under `DEBUG`

src/libpgagroal/configuration.c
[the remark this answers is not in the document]
jesperpedersen: Yes, use pgagroal_append with PGAGROAL_DEFAULT_CONFIGURATION_PATH and the default file names.

There is no `strlen` in pgagroal_append...

src/libpgagroal/configuration.c
Numbers0689: > This is confusing

its to avoid double free, since `pgagroal_append` reallocates the passed string and appends the other string, both `temp_path` and `dir_copy` are the same. if `temp_path` is NULL, dir_copy has to be freed, else freeing temp_path frees both.

src/libpgagroal/configuration.c
jesperpedersen: No, it has to be `INFO` by default

src/libpgagroal/configuration.c
jesperpedersen: Yes, but you need to look at your code -- it is confusing.

Make variables that tell what you are doing. `free()` them, copy them, ...

Will you remember what you did in 3 months ?

jesperpedersen:
@Numbers0689 How are you on this ?

Numbers0689:
> @Numbers0689 How are you on this ?

sorry I lost track of time. while using `pgagroal_append`, I kind of made things complicated with the unclear use of `free()` all over the place since the existing logic in `main` doesnt use dynamically allocated config_paths. I can make a commit with the improvements but I don't want to waste your time. please let me know if you'd like me to update with the changes on this or close it if you think it's better.

jesperpedersen:
@Numbers0689 No, just keep working on it. Of course, if you don't want to anymore then you can close this pull request, and we can unassign you

Numbers0689:
I'll continue working on it

Userfrom1995:
Hi @jesperpedersen, @fluca1978 

I've noticed that this PR (https://github.com/agroal/pgagroal/pull/523) has been open for a while without any recent updates. If no one is working on it anymore, can I take this one and get it completed? Or do you have any other plans regarding it?

jesperpedersen:
@Userfrom1995 Work from the existing patch

fluca1978:
> Hi @jesperpedersen, @fluca1978
> 
> I've noticed that this PR (#523) has been open for a while without any recent updates. If no one is working on it anymore, can I take this one and get it completed? Or do you have any other plans regarding it?

I think it is fine, but please give initial credits (if you re-use the code) to @Numbers0689

PR #535, opened 2025-04-02
[#536] feat: Improved macOS Build Workflow

### LINKED ISSUE #536: Improve macOS Build Workflow
## Is your feature request related to a problem? Please describe.
The current macOS build workflow in our GitHub Actions has several inefficiencies and redundancies that slow down the CI process. Specifically:

1. It unnecessarily installs Homebrew which is already pre-installed on GitHub runners
2. It runs duplicate builds with GCC and Clang, even though on macOS, GCC is just a symlink to Clang
3. It lacks dependency caching, causing repeated downloads of the same packages
4. The PostgreSQL installation and management is overly complex
5. The build process doesn't utilize parallel compilation

## Describe the solution you'd like
This PR streamlines the macOS build workflow by:

1. Removing the redundant Homebrew installation step
2. Consolidating to a single build with Clang (since GCC on macOS is just a symlink to Clang)
3. Adding caching for Homebrew packages to speed up subsequent builds
4. Simplifying PostgreSQL installation by using a specific version
5. Adding proper wait logic for PostgreSQL to start
6. Utilizing parallel make to speed up compilation
7. Improving error handling for pgagroal startup
8. Ensuring cleanup happens even if previous steps fail

## Additional information
The optimized workflow reduces build time while maintaining the same level of testing. It also provides better feedback when things go wrong by adding proper error handling. The workflow now follows GitHub Actions best practices by using caching and ensuring resources are properly cleaned up.

This change doesn't modify any functionality of the actual build process - it simply makes the CI workflow more efficient and reliable.

botirk38:
Fixes #536

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,95 +154,66 @@ jobs:
       working-directory: /home/runner/work/pgagroal/pgagroal/build/src/
 
 
-
-      #  build-macos:
-      #
-      #    runs-on: macos-latest
-      #
-      #    steps:
-      #    - uses: actions/checkout@v3
-      #    - name: Install Homebrew
-      #      run: /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
-      #    - name: Update system
-      #      run: brew update
-      #    - name: Install openssl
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

Issue #540, opened 2025-04-05
environment variables only support log_path


nesangcode:
The actual resolver is in the [log_file_open](https://github.com/agroal/pgagroal/blob/babd4c76be1521447713041a62c591f91b271986/src/libpgagroal/logging.c#L243), thus make the env works in `log_path`  only.

**Expected solution:**
- Integrates `pgagroal_resolve_path` into [extract_syskey_value](https://github.com/agroal/pgagroal/blob/babd4c76be1521447713041a62c591f91b271986/src/libpgagroal/configuration.c#L283).

nesangcode:
@fluca1978 @jesperpedersen

nesangcode:
Should we do the expected or any other solution?

fluca1978:
Clearly having the resolution available everywhere opens new possibilities.
I think we could do integrate, I don't see any path that should not take advantage of this.
Open a pull request and I'll test.

nesangcode:
Can you explain more about the idea? It already has been integrated into `extract_syskey_value`

fluca1978:
sorry, I was looking at the wrong link...

Shyam-123pandey:
Hi @fluca1978 sir,
I was looking through this issue and the discussion. It seems that the path resolution using `pgagroal_resolve_path()` has already been integrated into `extract_syskey_value`.
Since that change enables environment variable resolution more generally, I wanted to ask if there is still any work expected on this issue or if it is already considered resolved.
Thanks.

fluca1978:
> Hi [@fluca1978](https://github.com/fluca1978) sir, I was looking through this issue and the discussion. It seems that the path resolution using `pgagroal_resolve_path()` has already been integrated into `extract_syskey_value`. Since that change enables environment variable resolution more generally, I wanted to ask if there is still any work expected on this issue or if it is already considered resolved. Thanks.

I was waiting for a PR to test this in.
I think it has to be rebased and checked over, since time has passed. If you want to collaborate with @nesangcode you are welcome. @nesangcode would you like to complete this?

Shyam-123pandey:
> > Hi [@fluca1978](https://github.com/fluca1978) sir, I was looking through this issue and the discussion. It seems that the path resolution using `pgagroal_resolve_path()` has already been integrated into `extract_syskey_value`. Since that change enables environment variable resolution more generally, I wanted to ask if there is still any work expected on this issue or if it is already considered resolved. Thanks.
> 
[...]

PR #541, opened 2025-04-05
[#540] fix environment variables support

nesangcode:
Make full environment variable support

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -328,6 +328,13 @@ pgagroal_read_configuration(void* shm, char* filename, bool emit_warnings)
                key = NULL;
                value = NULL;
             }
+            else
+            {
+               free(key);
+               free(value);
+               key = NULL;
+               value = NULL;
+            }
          }
       }
    }
@@ -2162,7 +2169,8 @@ error:
 }
 
 /**
- * Given a line of text extracts the key part and the value.
+ * Given a line of text extracts the key part and the value
+ * and expands environment variables in the value (like $HOME).
  * Valid lines must have the form <key> = <value>.
  *
  * The key must be unquoted and cannot have any spaces
@@ -2225,6 +2233,8 @@ extract_syskey_value(char* str, char** key, char** value)
    // empty value
    if (c == length)
    {
+      free(k);
+      k = NULL;
       return 0;
    }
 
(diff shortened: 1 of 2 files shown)


nesangcode:
@jesperpedersen @fluca1978 So far, I don't understand what caused the memory leak. ASAN points [here](https://github.com/agroal/pgagroal/blob/babd4c76be1521447713041a62c591f91b271986/src/libpgagroal/configuration.c#L2240) but `v` is NULL, I don't understand what caused memory leak.

src/libpgagroal/configuration.c
```diff
@@ -328,6 +328,19 @@ pgagroal_read_configuration(void* shm, char* filename, bool emit_warnings)
                key = NULL;
                value = NULL;
             }
+            else
```
nesangcode, line 331: Got the solution, add `else` just in case one of `key` or `value` still exists, free it.

src/libpgagroal/configuration.c
```diff
@@ -2240,6 +2256,22 @@ extract_syskey_value(char* str, char** key, char** value)
       v = pgagroal_append_char(v, str[i]);
    }
 
+   char* resolved_path = NULL;
+
+   if (pgagroal_resolve_path(v, &resolved_path))
```
nesangcode, line 2261: Integrated `pgagroal_resolve_path` to syskey extractor.

src/libpgagroal/configuration.c
```diff
@@ -328,6 +328,19 @@ pgagroal_read_configuration(void* shm, char* filename, bool emit_warnings)
                key = NULL;
                value = NULL;
             }
+            else
+            {
+               if (key)
```
jesperpedersen, line 333: You don't need to guard a `free()`

src/libpgagroal/configuration.c
```diff
@@ -328,6 +328,19 @@ pgagroal_read_configuration(void* shm, char* filename, bool emit_warnings)
...
+               {
+                  free(key);
+                  key = NULL;
+               }
+               if (value)
```
jesperpedersen, line 338: Same

jesperpedersen (changes requested):

jesperpedersen:
Merged.

Issue #545, opened 2025-04-13
clang/ASAN: Improve coverage


bassamadnan:
Detect ASAN flags that are available during a Debug build depending on the clang version. Create a list of flags for each clang version, and then do a test if it is available if so enable.

bassamadnan:
Will be happy to take this up, removing warnings wherever possible.

jesperpedersen:
@bassamadnan Go for it !

PR #546, opened 2025-04-14
[#545] Add version coverage for clang and gcc Debug mode

bassamadnan:
Addressing #545  , The version details for different address sanitizer flags can be found here pgexporter/pgexporter#209 , as a refresher: 

1. `-fsanitize-address-use-after-scope`: helps detecting when a variable is accessed after it has gone out of scope such as when a pointer to a local variable continues to be used after the function returns or when a reference to a variable in an inner scope is accessed after that scope ends.
2. `-fsanitize-recover=address`: continues execution after error detection.
3. `-Wextra` for extra warnings around operators

Ignored the sign compare flag, although did resolve warnings which  were trivial. Most of the ones left arise from macro usage or are unecessary and may cause bugs. I removed one conditional block, and there are a few instances which are just empty functions so all params go unused in those. Will highlight this in my next comment


src/libpgagroal/configuration.c
```diff
@@ -934,11 +934,6 @@ pgagroal_vault_validate_configuration (void* shm)
       return 1;
    }
 
-   if (config->common.authentication_timeout < 0)
-   {
-      config->common.authentication_timeout = DEFAULT_AUTHENTICATION_TIMEOUT;
-   }
-
```
bassamadnan, line 941: this was completely removed since `authentication_timeout` is unsigned long and would always evaluate to false.

bassamadnan: done, added it back here and changed it at one other place ( `<=` to `==`)

jesperpedersen:
Merged.

PR #564, opened 2025-06-05
[#563] Provide a warning message if user passwords do not match

### LINKED ISSUE #563: `pgagroal-admin` provide a message when password do not match
When adding and/or editing a user via `pgagroal-admin`, if the password edited interactively do not match, the program simply repropose the prompt `Password:` and this could be unclear to users.
It would be better to present the user with a warning message.

fluca1978:
While editing password interactively, `pgagroal-admin` provides a warning message in the case the entered password do not match.

Close #563

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -628,6 +628,7 @@ password:
    {
       if ((unsigned char)(*(password + i)) & 0x80)
       {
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@fluca1978 LGTM, merge when ready

Issue #570, opened 2025-06-16
Fix: Add Two-Dot Notation Support to pgagroal-cli conf get Command


Userfrom1995:
Enhanced the pgagroal-cli conf get command to support two-dot notation (section.context.key) for precise configuration parameter retrieval, enabling commands like pgagroal-cli conf get limit.pgbench.max_size while maintaining backward compatibility with existing single-word queries.

jesperpedersen:
@Userfrom1995 Please, do ports for pgmoneta and pgexporter as well

PR #580, opened 2025-06-30
[#570] Prevent segfault when accessing scalar config values with dotted keys

Userfrom1995:
Previously, querying a scalar configuration value (e.g., "port" or "max_connections") with a dotted key (like "port.a") would cause a segmentation fault. This happened because the JSON API returned a scalar value cast as a pointer, which was then incorrectly treated as a JSON object.

This commit adds a defensive check:
    if ((uintptr_t)configuration_js < 0x1000)
to ensure that only valid JSON object pointers are passed to the iterator logic. If a scalar or invalid pointer is detected, the code now gracefully returns an error instead of crashing.

Before Changes:

```c
base@LAPTOP-409T8FMM:~/pgagroal/build/src$ ./pgagroal-cli conf get port.a
AddressSanitizer:DEADLYSIGNAL
=================================================================
==511894==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000929 (pc 0x7f4f2bda0d06 bp 0x7fffb21ca380 sp 0x7fffb21ca350 T0)
==511894==The signal is caused by a READ memory access.
==511894==Hint: address points to the zero page.
    #0 0x7f4f2bda0d06 in pgagroal_json_iterator_create /home/base/pgagroal/src/libpgagroal/json.c:234
    #1 0x55dd97c3297c in get_config_key_result /home/base/pgagroal/src/cli.c:1695
    #2 0x55dd97c304fe in process_get_result /home/base/pgagroal/src/cli.c:1393
    #3 0x55dd97c2f8c7 in conf_get /home/base/pgagroal/src/cli.c:1265
    #4 0x55dd97c2dbb8 in main /home/base/pgagroal/src/cli.c:810
    #5 0x7f4f2b39e1c9 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #6 0x7f4f2b39e28a in __libc_start_main_impl ../csu/libc-start.c:360
    #7 0x55dd97c28ed4 in _start (/home/base/pgagroal/build/src/pgagroal-cli+0x11ed4) (BuildId: e996bc25e118f998a6d30b30cf98bc58ed976b2f)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/base/pgagroal/src/libpgagroal/json.c:234 in pgagroal_json_iterator_create
==511894==ABORTING
```

After Changes :
```c
base@LAPTOP-409T8FMM:~/pgagroal/build/src$ ./pgagroal-cli conf get port.a
Error
2025-06-30 14:18:01 TRACE network.c:428 pgagroal_disconnect: fd=3
```

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -1612,6 +1612,8 @@ get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t ou
    struct json* outcome = NULL;
    struct json_iterator* iter;
    char* config_value = NULL;
+   enum value_type section_type;
+   uintptr_t section_data;
    int part_count = 0;
    char* parts[4] = {NULL, NULL, NULL, NULL}; // Allow max 4 to detect invalid input
    char* token;
@@ -1716,7 +1718,16 @@ get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t ou
 
    if (strlen(section) > 0)
    {
-      configuration_js = (struct json*)pgagroal_json_get(response, section);
+      section_data = pgagroal_json_get_typed(response, section, &section_type);
+
+      pgagroal_log_debug("Section '%s' has type: %s", section, pgagroal_value_type_to_string(section_type));
+
+      if (section_type != ValueJSON)
+      {
+         goto error;
+      }
+
+      configuration_js = (struct json*)section_data;
       if (!configuration_js)
       {
          goto error;
@@ -1727,7 +1738,11 @@ get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t ou
       configuration_js = response;
    }
 
-   pgagroal_json_iterator_create(configuration_js, &iter);
+   if (pgagroal_json_iterator_create(configuration_js, &iter))
+   {
+      goto error;
+   }
+
    while (pgagroal_json_iterator_next(iter))
    {
       // Handle three-part keys: section.context.key


fluca1978:
Do we have a better approach to understand if a result is an invalid object?

Userfrom1995:
I’m not sure if it's necessarily better, but I can think of a few more possible approaches to handle this:

- Type Checking During Iteration
    Iterate over the parent JSON and check each value’s type before accessing it. This avoids trying to iterate or dereference scalar values. ( the one that i just pushed ).

- Wrap Scalars in a Uniform JSON Structure
    Adjust the JSON API response so even scalar values are wrapped in a consistent structure (e.g., { "type": "scalar", "value": 42 }). This makes type checks safer and more predictable.

-  Refactor the CLI Logic
    Update the CLI’s get logic to explicitly handle scalars vs. objects, avoiding generic iteration and handling each case appropriately.

jesperpedersen:
@Jubilee101 Cc:

fluca1978:
> 
>     * Wrap Scalars in a Uniform JSON Structure
>       Adjust the JSON API response so even scalar values are wrapped in a consistent structure (e.g., { "type": "scalar", "value": 42 }). This makes type checks safer and more predictable.

This is what we should go for, because it makes the interface consistent (everything is an object) and stable.
Thoughts?

Jubilee101:
> This is what we should go for, because it makes the interface consistent (everything is an object) and stable.
Thoughts?

Second this. The value type system is designed on the assumption that users know what types they had put in when they try reading it out. So all the outcome should perhaps be json.

Userfrom1995:
I explored the idea of modifying the implementation to wrap scalar values before returning them from `pgagroal_json_get()`, so that we always return a JSON object rather than a raw scalar. Initially, this sounded like a clean way to handle things more uniformly and avoid issues like the segmentation fault we saw.

But after looking into it more deeply, I think this approach would introduce unnecessary complexity.

Right now, the code assumes that we know what we’re putting in and getting out — which is why we do things like:

    id = (int32_t)pgagroal_json_get(header, MANAGEMENT_ARGUMENT_COMMAND);

If we start wrapping scalars, we’d need to:
  - Change all such direct casts.
  - Add utility functions to unwrap the actual values from the wrapper.
  - Handle additional memory management for the wrapper objects.

That alone would touch a lot of code and introduce risk.

There is a second approach I considered — modifying the `pgagroal_json_put` logic so that we always wrap scalars when inserting them into the JSON. That way, we ensure that the structure is consistent for all fields. Here's a comparison between the **current** and **modified** JSON structures:

---  
**(1) Current JSON Structure:**
```c
{
  "Header": {
    "ClientVersion": "2.0.0",
    "Command": 3,
    "Compression": 0,
    "Encryption": 0,
    "Output": 0,
    "Timestamp": "20250715184659"
  },
  "Request": {}
}
```
---

**(2) Modified JSON Structure (wrapped scalars):**
```c
{
  "Header": {
    "ClientVersion": {
      "value": "2.0.0"
    },
    "Command": {
      "value": 3
    },
    "Compression": {
      "value": 0
    },
    "Encryption": {
      "value": 0
    },
    "Output": {
      "value": 0
    },
    "Timestamp": {
      "value": "20250715191154"
    }
  },
  "Outcome": {
    "Error": {
      "value": 2
    },
    "Status": {
      "value": false
    }
  },
  "Request": {}
}
```
---

This modified format keeps things consistent and simplifies value extraction logic across the board.

However, it still comes with some drawbacks:
  - The JSON becomes more verbose and harder to read.
  - We’d still need utility functions to extract values from the `value` field.
  - It would require refactoring all `put` and `get` calls across the codebase — which is a big and risky change.

In the end, this feels like a large change to fix a relatively small issue (a segmentation fault), which already has a simpler workaround (like the iterative approach I proposed earlier). So from a practicality standpoint, I’d suggest sticking with the current implementation.

That said, if we do want to move forward with modifying the JSON APIs, we should first have a more detailed discussion:
  - Should we modify the `get` path, the `put` path, or both?
  - How big is the impact across the codebase?
  - Is there a cleaner or smaller alternative we’re overlooking?

If we decide to go this route, it should definitely be done in a separate PR, as it touches core behavior.

Let me know your thoughts — I’m happy to help implement the change if we agree on a direction.

jesperpedersen:
@Userfrom1995 I don't like it - it shouldn't be necessary to wrap all JSON values

jesperpedersen:
@Userfrom1995 What is important is to create a chapter / appendix in the developer that describes the format (name/type) and so on

fluca1978:
@Userfrom1995 this is an excellent and detailed analysis, and while I think wrapping is still the right path to get things done, it requires too many changes so far, and probably has also a runtime impact we don't want for a few scalar values.
Therefore, we should step back to the iterating approach you proposed.
Do you think it is worth providing a `pgagroal_json_get_scalar` defensive function to clearly indicate we want to extract a single value out of JSON?

jesperpedersen:
@Jubilee101 will look at it - so hold

jesperpedersen:
@Userfrom1995 See #592

Userfrom1995:
> Do you think it is worth providing a `pgagroal_json_get_scalar` defensive function to clearly indicate we want to extract a single value out of JSON?

I think it will be a good addition. It will provide a defensive mechanism that will help us prevent the wrong casting of non-scalar type pointers to scalar pointers or scalar variables.

For example, if we try to cast a JSON pointer to a scalar, it can cause issues - it can cause memory issues, segmentation faults, and other problems. So we can implement the `pgagroal_json_get_scalar` function in such a way that it returns NULL or zero for non-scalar values or in the case the tag does not exist. So before casting, we can check and make sure we don't cast JSON or any non-scalar into a scalar. So it will be a good defensive practice, but that's the only advantage - it will not help us in any other way. It just provides a defensive mechanism. Almost anything else can be achieved with the `pgagroal_json_get_typed` function.

We can do two kinds of implementation of the scalar functions: `pgagroal_json_get_scalar` and `pgagroal_json_get_scalar_typed`.

## 1. `pgagroal_json_get_scalar()` - Simple Defensive Version

```c
uintptr_t
pgagroal_json_get_scalar(struct json* item, char* tag)
{
   enum value_type found_type;
   uintptr_t data;
   
   if (item == NULL || item->type != JSONItem || tag == NULL || strlen(tag) == 0)
   {
      return 0;  // Return NULL/0 for safety
   }
   
   data = pgagroal_art_search_typed(item->elements, tag, &found_type);
   
   // Defensive check: reject complex types - SAFETY MECHANISM
   if (found_type == ValueJSON || found_type == ValueDeque || found_type == ValueART || found_type == ValueNone) 
   {
      return 0;  // Return NULL/0 to prevent dangerous casting
   }
   
   return data;
}
```

## 2. `pgagroal_json_get_scalar_typed()` - Defensive + Type Info

```c
uintptr_t
pgagroal_json_get_scalar_typed(struct json* item, char* tag, enum value_type* type)
{
   enum value_type found_type;
   uintptr_t data;
   
   if (item == NULL || item->type != JSONItem || tag == NULL || strlen(tag) == 0)
   {
      *type = ValueNone;
      return 0;
   }
   
   data = pgagroal_art_search_typed(item->elements, tag, &found_type);
   
   // Defensive check: reject complex types AND handle not found
   if (found_type == ValueJSON || found_type == ValueDeque || found_type == ValueART || found_type == ValueNone) 
   {
      *type = ValueNone;
      return 0;
   }
   
   *type = found_type;
   return data;
}
```

## Usage Examples

Simple Defensive Version:
```c
// Safe scalar extraction - prevents segfaults
uintptr_t data = pgagroal_json_get_scalar(json_obj, "port");
if (data == 0) {
    printf("Error: 'port' is not a scalar or doesn't exist\n");
    return;
}
int32_t port = (int32_t)data;  // Safe to cast
```

Typed Version for More Control:
```c
// Safe scalar extraction with type info
enum value_type type;
uintptr_t data = pgagroal_json_get_scalar_typed(json_obj, "port", &type);
if (type == ValueNone) {
    printf("Error: 'port' is not a scalar\n");
    return;
}

switch (type) {
    case ValueInt32:
        int32_t port = (int32_t)data;
        break;
    case ValueString:
        char* hostname = (char*)data;
        break;
}
```

These functions provide safety mechanisms that prevent dangerous casting of complex types to scalars, which can cause segmentation faults and memory corruption issues.

Userfrom1995:
> Do you think it is worth providing a `pgagroal_json_get_scalar` defensive function to clearly indicate we want to extract a single value out of JSON?

I think it will be a good addition. It will provide a defensive mechanism that will help us prevent the wrong casting of non-scalar type pointers to scalar pointers or scalar variables.

For example, if we try to cast a JSON pointer to a scalar, it can cause issues - it can cause memory issues, segmentation faults, and other problems. So we can implement the `pgagroal_json_get_scalar` function in such a way that it returns NULL or zero for non-scalar values or in the case the tag does not exist. So before casting, we can check and make sure we don't cast JSON or any non-scalar into a scalar. So it will be a good defensive practice, but that's the only advantage - it will not help us in any other way. It just provides a defensive mechanism. Almost anything else can be achieved with the `pgagroal_json_get_typed` function.

We can do two kinds of implementation of the scalar functions: `pgagroal_json_get_scalar` and `pgagroal_json_get_scalar_typed`.

## 1. `pgagroal_json_get_scalar()` - Simple Defensive Version

```c
uintptr_t
pgagroal_json_get_scalar(struct json* item, char* tag)
{
   enum value_type found_type;
   uintptr_t data;
   
   if (item == NULL || item->type != JSONItem || tag == NULL || strlen(tag) == 0)
   {
      return 0;  // Return NULL/0 for safety
   }
   
   data = pgagroal_art_search_typed(item->elements, tag, &found_type);
   
   // Defensive check: reject complex types - SAFETY MECHANISM
   if (found_type == ValueJSON || found_type == ValueDeque || found_type == ValueART || found_type == ValueNone) 
   {
      return 0;  // Return NULL/0 to prevent dangerous casting
   }
   
   return data;
}
```

## 2. `pgagroal_json_get_scalar_typed()` - Defensive + Type Info

```c
uintptr_t
pgagroal_json_get_scalar_typed(struct json* item, char* tag, enum value_type* type)
{
   enum value_type found_type;
   uintptr_t data;
   
   if (item == NULL || item->type != JSONItem || tag == NULL || strlen(tag) == 0)
   {
      *type = ValueNone;
      return 0;
   }
   
   data = pgagroal_art_search_typed(item->elements, tag, &found_type);
   
   // Defensive check: reject complex types AND handle not found
   if (found_type == ValueJSON || found_type == ValueDeque || found_type == ValueART || found_type == ValueNone) 
   {
      *type = ValueNone;
      return 0;
   }
   
   *type = found_type;
   return data;
}
```

## Usage Examples

Simple Defensive Version:
```c
// Safe scalar extraction - prevents segfaults
uintptr_t data = pgagroal_json_get_scalar(json_obj, "port");
if (data == 0) {
    printf("Error: 'port' is not a scalar or doesn't exist\n");
    return;
}
int32_t port = (int32_t)data;  // Safe to cast
```

Typed Version for More Control:
```c
// Safe scalar extraction with type info
enum value_type type;
uintptr_t data = pgagroal_json_get_scalar_typed(json_obj, "port", &type);
if (type == ValueNone) {
    printf("Error: 'port' is not a scalar\n");
    return;
}

switch (type) {
    case ValueInt32:
        int32_t port = (int32_t)data;
        break;
    case ValueString:
        char* hostname = (char*)data;
        break;
}
```

These functions provide a helpful safety mechanism to prevent dangerous casting of complex types (like nested JSON) into scalars, which can easily lead to segmentation faults or memory corruption. That said, they don’t completely eliminate the risk of miscasting, and they don’t remove the need to know what type of data you’re expecting.

We still need to either know exactly what scalar type we are working with, or perform a type check using the type returned by `pgagroal_json_get_typed` or `pgagroal_json_get_scalar_typed` before casting. What these functions really do is reduce the chances of a specific class of bugs—particularly those where a non-scalar is mistakenly treated as a scalar.

Jubilee101:
> We still need to either know exactly what scalar type we are working with, or perform a type check using the type returned by pgagroal_json_get_typed or pgagroal_json_get_scalar_typed before casting.

Feel free to implement some helper functions to filter out the scalar types (don't know if it's appropriate to call them that, since string is included here). But I don't think it's needed to make this a public API. After all this is just to condone the fact that we are using the same key for different types.

Jubilee101:
The current patch looks good to me @jesperpedersen

fluca1978:
Looks good to me, except that I would not protect the debug logging statement.
Using the typed function is the correct way to go.

Userfrom1995:
> Looks good to me, except that I would not protect the debug logging statement. Using the typed function is the correct way to go.

Won't work without debug condition because `pgagroal_value_type_to_string` is defined inside a debug statement have a look here https://github.com/agroal/pgagroal/blob/d6c7e845ab6c140ea05f4ee3b7257f7d69f5b636/src/libpgagroal/value.c#L315
Have a look at this CI run :
https://github.com/agroal/pgagroal/actions/runs/16402230693/job/46343296315

fluca1978:
> > Looks good to me, except that I would not protect the debug logging statement. Using the typed function is the correct way to go.
> 
> Won't work without debug condition because `pgagroal_value_type_to_string` is defined inside a debug statement have a look here
> 
> https://github.com/agroal/pgagroal/blob/d6c7e845ab6c140ea05f4ee3b7257f7d69f5b636/src/libpgagroal/value.c#L315
> 

My fault, sorry!

fluca1978:
@jesperpedersen PTAL

src/cli.c
```diff
@@ -1716,7 +1718,18 @@ get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t ou
 
    if (strlen(section) > 0)
    {
-      configuration_js = (struct json*)pgagroal_json_get(response, section);
+      section_data = pgagroal_json_get_typed(response, section, &section_type);
+
+      #ifdef DEBUG
```
jesperpedersen: Remove

jesperpedersen (changes requested):

jesperpedersen:
@Userfrom1995 @Jubilee101 will fix the `#ifdef DEBUG` in the core api

Userfrom1995:
I have removed the `#ifdef DEBUG`, but CI will fail for all release builds.  
We can just remove the guard `#ifdef DEBUG` around the function `pgagroal_value_type_to_string` in this PR or in a separate PR — I think that would be the best approach, since I don’t see any other use of this function anywhere else in the code. As far as I can tell, this is the first time we are using it.

jesperpedersen:
@Userfrom1995 It will be done separate

jesperpedersen:
@Userfrom1995 Then we will bring in your patch

jesperpedersen:
Rebased, and merged.

Issue #581, opened 2025-07-04
pgagroal-vault crashes on startup with AddressSanitizer errors when attempting to initialize the event loop.


Userfrom1995:
**Describe the bug**

pgagroal-vault crashes on startup with AddressSanitizer errors when attempting to initialize the event loop. The vault process fails to start due to memory access violations in the event loop setup code, which incorrectly assumes a main configuration structure when accessing vault configuration.

**To Reproduce**

Just try to start the vault with valid configuration.

**Version**

pgagroal version: 1.7.0 (current development branch).

**PostgreSQL**

PostgreSQL version: 17.x (issue is independent of PostgreSQL version)
**liburing**

liburing version: 2.3+ (issue occurs with any version).

**OpenSSL**
OpenSSL version: 3.0.x (issue occurs regardless of TLS configuration)

**Access method**

Not applicable (crash occurs before authentication)

**OS**

Operating System: Ubuntu 22.04 LTS (Linux), but issue affects all supported platforms

**ulimit**
```c
real-time non-blocking time  (microseconds, -R) unlimited
core file size              (blocks, -c) 0
data seg size               (kbytes, -d) unlimited
scheduling priority                 (-e) 0
file size                   (blocks, -f) unlimited
pending signals                     (-i) 31178
max locked memory           (kbytes, -l) 65536
max memory size             (kbytes, -m) unlimited
open files                          (-n) 1024
pipe size                (512 bytes, -p) 8
POSIX message queues         (bytes, -q) 819200
real-time priority                  (-r) 0
stack size                  (kbytes, -s) 8192
cpu time                   (seconds, -t) unlimited
max user processes                  (-u) 31178
virtual memory              (kbytes, -v) unlimited
file locks                          (-x) unlimited
```

**Configuration**

[pgagroal-vault]
host = localhost
port = 2500
log_type = console
log_level = debug5
log_path = /tmp/pgagroal-vault.log

[main]
host = localhost
port = 2347
user = admin

**Debug logs**
```c
=================================================================
==625362==ERROR: AddressSanitizer: unknown-crash on address 0x7f1920098af8 at pc 0x7f192126594d bp 0x7ffd210d5eb0 sp 0x7ffd210d5ea0
READ of size 4 at 0x7f1920098af8 thread T0
    #0 0x7f192126594c in setup_ops /home/base/pgagroal/src/libpgagroal/ev.c:159
[...]

PR #582, opened 2025-07-04
[#581] Add ev_backend support to vault

Userfrom1995:
This PR implements execution context tracking to fix pgagroal-vault startup crashes by safely distinguishing between [vault_configuration] and [main_configuration] structures during event loop initialization, while adding ev_backend configuration support to enable I/O performance optimization in the vault


jesperpedersen:
Merged.

PR #590, opened 2025-07-09
[#589] Add runtime validation for limit and HBA configuration on conf set

Userfrom1995:


jesperpedersen:
Merged.

Issue #595, opened 2025-07-22
Improve switch-to command error handling and logging


Userfrom1995:
# Problem
- switch-to always returns ok response (status=true) even when failing
- No specific error messages for different failure types
- Poor operational visibility for troubleshooting

PR #596, opened 2025-07-22
[#595] Improve switch-to command error handling and logging

Userfrom1995:


src/include/management.h
```diff
@@ -167,6 +167,8 @@ extern "C" {
 #define MANAGEMENT_ERROR_CONF_ALIAS_NETWORK                   1201
 #define MANAGEMENT_ERROR_CONF_ALIAS_ERROR                     1202
 
+#define MANAGEMENT_ERROR_SWITCH_TO_FAILED    25
```
jesperpedersen: This need to go into the 13xx section

src/libpgagroal/server.c
```diff
@@ -316,43 +316,73 @@ pgagroal_server_clear(char* server)
...
+      pgagroal_log_debug("pgagroal: Current primary is '%s', but target '%s' does not exist",
+                         config->servers[old_primary].name, server);
+      return 1;
+   }
+   else
+   {
 
-   return 1;
+      pgagroal_log_warn("pgagroal: Switch to server '%s' failed: no current primary server found and target server not found in configuration", server);
+      pgagroal_log_debug("pgagroal: System has no primary server and target '%s' does not exist", server);
```
fluca1978: is the log needed since it has the same information of the previous line, which in turn has an higher log value?

fluca1978 (approved):

src/libpgagroal/server.c
Userfrom1995: I’ll remove the pgagroal_log_debug statement.

jesperpedersen:
Merged.

Issue #597, opened 2025-07-22
Standardize JSON Output Format for conf ls Command


Userfrom1995:
# Summary
- conf ls --format json returns inconsistent JSON structure compared to other CLI commands
- Server already sends standard format, but client strips it using special processing
- Fix: Use standard process_result() instead of process_ls_result()
- Remove unused functions: process_ls_result() and get_conf_path_result()

PR #598, opened 2025-07-22
[#597] Standardize JSON Output Format for conf ls Command

Userfrom1995:


jesperpedersen:
@fluca1978 Can you take this one ?

jesperpedersen:
@Userfrom1995 Looks like this is porting material for pgmoneta and pgexporter as well

fluca1978 (approved):
Approved

fluca1978:
What's the problem with FreeBSD CI? Then we can merge.

Userfrom1995:
> What's the problem with FreeBSD CI? Then we can merge.

This issue keeps happening from time to time.  
See a similar case here: https://github.com/vmactions/freebsd-vm/issues/91
I’ll look into it and start with some basic checks — for example, upgrading the FreeBSD version or adjusting parameters like RAM. If that doesn’t help, I’ll search for other solutions. And if none of these approaches work, we may eventually need to consider an alternative way to run FreeBSD CI.

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Userfrom1995 @fluca1978 The FreeBSD CI issue is separate from this, so we can look at it in another PR

PR #610, opened 2025-08-03
[#607] Add pgagroal-vault and pgagroal_vault.conf man page generation

Userfrom1995:


jesperpedersen:
@Userfrom1995 You need to update the pgagroal.spec file too -- search for existing documentation

Userfrom1995:
> @Userfrom1995 You need to update the pgagroal.spec file too -- search for existing documentation

I checked the spec file, and it already includes the vault man pages:
https://github.com/agroal/pgagroal/blob/8b3d176b1c883b635d2ff7bfa556c93af2beeb74/pgagroal.spec#L79

While building on my Fedora system, I found a couple of issues in the spec file. I’ll open a separate PR to address those.

Regarding the vault man pages specifically, I don’t think any updates are required. However, I may be missing something — please correct me if I’m wrong.

jesperpedersen:
Merged.

Issue #621, opened 2025-08-12
Fix TLS


Userfrom1995:
https://github.com/agroal/pgagroal/discussions/560

PR #622, opened 2025-08-12
[#621] Fix TLS

Userfrom1995:
The issue turned out to be that we were using the wrong SSL context and file descriptors in the send and receive functions. Both functions were hard-coded to always use the server’s SSL context and file descriptor. When these functions were called from the client side, they still attempted to use the server’s context and descriptor, which caused the problem.


jesperpedersen:
Merged.

Issue #625, opened 2025-08-18
Cannot acquire connections when pool is prefilled (false “pool full”)


Userfrom1995:
Unable to connect when all connections are prefilled for a database in the configuration; the pool reports “full” even though all connections are free.

```c
2025-08-18 08:09:15 DEBUG worker.c:301 After client: PID 90624 Slot -1 (1)
2025-08-18 08:09:15 DEBUG main.c:2226 pgagroal: Transfer client done
2025-08-18 08:09:15 DEBUG main.c:2236 pgagroal: Transfer client done: PID 90624
2025-08-18 08:09:27 DEBUG security.c:236 SSL request from client: 9
2025-08-18 08:09:27 DEBUG pool.c:1357 find_best_rule: user=myuser, database=mydb -> rule 0 (real_database=mydb)
2025-08-18 08:09:27 DEBUG pool.c:1421 Connection count for rule 0 (user=myuser): 2 connections
2025-08-18 08:09:27 DEBUG pool.c:1421 Connection count for rule 0 (user=myuser): 2 connections
2025-08-18 08:09:27 DEBUG pool.c:1421 Connection count for rule 0 (user=myuser): 2 connections
2025-08-18 08:09:27 DEBUG pool.c:1421 Connection count for rule 0 (user=myuser): 2 connections
2025-08-18 08:09:27 DEBUG pool.c:1421 Connection count for rule 0 (user=myuser): 2 connections
2025-08-18 08:09:27 DEBUG pool.c:1421 Connection count for rule 0 (user=myuser): 2 connections
2025-08-18 08:09:27 DEBUG security.c:364 authenticate: pool is full
2025-08-18 08:09:27 DEBUG security.c:452 authenticate: BAD_PASSWORD
2025-08-18 08:09:27 DEBUG worker.c:293 client disconnect: 9
2025-08-18 08:09:27 DEBUG main.c:2226 pgagroal: Transfer client done
2025-08-18 08:09:27 DEBUG main.c:2236 pgagroal: Transfer client done: PID 90678
2025-08-18 08:09:27 DEBUG pool.c:1267 pgagroal_pool_status: 0/8
2025-08-18 08:09:27 DEBUG pool.c:1561 pgagroal_pool_status: State: FREE
2025-08-18 08:09:27 DEBUG pool.c:1562                       Slot: 0
2025-08-18 08:09:27 DEBUG pool.c:1563                       Server: 0
[...]

PR #627, opened 2025-08-18
[#625] Cannot acquire connections when pool is prefilled (false “pool full”)

Userfrom1995:


jesperpedersen:
Merged.

Issue #633, opened 2025-09-08
Core API test cases


Userfrom1995:

jesperpedersen:
This is just about the test infrastructure and any bug fixes that are found

Userfrom1995:
Addressed in: https://github.com/agroal/pgagroal/issues/636

PR #634, opened 2025-09-08
[#633] Core API test cases

Userfrom1995:


jesperpedersen:
We don't need the core API tests here only the fixes; the core API live in pgmoneta

Userfrom1995:
> We don't need the core API tests here only the fixes; the core API live in pgmoneta

Understood. I’ll remove the core API test part and keep only the fixes. I’ll also update the PR title and issue to reflect “core API bug fixes".

Userfrom1995:
@jesperpedersen I’m closing this one, along with the related issue, since @Jubilee101 has already addressed it in this commit: https://github.com/agroal/pgagroal/commit/1b4e42a38590c08c778d55ce020dcca80d7253d0

Issue #642, opened 2025-10-02
vault: /status endpoint always returns 500


Userfrom1995:
```c
base@LAPTOP-409T8FMM:~/pgagroal/build/src$ ./pgagroal-vault
2025-10-02 21:39:04 DEBUG configuration.c:1102 pgagroal-vault: Selected backend 'epoll'
2025-10-02 21:39:04 INFO  vault.c:928 pgagroal-vault 2.0.0: Started on localhost:2500
2025-10-02 21:39:04 DEBUG vault.c:931 Socket: 3
2025-10-02 21:39:04 DEBUG vault.c:935 Metrics: 5
2025-10-02 21:39:11 TRACE vault.c:1018 accept_vault_cb: client address: ::
2025-10-02 21:39:11 TRACE network.c:428 pgagroal_disconnect: fd=6
2025-10-02 21:39:11 TRACE network.c:428 pgagroal_disconnect: fd=3
2025-10-02 21:39:11 TRACE network.c:428 pgagroal_disconnect: fd=5
2025-10-02 21:39:11 DEBUG security.c:473 SSL request from client: 6
2025-10-02 21:39:11 DEBUG vault.c:162 router: Processing status endpoint request
2025-10-02 21:39:11 DEBUG vault.c:514 test_pgagroal_connectivity: Testing connection to localhost:2347
2025-10-02 21:39:11 TRACE network.c:428 pgagroal_disconnect: fd=3
2025-10-02 21:39:11 DEBUG network.c:349 pgagroal_connect: Connection refused
2025-10-02 21:39:11 TRACE network.c:428 pgagroal_disconnect: fd=-1
2025-10-02 21:39:11 DEBUG vault.c:518 test_pgagroal_connectivity: Connection failed
2025-10-02 21:39:11 DEBUG vault.c:408 Status request completed
2025-10-02 21:39:11 WARN  vault.c:412 Status endpoint: Failed to generate response
2025-10-02 21:39:11 DEBUG vault.c:166 router: Status response generation failed, sending 500 error

=================================================================
==81620==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 428 byte(s) in 1 object(s) allocated from:
    #0 0x55f1ba965a20 in realloc (/home/base/pgagroal/build/src/pgagroal-vault+0xcea20) (BuildId: c4d0784d47af05bbc683b7008fd5ece585b727f6)
[...]

PR #643, opened 2025-10-02
[#642] vault: fix route_status fallthrough causing 500 response and memory leak

Userfrom1995:


jesperpedersen:
Merged.

Issue #644, opened 2025-10-05
RPM build fails with CMake error: 'add_subdirectory given source "test" which is not an existing directory' when building from CPack-generated source tarball


Userfrom1995:
```c
-- Performing Test LINKER_HAS_RELRO
-- Performing Test LINKER_HAS_RELRO - Success
-- Performing Test LINKER_HAS_NOW
-- Performing Test LINKER_HAS_NOW - Success
CMake Error at CMakeLists.txt:192 (add_subdirectory):
  add_subdirectory given source "test" which is not an existing directory.


-- Configuring incomplete, errors occurred!
error: Bad exit status from /var/tmp/rpm-tmp.IA4ocp (%build)

RPM build warnings:
    %source_date_epoch_from_changelog is set, but %changelog has no entries to take a date from

RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.IA4ocp (%build)
```

jesperpedersen:
Likely similar to https://github.com/pgexporter/pgexporter/commit/2e7ba41926f7f5c0a0b15198b373cccbd945e8db

Userfrom1995:
> Likely similar to [pgexporter/pgexporter@2e7ba41](https://github.com/pgexporter/pgexporter/commit/2e7ba41926f7f5c0a0b15198b373cccbd945e8db)

Yes, I saw that one and tried building it here as well , I encountered the same issue here.
https://github.com/pgexporter/pgexporter/issues/335 similar to this one.

PR #645, opened 2025-10-05
[#644] cmake: Make test subdirectory conditional to fix RPM build from source tarball (#644)

Userfrom1995:


jesperpedersen:
Merged.

jesperpedersen:
@Userfrom1995 Please, port if necessary

Issue #646, opened 2025-10-15
Double-free Error When TLS is Enabled but Certificate/Key Files Are Missing


Userfrom1995:
## Description
When `tls = on` is configured in `pgagroal_vault.conf` but either `tls_cert_file` or `tls_key_file` is not provided (or points to non-existent files), the vault crashes with a **double-free memory error** during startup.

## Steps to Reproduce
1. Set `tls = on` in `pgagroal_vault.conf`
2. Either omit `tls_cert_file` and `tls_key_file`, or point them to non-existent paths
3. Start `pgagroal-vault`
4. Observe segmentation fault or double-free error

## Expected Behavior
The vault should detect the missing or invalid TLS configuration during startup and **exit gracefully** with a clear error message, similar to how the main `pgagroal` server validates TLS configuration.

## Proposed Solution
Add a `pgagroal_vault_tls_valid()` function that validates TLS configuration during startup, before initializing SSL contexts.  
This function should check:

- If `tls` is enabled, both `tls_cert_file` and `tls_key_file` must be configured  
- Certificate and key files must exist and be readable  
- Files must have proper ownership and permissions

```c
base@LAPTOP-409T8FMM:~/pgagroal/build/src$ ./pgagroal-vault
2025-10-15 13:11:36 DEBUG configuration.c:1102 pgagroal-vault: Selected backend 'epoll'
2025-10-15 13:11:36 INFO  vault.c:926 pgagroal-vault 2.0.0: Started on localhost:2500
2025-10-15 13:11:36 DEBUG vault.c:932 Socket: 3
2025-10-15 13:11:36 DEBUG vault.c:936 Metrics: 5
2025-10-15 13:11:39 TRACE vault.c:1019 accept_vault_cb: client address: ::
2025-10-15 13:11:39 TRACE network.c:428 pgagroal_disconnect: fd=6
2025-10-15 13:11:39 TRACE network.c:428 pgagroal_disconnect: fd=3
2025-10-15 13:11:39 TRACE network.c:428 pgagroal_disconnect: fd=5
2025-10-15 13:11:39 DEBUG security.c:473 SSL request from client: 6
2025-10-15 13:11:39 ERROR security.c:4392 Couldn't load TLS private key: /home/base/tls/vault/vault_server1.key
2025-10-15 13:11:39 ERROR security.c:4393 Reason: (null)
=================================================================
==622821==ERROR: AddressSanitizer: attempting double-free on 0x50b0000005c0 in thread T0:
[...]

PR #648, opened 2025-10-15
[#646] Add vault TLS validation function

Userfrom1995:


src/vault.c
```diff
@@ -865,6 +865,13 @@ main(int argc, char** argv)
       }
    }
 
+   // -- Validate TLS configuration if enabled --
+   if (pgagroal_vault_tls_valid())
+   {
+      pgagroal_log_fatal("pgagroal-vault: Invalid TLS configuration");
```
jesperpedersen: We don't need the double logging - so you can remove this

jesperpedersen:
Merged.

Issue #650, opened 2025-10-16
Combine documentation into one


jesperpedersen:
Take all documentation and tutorials to combine them into one such there are no duplicates of information.

Create the same setup as the sister projects like pgmoneta

jesperpedersen:
@fluca1978 @Userfrom1995 Please, help to look this over

We need a table of content that will high-light to users where they should look

Userfrom1995:
> [@fluca1978](https://github.com/fluca1978) [@Userfrom1995](https://github.com/Userfrom1995) Please, help to look this over
> 
> We need a table of content that will high-light to users where they should look

We could add a table listing each file in the manual along with a short description and a link to it. This could go in the [GETTING_STARTED.md](https://github.com/agroal/pgagroal/blob/master/doc/GETTING_STARTED.md), or alternatively in the [introduction.md](https://github.com/agroal/pgagroal/blob/master/doc/manual/en/01-introduction.md) within the manual.

If this approach sounds good, I can work on it — or we could even create a separate `index.md` under the `doc` folder to serve as the main table of contents.

jesperpedersen:
@Userfrom1995 It sounds like something for the developers chapter - try and make a patch for it

fluca1978:
I would thumb-up the idea for listing all the manual files, also because most of them are already numbered.
I would also extract the top level headings from each page, this should be quite trivial since it's markdown. In this way we have an argument (the file name) and subsections with the links.

@Userfrom1995 let me know if you need help automating this
@jesperpedersen it is not clear to me what you mean we should look at in `pgmoneta` for this task

PR #652, opened 2025-10-20
[#650] Add comprehensive table of contents and integrate missing manual chapters

Userfrom1995:


Userfrom1995:
@jesperpedersen @fluca1978
This is the work I’ve done so far on the table and the manual. I’m a bit unsure about the next steps — should we automate the table creation as @fluca1978 suggested, or keep the tables in separate files? Let’s discuss and finalize the approach here. Once everyone is satisfied, we can close both the issue and this PR together.

fluca1978:
This looks good for me, but I'm unsure if it is what @jesperpedersen had in mind.
Quite frankly, I don't like the idea of having two different links and let the user to choose for the right one, but it could be something to automate later on.

jesperpedersen:
I think this is ok - we just need to remove all references to the tutorial files since I removed them.

I forgot to update the .spec file so do that as well, please

jesperpedersen:
Merged.

Issue #659, opened 2025-12-06
Add pre-flight validation for configuration file to prevent parsing errors


saraanbih:
Problem
-------
The daemon attempts to parse the configuration file directly and may
produce parsing errors that are hard to interpret when the underlying problem
is a missing file, a directory path, insufficient permissions, or a binary file.

Proposed solution
-----------------
Add a pre-flight validation step that:
- checks the file exists,
- checks it is a regular file,
- checks it is readable,
- checks the file is not binary (contains no NUL bytes).

Benefits
--------
- Clear and immediate errors for common configuration mistakes.
- Better `systemd` STATUS messaging for faster debugging.
- Avoids unnecessary parse attempts.

PR #660, opened 2025-12-06
[#659] Add pre-flight validation of the configuration file before attempting

saraanbih:
## Problem

The daemon may fail while parsing the main configuration file and emit ambiguous parsing errors when the root cause is a filesystem issue (missing file, a directory path, insufficient permissions, or a binary file).
These cases produce unclear operator feedback (including poor systemd STATUS messages) and make diagnosis harder.
Proposed Solution

## Solution

- checks the path exists,
- verifies it is a regular file,
- ensures it is readable,
- ensures the file is not binary (no NUL bytes).

Fixes [#659]

### DIFF
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -46,6 +46,8 @@
 #include <openssl/pem.h>
 #include <sys/types.h>
 #include <sys/utsname.h>
+#include <sys/stat.h>
+#include <unistd.h>
 #ifdef HAVE_EXECINFO_H
 #include <execinfo.h>
 #endif


saraanbih:
Hi Mr. @jesperpedersen, I hope you’re doing well. I’ve made the changes you requested. Could you please review them?

jesperpedersen:
@saraanbih Can you look at the CI ?

src/libpgagroal/configuration.c
```diff
@@ -1195,6 +1196,40 @@ pgagroal_read_hba_configuration(void* shm, char* filename)
...
+   }
+
+   if (!pgagroal_exists(path))
+   {
+      return -ENOENT;
+   }
+
+   if (!pgagroal_is_file(path))
+   {
+      return -EISDIR;
```
fluca1978: it does not mean it is a directory, I would rather return another `ENOENT`.

src/include/configuration.h
```diff
@@ -256,6 +256,18 @@ pgagroal_vault_validate_configuration(void* shmem);
...
 pgagroal_read_hba_configuration(void* shmem, char* filename);
 
+/**
+ * Validate a configuration file for existence, type, readability and binary content
+ * @param path The file path
+ * @return 0 if the file is ok, otherwise:
+ *         1 = not exists
+ *         2 = not a regular file
+ *         3 = not readable
+ *         4 = contains binary data or read error
```
fluca1978: you are returning negative E flags, like `-ENOENT`, so the documentation is wrong.
Also, since `0` means ok, there is no need to return a negative error value.

src/main.c
```diff
@@ -556,6 +556,49 @@ main(int argc, char** argv)
 
    // the main configuration file is mandatory!
    configuration_path = configuration_path != NULL ? configuration_path : PGAGROAL_DEFAULT_CONF_FILE;
+
+   int cfg_ret = pgagroal_validate_config_file(configuration_path);
+
+   if (cfg_ret < 0)
```
fluca1978: `if ( cfg_ret )`

src/main.c
```diff
@@ -556,6 +556,49 @@ main(int argc, char** argv)
...
+            sd_notifyf(0, "STATUS=Configuration file not found: %s", configuration_path);
+#endif
+            errx(1, "Configuration file not found: %s", configuration_path);
+            break;
+
+         case EISDIR:
+#ifdef HAVE_SYSTEMD
+            sd_notifyf(0, "STATUS=Configuration path is not a file: %s", configuration_path);
+#endif
+            errx(1, "Configuration path is not a file: %s", configuration_path);
```
fluca1978: this case is useless to me, while we can have `ENOENT` for both not existing file or not a regular file, we can squash these two branches with a single one and an error message like `Configuration path ... is not a regular file`.

src/main.c
```diff
@@ -556,6 +556,49 @@ main(int argc, char** argv)
...
+            break;
+
+         case EINVAL:
+#ifdef HAVE_SYSTEMD
+            sd_notifyf(0, "STATUS=Configuration file contains binary data: %s", configuration_path);
+#endif
+            errx(1, "Configuration file contains binary data: %s", configuration_path);
+            break;
+
+         default:
```
fluca1978, line 587: Uhm, do we have a validation error without a catched error?

jesperpedersen:
@fluca1978 @saraanbih is coordinating this feature across the projects - you can see comments from pgexport [here](https://github.com/pgexporter/pgexporter/pull/360)

src/include/configuration.h
saraanbih: @fluca1978 I was returning what i write in the documentation (positive numbers) but in pgexporter they advise me to use error codes
> Usually when you return values like this, you use enums or macros containing these values (in C it's usually the latter). This way, the user of pgexporter_validate_config_file doesn't have to look into the implementation to guess what each error value means. It's also a good practice to use err codes already baked into standard C (and often the OS) instead of custom error codes. Take a look at man 3 errno (either in your terminal or [here](https://man7.org/linux/man-pages/man3/errno.3.html)). They're also often returned as negative values (like return -ENOENT;).

src/main.c
saraanbih: @fluca1978 oh yeah Great note! I didn't notice it!

fluca1978:
@saraanbih looks fine to me but CI is failing, there should be something wrong with the includes.

saraanbih:
Yeah Mr. @fluca1978, I have fixed it.

fluca1978:
thanks @saraanbih 
Just the next time, please force push a single commit per issue.

saraanbih:
Many Thanks Mr. @fluca1978 Oh i rebase it local and forget to push "(

Issue #663, opened 2025-12-22
Replace Uncrustify with ClangFormat for C code formatting


TutTrue:
**Is your feature request related to a problem? Please describe.**
Currently, our repository uses Uncrustify for C code formatting. This creates friction for developers using Language Server Protocol (LSP) tools since ClangFormat has better native integration with modern IDEs and code editors. LSP users cannot easily make formatting changes or see formatting issues in real-time, reducing productivity and code quality.

**Describe the solution you'd like**
Replace Uncrustify with ClangFormat as our C code formatter. This would involve:

1. Creating a .clang-format file that matches our current style guidelines
2. Updating documentation to reflect the change
3. Update ci

PR #664, opened 2025-12-22
[#663]Add clang-format configuration and script for code formatting

TutTrue:
This PR replaces Uncrustify with ClangFormat as the C code formatter used in this repository while keeping formatting behavior consistent with the existing style.


fluca1978:
I think this could be interesting, while it seems to me mainly a matter of taste.
Please note that, even if the reformatting appears a lot more consistent, there are places that are weird, according to me, like the space removal after a case (`struct art_node48* n = (struct art_node48*)node;`, no space before `node`) or before the parenthesis in a call to `assert` (i.e., `assert (` -> `assert(`). Therefore I believe there is more tuning to do to make the impact of clang formatter less evident if applied.
However, it sounds a good idea to me to have a more integrable tool for code reformatting.
@jesperpedersen thoughts?

jesperpedersen:
@fluca1978 I think we can start with this - we will most of the way there with this PR. I'm ok with the removed space...

See the similar work in pgmoneta

fluca1978:
> @fluca1978 I think we can start with this - we will most of the way there with this PR. I'm ok with the removed space...
> 
> See the similar work in pgmoneta

I've done a little testing, seems ok to me to merge. 
To be absolutely clear, do I have to merge this?

jesperpedersen:
Rebased, and merged.

Issue #676, opened 2026-01-15
./check.sh causes Segmentation fault


ankushT369:
ran ```./check.sh``` in test/ dir

```shell
  Users:
    - myuser
    - ankush
Add user ankush to pgagroal_users.conf file ... ok
./check.sh: line 323: 79692 Segmentation fault      (core dumped) $EXECUTABLE_DIRECTORY/pgagroal-admin -f $CONFIGURATION_DIRECTORY/pgagroal_users.conf -U "$PG_UTF8_USER_NAME" -P "$PG_UTF8_USER_PASSWORD" user add
Clean up
Shutdown pgagroal
pgagroal cleanup completed
Clean Test Resources
```

[log.txt](https://github.com/user-attachments/files/24647728/log.txt)

jesperpedersen:
@ankushT369 Van you investigate a bit further ?

ankushT369:
@jesperpedersen i just found it. I am investigating it further

fluca1978:
Could it be related to https://github.com/pgagroal/pgagroal/commit/3807fb0ac60a844e26a27ac2ebf29845925610e9 ?
I'll try to investigate too as soon as possible.

fluca1978:
Apprently I cannot reproduce it.
@ankushT369  are you running on HEAD? This  is the output on https://github.com/pgagroal/pgagroal/commit/3807fb0ac60a844e26a27ac2ebf29845925610e9

```
...
Master Key stored into /home/luca/.pgagroal/master.key
Header:
  ClientVersion: 2.0.0
  Command: 20
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20260116120530
Outcome:
  Status: true
  Time: 00:00:00
Response:
  Users:
    - myuser
Add user myuser to pgagroal_users.conf file ... ok
Header:
  ClientVersion: 2.0.0
  Command: 20
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20260116120530
Outcome:
  Status: true
  Time: 00:00:00
Response:
  Users:
    - myuser
    - luca
Add user luca to pgagroal_users.conf file ... ok
Header:
  ClientVersion: 2.0.0
  Command: 20
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20260116120530
Outcome:
  Status: true
  Time: 00:00:00
Response:
  Users:
    - myuser
    - luca
    - utf8user
[...]

PR #677, opened 2026-01-18
[#676]update check.sh to support gcc compilation without coverage

ankushT369:


test/check.sh
```diff
@@ -170,46 +170,48 @@ cleanup() {
...
-         --format=text > $COVERAGE_DIR/coverage-pgagroal.txt
-      echo "Generating $COVERAGE_DIR/coverage-pgagroal-cli.txt"
-      llvm-cov show $EXECUTABLE_DIRECTORY/pgagroal-cli \
-         --instr-profile=$COVERAGE_DIR/coverage.profdata \
-         --format=text > $COVERAGE_DIR/coverage-pgagroal-cli.txt
-      echo "Generating $COVERAGE_DIR/coverage-pgagroal-admin.txt"
-      llvm-cov show $EXECUTABLE_DIRECTORY/pgagroal-admin \
-         --instr-profile=$COVERAGE_DIR/coverage.profdata \
-         --format=text > $COVERAGE_DIR/coverage-pgagroal-admin.txt
+      if [[    $MODE != "gcc" ]]; then
```
jesperpedersen: There are some formatting issues in this patch

ankushT369: Ok i will fix it

test/check.sh
```diff
@@ -514,17 +517,24 @@ run_tests() {
...
       build_postgresql_image
     fi
   fi
 
   echo "Preparing the pgagroal directory"
-  # Set up LLVM coverage
-  export LLVM_PROFILE_FILE="$COVERAGE_DIR/coverage-%m-%p.profraw"
+  # Set up LLVM coverage or GCC
+  if [[ $MODE != "gcc" ]]; then 
+    export LLVM_PROFILE_FILE="$COVERAGE_DIR/coverage-%m-%p.profraw"
```
fluca1978:
```
else 
   export LLVM_PROFILE_FILE=""
  export COVERAGE_DIR=""
fi
```

this can ease the management of the following piece.

test/check.sh
```diff
@@ -514,17 +517,24 @@ run_tests() {
...
+  fi
   sudo rm -Rf "$PGAGROAL_ROOT_DIR"
   mkdir -p "$PGAGROAL_ROOT_DIR"
-  mkdir -p "$LOG_DIR" "$PG_LOG_DIR" "$COVERAGE_DIR" "$BASE_DIR" "$RESOURCE_DIRECTORY"
```
fluca1978, line 527: assuming `COVERAGE_DIR` has been set or not, this line could remain intact and the directory will not be created if the variable is empty.

test/check.sh
```diff
@@ -514,17 +517,24 @@ run_tests() {
...
+  if [[ $MODE != "gcc" ]]; then
```
fluca1978: and this is if is not required anymore if `COVERAGE_DIR` is empty or not.

test/check.sh
```diff
@@ -540,12 +550,20 @@ run_tests() {
...
+     export CC=$(which clang)
+     echo "Using Clang compiler with LLVM coverage: $CC"
+     cmake -DCMAKE_C_COMPILER=clang \
+          -DCMAKE_BUILD_TYPE=Debug \
+         ..
+   else
+     export CC=$(which gcc)
+     echo "Using GCC compiler without coverage: $CC"
+     cmake -DCMAKE_C_COMPILER=gcc \
+          -DCMAKE_BUILD_TYPE=Debug \
```
fluca1978: I don't see the need for having the same code replicated in two branches.
The only difference I spot is the `export CC`, so trim the `if` down to that and proceed:

```
if [[ $MODE != gcc ]]; then
   export CC=$(which clang)
else
  export CC=$(which gcc)
fi

cmake -DCMAKE_C_COMPILER=$CC \
          -DCMAKE_BUILD_TYPE=Debug \
         ..
```

test/check.sh
```diff
@@ -170,46 +170,48 @@ cleanup() {
       fi
       
       # Generate LLVM coverage reports
-      if ls "$COVERAGE_DIR"/*.profraw >/dev/null 2>&1; then
-       echo "Generating coverage report, expect error when the binary is not covered at all"
```
fluca1978, line 174: doesn't suffice to add a branch in here:
```
if [ $MODE != gcc ]]; then
  ... same code
fi
```

this is going to trim down the patch.

test/check.sh
```diff
@@ -170,46 +170,48 @@ cleanup() {
...
-         --format=text > $COVERAGE_DIR/coverage-pgagroal.txt
-      echo "Generating $COVERAGE_DIR/coverage-pgagroal-cli.txt"
-      llvm-cov show $EXECUTABLE_DIRECTORY/pgagroal-cli \
-         --instr-profile=$COVERAGE_DIR/coverage.profdata \
-         --format=text > $COVERAGE_DIR/coverage-pgagroal-cli.txt
-      echo "Generating $COVERAGE_DIR/coverage-pgagroal-admin.txt"
-      llvm-cov show $EXECUTABLE_DIRECTORY/pgagroal-admin \
-         --instr-profile=$COVERAGE_DIR/coverage.profdata \
-         --format=text > $COVERAGE_DIR/coverage-pgagroal-admin.txt
+      if [[ $COVERAGE == "YES" ]]; then
```
jesperpedersen, line 173: There is an indentation of 1 in this group - make it 2

ankushT369:
@jesperpedersen is this ok now ?

test/check.sh
```diff
@@ -170,46 +170,48 @@ cleanup() {
...
+        if ls "$COVERAGE_DIR"/*.profraw >/dev/null 2>&1; then
```
jesperpedersen: Looks like there is still 1 indentation here

ankushT369: if will be like this?
```shell
      if [[ $COVERAGE == "YES" ]]; then
         if ls "$COVERAGE_DIR"/*.profraw >/dev/null 2>&1; then
```

fluca1978: seemed fine to me, there was 1 indentation even before (the commands where aligned under the `f` of `if`)

fluca1978:
It is working, so it is fine for me. However I don't understand the failure in the FreeBSD builds. @ankushT369 any idea?
@jesperpedersen is this ok on your side?

jesperpedersen:
@fluca1978 Feel free to merge

ankushT369:
> It is working, so it is fine for me. However I don't understand the failure in the FreeBSD builds. @ankushT369 any idea? @jesperpedersen is this ok on your side?

I am checking this out I think its kind of rsync related issue!! just guessing

fluca1978:
I would like to merge once we fixed the CI part.

ankushT369:
@fluca1978 can u check this ?

ankushT369:
@jesperpedersen any updates??

jesperpedersen:
Rebased, and merged.

Issue #678, opened 2026-01-20
ev: Fix io_uring CQE reuse race condition


Userfrom1995:
### Description

In `pgagroal_event_prep_submit_send()`, `cqe->res` was being read after calling `io_uring_cqe_seen()`. Once `io_uring_cqe_seen()` is called, the completion queue entry (CQE) can be reused by the kernel, which could potentially overwrite `cqe->res` before we read it.

This is the same bug that was recently fixed in PostgreSQL's io_uring implementation:

https://github.com/postgres/postgres/commit/8077649907d40b9831358d0c5121a8c58267a988

https://www.postgresql.org/message-id/bwo3e5lj2dgi2wzq4yvbyzu7nmwueczvvzioqsqo6azu6lm5oy@pbx75g2ach3p

### Impact

- The race window is very narrow
- Likelihood of occurrence is low (requires high io_uring CQE utilization)
- Consequences would be severe: corrupted result values leading to incorrect error handling or data corruption

### Fix

Save `cqe->res` into a local variable before calling `io_uring_cqe_seen()`:

```c
cqe_res = cqe->res;
io_uring_cqe_seen(&loop->ring_snd, cqe);
// Now use cqe_res instead of cqe->res

PR #679, opened 2026-01-20
[#678] ev: Fix io_uring CQE reuse race condition 

Userfrom1995:
Read cqe->res into a local variable before calling io_uring_cqe_seen() to prevent the completion queue entry from being reused before we read the result. This mirrors an upstream PostgreSQL fix for the same issue in their io_uring implementation.

The window for this race is very narrow and the likelihood of it happening is low, but the consequences (corrupted result value) would be severe.

Also removed verbose per-recv trace logging that flooded log files at TRACE level.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,7 @@ jobs:
             bridge-utils
           sudo systemctl start libvirtd
       - name: Setup FreeBSD VM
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

fluca1978:
Sorry, I didn't get this in time for a review.
Great job @Userfrom1995 !

PR #690, opened 2026-01-24
[#689] io_uring: Improve event handler stability and cleanup

Userfrom1995:
Refine the io_uring event handler to better handle edge cases:
- Handle accept errors and unexpected multishot termination by rearming the watcher.
- Prevent invalid file descriptors from reaching the accept callback.
- Simplify SSL cleanup in worker by using the standard helper function.
- Add safety check for Prometheus slot reset.

This ensures the event loop doesn't stop on transient errors and standardizes cleanup logic.


jesperpedersen:
Merged.

PR #691, opened 2026-01-25
[#687] Implement Health Check

### LINKED ISSUE #687: Implement a PostgreSQL health check
Implement a way to health-check, periodically, a PostgreSQL cluster status.
This could mean:
- add a `status` flag into the `server` struct, to store the last reported status (OK, KO); it could require a timestamp too;
- report the last known cluster status in `pgagroal-cli` commands, mainly `status` and `status detail`;
- modify periodically, using a timer, the status for each server. This could require to send a custom query (e.g., `SELECT 1`) or get info from a currently in-use connection to such database server. The former case could require an extra connection to the cluster;
- implement a configuration tunable to turn this feature on or off (could it be worth to configure this on a per-PostgreSQL server, so instead of the `pgagroal` configuration section, the *server* one).

Moreover, the failover workflow should automatically exclude any PostgreSQL server that has been reported as failing lastly.
Related to #686

trxvorr:
this PR introduces the core health check functionality to pgagroal.

by implementing support for Trust, MD5, and SCRAM-SHA-256 authentication out of the box, the health checker is production ready for various security environments. additionally refactoring the client side authentication handshake into shared, public functions ensures that the health checker uses the same verified logic as the standard connection pool, maintaining architectural consistency and reducing code duplication.

the introduction of the `health_check_user` configuration further supports the principle of least privilege by allowing dedicated, low privilege monitoring users.

### Verification
i have validated this implementation against a live PostgreSQL 17 instance across 8 distinct scenarios, verifying both success and failure states to ensure pooler stability.

| Scenario | Auth Method | Expected Status | Validation Point |
| :--- | :--- | :--- | :--- |
| Success | Trust | UP | Connects without credentials |
| Failure | Trust (Reject) | DOWN | pg_hba.conf reject is handled correctly |
| Success | MD5 | UP | Validated legacy hash handshake |
| Failure | MD5 (Wrong Pass) | DOWN | Correct failure on bad credentials |
| Success | SCRAM | UP | Validated modern proof handshake |
| Failure | SCRAM (Wrong Pass) | DOWN | Correct failure on bad credentials |
| Explicit User | SCRAM/MD5 | UP | health_check_user setting works |
| Logic Check | - | DOWN | Graceful exit if user missing in pgagroal_users.conf |

fixes #687


src/libpgagroal/shmem.c
```diff
@@ -85,6 +85,13 @@ pgagroal_create_shared_memory(size_t size, unsigned char hp, void** shmem)
 
    *shmem = s;
 
+   struct main_configuration* config = (struct main_configuration*)s;
```
jesperpedersen: This need to be in configuration.c - init function

src/libpgagroal/health.c
```diff
@@ -0,0 +1,364 @@
...
+{
+   struct main_configuration* config = (struct main_configuration*)shmem;
+   struct server* server = &config->servers[server_idx];
+   int fd = -1;
+   char buffer[1024];
+   int offset = 0;
+
+   /* * NOTE: Current Implementation Limitations
+    * 1. Auth: Currently supports TRUST authentication only.
+    * Future work: Support MD5/SCRAM using pgagroal_users.conf credentials.
```
jesperpedersen: Yes, we need this

src/libpgagroal/health.c
```diff
@@ -0,0 +1,364 @@
...
+    * 2. User: Hardcoded to "postgres".
+    * Future work: Make the health check user configurable.
+    */
+
+   /* ---------------------------------------------------------
```
jesperpedersen: Remove

src/libpgagroal/health.c
```diff
@@ -0,0 +1,364 @@
...
+    * STEP 2: Construct Startup Packet (Protocol v3.0)
+    * --------------------------------------------------------- */
+   /* Format: [Length (4)] [Version (4)] [user\0name\0] [database\0name\0] [\0] */
+
+   memset(buffer, 0, sizeof(buffer));
+   offset = 8; // Skip Length (4) and Version (4) for now
+
+   /* Parameter: user */
+   sprintf(buffer + offset, "user");
```
jesperpedersen: pgagroal_snprintf or pgagroal_append

jesperpedersen:
@trxvorr You need to run with clang 21+, or let CI do it

trxvorr:
> @trxvorr You need to run with clang 21+, or let CI do it

yess i was planning on doing it after i get a reply about the auth part

also how do i make CI do it?

jesperpedersen:
@trxvorr @fluca1978 and I will trigger CI - after your 1st contribution it will be automatic. As to auth, we need SCRAM-SHA-256 and MD5 for the health connection

trxvorr:
> As to auth, we need SCRAM-SHA-256 and MD5 for the health connection

okay ill work on it right now thank you

jesperpedersen:
@trxvorr You should have all of it in security.c - so take some time to learn the APIs... You likely need https://github.com/pgagroal/pgagroal/blob/master/doc/CONFIGURATION.md#pgagroal_superuser-configuration for this...

fluca1978:
@trxvorr I've left a few comments in the code.

Having a second thought to this, we should include also arguments to specify at least the connection user (password could be caught by `pgagroal_users` and friends).
Now, I was thinking if the process should do the scheduling by itself or if we should use `ev.h` for this. So far, I believe having a separated process with its own scheduler is much simpler.

src/include/health.h
```diff
@@ -0,0 +1,49 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <pgagroal.h>
+
+#include <stdlib.h>
+
+/**
+ * Start the health check process
```
jesperpedersen, line 41: Documentation

src/include/health.h
```diff
@@ -0,0 +1,49 @@
...
+ */
+void pgagroal_start_health_check(int argc, char** argv);
```
jesperpedersen: Formatting

src/include/pgagroal.h
```diff
@@ -358,6 +365,8 @@ struct server
    char tls_key_file[MAX_PATH];  /**< TLS key path */
    char tls_ca_file[MAX_PATH];   /**< TLS CA certificate path */
    atomic_schar state;           /**< The state of the server */
+   atomic_schar health_state;    /**< The health state of the server */
+   int failures;                 /**< The number of failures */
```
jesperpedersen: unsigned int

src/include/pgagroal.h
```diff
@@ -673,6 +682,10 @@ struct main_configuration
    int validation;                                /**< Validation mode */
    unsigned int background_interval;              /**< Background validation timer in seconds */
    int max_retries;                               /**< The maximum number of retries */
+   bool health_check;                             /**< Is health check enabled */
+   int health_check_period;                       /**< Health check period in seconds */
+   unsigned int health_check_timeout;             /**< Health check timeout */
+   char health_check_query[MISC_LENGTH];          /**< The health check query */
```
jesperpedersen: Do we need this ?

src/libpgagroal/configuration.c
```diff
@@ -158,6 +160,10 @@ pgagroal_init_configuration(void* shm)
    config->validation = VALIDATION_OFF;
    config->background_interval = DEFAULT_BACKGROUND_INTERVAL;
    config->max_retries = 5;
+   config->health_check = false;
+   config->health_check_period = 30;
+   config->health_check_timeout = 5;
+   snprintf(config->health_check_query, MISC_LENGTH, "SELECT 1");
```
jesperpedersen: I think we can remove this for now

src/libpgagroal/configuration.c
```diff
@@ -5535,6 +5541,42 @@ pgagroal_apply_main_configuration(struct main_configuration* config,
...
+      }
+   }
+   else if (key_in_section("health_check_period", section, key, true, &unknown))
+   {
+      if (as_int(value, &config->health_check_period))
+      {
+         unknown = true;
+      }
+
+      if (config->health_check_period < 1)
```
jesperpedersen: This belongs in `validate_main_configuration`

src/libpgagroal/configuration.c
```diff
@@ -5535,6 +5541,42 @@ pgagroal_apply_main_configuration(struct main_configuration* config,
...
+      }
+   }
+   else if (key_in_section("health_check_timeout", section, key, true, &unknown))
+   {
+      if (as_seconds(value, &config->health_check_timeout, 5))
+      {
+         unknown = true;
+      }
+   }
+   else if (key_in_section("health_check_query", section, key, true, &unknown))
```
jesperpedersen: Remove

src/libpgagroal/health.c
```diff
@@ -0,0 +1,359 @@
...
+#include <stdlib.h>
+#include <unistd.h>
+#include <signal.h>
+#include <stdatomic.h>
+#include <arpa/inet.h>
+#include <string.h>
+#include <errno.h>
+
+static void health_check_loop(void);
+static void write_int32(char* buf, int value);
```
jesperpedersen: Use `utils.h`

src/libpgagroal/health.c
```diff
@@ -0,0 +1,359 @@
...
+static bool pgagroal_server_probe(int server_idx);
```
jesperpedersen: static functions are not prefix'ed by `pgagroal_`

src/libpgagroal/health.c
```diff
@@ -0,0 +1,359 @@
...
+      return;
+   }
+   else if (pid == 0)
+   {
+      /* Child process */
+      pgagroal_set_proc_title(argc, argv, "health check", NULL);
+      health_check_loop();
+   }
+
+   /* Parent process returns */
```
jesperpedersen: Remove

jesperpedersen:
@trxvorr My first 10 comments... maybe there are similar ones later in the patch

src/include/pgagroal.h
fluca1978: Are you suggesting to hardcode the query? We were thinking about having a way to inject a custom query to check availability.

src/libpgagroal/configuration.c
fluca1978: Ok, so we don't want this tunable at this stage, I had suggested this.

src/include/pgagroal.h
jesperpedersen: Yes, for now. We have to think about SQL injection attacks if we allow custom queries

fluca1978:
> @trxvorr My first 10 comments... maybe there are similar ones later in the patch

I had suggested we put the `health_check_query` in place at this stage, but we can introduce it later.

trxvorr:
@jesperpedersen @fluca1978 i have pushed a new commit that addresses the code review feedback and fixes several stability issues.

the issues i faced and fixed:
fixed nanosleep overflow by switching to standard sleep().
corrected the startup packet construction and suthtype offset parsing.
and increased stack buffer size to prevent overflows.

ive thoroughly tested it, now that the loop is stable i will implement the scram and md5 support

src/libpgagroal/health.c
```diff
@@ -44,13 +45,16 @@
...
+static bool server_probe(int server_idx);
 
 void
-pgagroal_start_health_check(int argc, char** argv)
+pgagroal_start_health_check(int slot)
 {
+   (void)slot;
+
    pid_t pid;
+   char title[512] = "pgagroal: health-check";
```
fluca1978: 512?

src/libpgagroal/health.c
```diff
@@ -44,13 +45,16 @@
...
 static void health_check_loop(void);
-static void write_int32(char* buf, int value);
-static bool pgagroal_server_probe(int server_idx);
+static bool server_probe(int server_idx);
 
 void
-pgagroal_start_health_check(int argc, char** argv)
+pgagroal_start_health_check(int slot)
 {
+   (void)slot;
```
fluca1978: What is the aim of `slot`? Seems unused to me.

trxvorr:
@jesperpedersen @fluca1978  implemented SCRAM-SHA-256 authentication for the backend health check. i have verified that it successfully authenticates with the server and correctly handles authentication failures (negative tests passed).

i am currently refactoring the user selection logic to respect the superuser and health_check_user configurations (with fallback) and cleaning up the handshake logic within health.c. once the refactoring is complete i will begin working on MD5 support to ensure full coverage.

had a few questions:

currently the SCRAM network handshake lives inside health.c while I use security.c only for the cryptographic primitives. would you prefer I move the entire handshake logic into a reusable function in security.c or is keeping the network flow in health.c acceptable?

jesperpedersen:
@trxvorr I havn't looked, but you should look at `pgagroal_remote_management_auth` and `pgagroal_remote_management_scram_sha256` to see how much you can reuse / split out

fluca1978:
@trxvorr please see my comments on [86221a3](https://github.com/pgagroal/pgagroal/commit/86221a32dce75576ad6dddd84f2af3735fa9a52b)

fluca1978:
Commit https://github.com/pgagroal/pgagroal/commit/f10da4d066c0aa09e77eef74d84e0b4482c8e593 removes the part that is not properly of this patch, open a new issue and solve the problem in there.
In the meantime, please take a look also at other comments.

trxvorr:
@fluca1978 i have updated the PR addressing all your feedback 
ready for another look when you have a moment 
thanks

trxvorr:
@fluca1978 @jesperpedersen before pushing my changes, i wanted to flag a small fix i found in utils.c while testing the health check worker.

i noticed that if the parent process sets a short title (pgagroal: main, length: 14) it updates the global static variable max_process_title_size to match that current length (14) instead of preserving the total buffer capacity

in STRICT mode, this causes the workers title to be truncated to pgagroal: healt

in default mode, the worker writes the full title (pgagroal: health check worker), writing past what the process thinks is its safe limit.

the fix would be removing the line max_process_title_size = size; in pgagroal_set_proc_title. This ensures the global variable always tracks the capacity of the argv buffer, allowing child processes to inherit the full space and display their titles correctly without truncation or potential memory safety issues

is this intended behavior?

fluca1978:
> @fluca1978 @jesperpedersen before pushing my changes, i wanted to flag a small fix i found in utils.c while testing the health check worker.
> 
> i noticed that if the parent process sets a short title (pgagroal: main, length: 14) it updates the global static variable max_process_title_size to match that current length (14) instead of preserving the total buffer capacity
> 
> in STRICT mode, this causes the workers title to be truncated to pgagroal: healt
> 
> in default mode, the worker writes the full title (pgagroal: health check worker), writing past what the process thinks is its safe limit.
> 
> the fix would be removing the line max_process_title_size = size; in pgagroal_set_proc_title. This ensures the global variable always tracks the capacity of the argv buffer, allowing child processes to inherit the full space and display their titles correctly without truncation or potential memory safety issues
> 
> is this intended behavior?

I don't this is wanted, so feel free to open an issue and work on it.
Thanks.

trxvorr:
@fluca1978 @jesperpedersen i noticed that the macOS and FreeBSD CI tests were failing during compilation because of the Linux specific sys/prctl.h header.

to address this, ive wrapped the prctl(PR_SET_PDEATHSIG, SIGTERM) logic in #ifdef __linux__. this ensures the build passes on non-Linux platforms while maintaining robust orphan process protection on Linux (ensuring the health check worker terminates if the main process crashes).

on macOS and FreeBSD, the worker currently relies on the standard pgagroal_health_check_stop cleanup during reloads/shutdown, but it won't automatically die if the parent is SIGKILL'd.

ill go ahead and force push this #ifdef fix for now to unbreak the CI also, im open to suggestions if we want to implement similar protection for macOS (via kqueues) or FreeBSD (via procctl) in a follow-up.

fluca1978:
> @fluca1978 @jesperpedersen i noticed that the macOS and FreeBSD CI tests were failing during compilation because of the Linux specific sys/prctl.h header.
> 
> to address this, ive wrapped the prctl(PR_SET_PDEATHSIG, SIGTERM) logic in #ifdef **linux**. this ensures the build passes on non-Linux platforms while maintaining robust orphan process protection on Linux (ensuring the health check worker terminates if the main process crashes).
> 
> on macOS and FreeBSD, the worker currently relies on the standard pgagroal_health_check_stop cleanup during reloads/shutdown, but it won't automatically die if the parent is SIGKILL'd.
> 
> ill go ahead and force push this #ifdef fix for now to unbreak the CI also, im open to suggestions if we want to implement similar protection for macOS (via kqueues) or FreeBSD (via procctl) in a follow-up.

Uhm, while `prctl` or `procctl` can be used in a very similar way, OpenBSD does not support neither. This makes the approach even less portable, so I don't know it is worth the effort.
The kqueue approach, which in turn is portable, is probably not worth because of race conditions: are you thinking about filtering for an exit of the parent?


@jesperpedersen thoughts?

trxvorr:
@fluca1978 i completely agree on the portability concern so ive removed the prctl/procctl logic to keep the implementation compatible with OpenBSD and other systems.

instead implemented a robust check in pgagroal_health_check_stop that waits up to 5 seconds for the worker process to exit after a SIGTERM this ensures clean lifecycle management during standard reloads/shutdowns without relying on platform specific orphan protection

fluca1978:
@trxvorr there are some issues with process title and, most notably, zombies.
Moreover, I may be lost, but where is the output of the health check? Somehow somewhere there must be an evidence of the server health check.

fluca1978:
@trxvorr I can confirm I launch the daemon with `pgagroal`, since it is in my `PATH`, so I guess we have another issue with the process title here, in the case we need a separate issue, I see #725.
Also, please do not merge branches, rebase.

trxvorr:
@fluca1978 regarding the visibilty, right now it only uupdates the internal state and prints to DEBUG logs

we could add a CLI integration which includes UP/DOWN for each server in the JSON management response, a pgagroal_server_health metric to prometheus.c and update the health loop to log at WARN when a server goes down, and INFO when it recovers

how does that sound

fluca1978:
> @fluca1978 regarding the visibilty, right now it only uupdates the internal state and prints to DEBUG logs
> 
> we could add a CLI integration which includes UP/DOWN for each server in the JSON management response, a pgagroal_server_health metric to prometheus.c and update the health loop to log at WARN when a server goes down, and INFO when it recovers
> 
> how does that sound

Sounds great to me, it would be very useful.

Please note that you are committing mixed things, for example https://github.com/pgagroal/pgagroal/commit/4f7922fa054fb969c32c70730140ab03f4135f71 has traces of a `restart_time` that you are probabling doing in a separate PR #732 .

fluca1978:
See my comments and the CI.

Seems also we have still some issues with shutting down the process:

```

% ps -auxw | grep pgagroal
luca        4765  0.6  1.0  12200  8216 pts/0    S+   14:25   0:00 pgagroal: main
luca        4766  0.0  0.2  12200  2212 pts/0    S+   14:25   0:00 pgagroal: health check worker
luca        4804  0.0  0.2   6408  2304 pts/1    S+   14:25   0:00 grep --color=auto pgagroal

% pgagroal-cli shutdown 
Header:
  ClientVersion: 2.1.0
  Command: shutdown-gracefully
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260223142534
Outcome:
  Status: true
  Time: 00:00:00
Request: {}

% ps -auxw | grep pgagroal
luca        4766  0.0  0.3  12200  2596 pts/0    S    14:25   0:00 pgagroal: health check worker
luca        4809  0.0  0.2   6408  2176 pts/1    S+   14:25   0:00 grep --color=auto pgagroal
```

trxvorr:
@fluca1978 the manual health_check_period.s was just a temp workaround until we finalized the approach in #732. I will remove those manual checks from this branch to keep itclean. once #732 is merged, i will rebase this branch and swap in the proper restart_time() helper

trxvorr:
```
47436 pgagroal: main
47437 pgagroal: health check worker
---Prometheus---
#HELP pgagroal_server_health The health state of the server (0 = DOWN, 1 = UP)
#TYPE pgagroal_server_health gauge
pgagroal_server_health{server="primary"} 1
---CLI JSON---
    "Servers": [
      {
        "Health": "UP",
        "Host": "127.0.0.1",
        "Port": 5432,
---Shutdown Test---
Header:
  ClientVersion: 2.1.0
  Command: shutdown-gracefully
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260301102407
Outcome:
  Status: true
  Time: 00:00:00
Request: {}
No pgagroal processes
---Logs---
2026-03-01 10:21:07 DEBUG health.c:161 Health check run
2026-03-01 10:21:07 DEBUG health.c:245 Health: Probing server 0 (user: postgres, database: postgres)
2026-03-01 10:21:07 DEBUG health.c:293 Health: Server 0 requested Trust auth
2026-03-01 10:21:09 DEBUG health.c:161 Health check run
2026-03-01 10:21:09 DEBUG health.c:245 Health: Probing server 0 (user: postgres, database: postgres)
2026-03-01 10:21:09 DEBUG health.c:293 Health: Server 0 requested Trust auth
2026-03-01 10:21:11 DEBUG health.c:161 Health check run
2026-03-01 10:21:11 DEBUG health.c:245 Health: Probing server 0 (user: postgres, database: postgres)
2026-03-01 10:21:11 DEBUG health.c:293 Health: Server 0 requested Trust auth
2026-03-01 10:24:02 INFO  health.c:144 Health check started
2026-03-01 10:24:04 DEBUG health.c:161 Health check run
2026-03-01 10:24:04 DEBUG health.c:245 Health: Probing server 0 (user: postgres, database: postgres)
2026-03-01 10:24:04 DEBUG health.c:293 Health: Server 0 requested Trust auth
2026-03-01 10:24:04 INFO  health.c:185 Health: Server 0 is UP
2026-03-01 10:24:06 DEBUG health.c:161 Health check run
2026-03-01 10:24:06 DEBUG health.c:245 Health: Probing server 0 (user: postgres, database: postgres)
2026-03-01 10:24:06 DEBUG health.c:293 Health: Server 0 requested Trust auth
2026-03-01 10:24:07 DEBUG main.c:2120 pgagroal: graceful shutdown triggered  - connections=0, gracefully=true, keep_running=true
2026-03-01 10:24:07 INFO  main.c:1371 pgagroal: shutdown
2026-03-01 10:24:07 INFO  health.c:206 Health check stopped
```

fluca1978:
Looks fine to me, I'm only in doubt if we should expose the `unknown` health to the prometheus metrics.
For the rest, it is fine, and passes the tests of reloading.

@jesperpedersen PTAL, it is fine for me to merge even without the small change in prometheus.

doc/manual/en/04-configuration.md
```diff
@@ -72,6 +72,11 @@ The available keys and their accepted values are reported in the table below.
...
+| health_check | `off` | Bool | No | Enable health check. If enabled, pgagroal will periodically check the health of the servers. |
+| health_check_period | 30 | Int | No | The interval in seconds between health check scans. |
+| health_check_timeout | 5 | String | No | The amount of time the process will wait for a response during a health check. If this value is specified without units, it is taken as seconds. It supports the following units as suffixes: 'S' for seconds (default), 'M' for minutes, 'H' for hours, 'D' for days, and 'W' for weeks. |
+| health_check_user | `postgres` | String | No | The user used for connecting to the health check. This user will also be used as the database name. |
```
jesperpedersen: This is typically a `SUPERUSER`. The default should be blank, and it should be required if `health_check` is `on`

jesperpedersen:
@trxvorr We need more documentation in the manual on how to set this up - with pros/cons

jesperpedersen:
@trxvorr CI, and conflict

fluca1978:
@trxvorr I think the code is fine, I have only some doubts about the documentation. I've written a few comments inline, most notably we are connecting to a database with the same name of the user, right? So we need to adjust and specify this detail (in the documentation you are referring the `postgres` database).

Once the documentation is fine, we can definitely merge.

fluca1978:
@jesperpedersen LGTM, PTAL I consider this complete.

jesperpedersen:
@fluca1978 Go ahead
@trxvorr Think about cross-porting

fluca1978:
@trxvorr thanks, merged with a modified commit message

PR #693, opened 2026-01-26
[#688]  feat: include PostgreSQL server connectivity in ping and status responses

### LINKED ISSUE #688: Improve `pgagroal-cli ping` command
`pgagroal-cli ping` command verifies if the pgagroal pooler is alive, but tells nothing about the availability of a PostgreSQL server behind the pooler.
It could be worth to enhance the `ping` command (via a subcommand) to check that pgagroal is alive, and it has a correctly configured and working PostgreSQL server (at least one) to serve connections on. In other words, this *super-ping* should return information about the fact that we can effectively connect to a server behind the pooler.

Partially related to #687

shashidhar-bm:
- Closes #688

#### example output:  (as per latest changes)
```
  ServerVersion: 2.1.0
  Servers:
    primary:
      Host: 127.0.0.1
      Port: 5433
      Primary: Yes
      Status: Running
    standby:
      Behind: 5041040
      Host: 127.0.0.1
      Port: 5434
      Primary: No
      Status: Running

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -356,7 +356,7 @@ usage(void)
    printf("                           - 'idle' to flush only idle connections\n");
    printf("                           - 'all' to flush all connections. USE WITH CAUTION!\n");
    printf("                           If no [database] name is specified, applies to all databases.\n");
-   printf("  ping                     Verifies if pgagroal is up and running\n");
+   printf("  ping                     Verifies if pgagroal is up and checks PostgreSQL server connectivity\n");
    printf("  enable   [database]      Enables the specified databases (or all databases)\n");
    printf("  disable  [database]      Disables the specified databases (or all databases)\n");
    printf("  shutdown [mode]          Stops pgagroal pooler. The [mode] can be:\n");
--- a/src/include/management.h
+++ b/src/include/management.h
@@ -97,6 +97,7 @@ extern "C" {
  */
 #define MANAGEMENT_ARGUMENT_ACTIVE_CONNECTIONS  "ActiveConnections"
 #define MANAGEMENT_ARGUMENT_APPNAME             "AppName"
+#define MANAGEMENT_ARGUMENT_BEHIND              "Behind"
 #define MANAGEMENT_ARGUMENT_CLIENT_VERSION      "ClientVersion"
 #define MANAGEMENT_ARGUMENT_COMMAND             "Command"
 #define MANAGEMENT_ARGUMENT_COMPRESSION         "Compression"
@@ -121,6 +122,7 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_PASSWORD            "Password"
 #define MANAGEMENT_ARGUMENT_PID                 "PID"
 #define MANAGEMENT_ARGUMENT_PORT                "Port"
+#define MANAGEMENT_ARGUMENT_PRIMARY             "Primary"
 #define MANAGEMENT_ARGUMENT_RESTART             "Restart"
 #define MANAGEMENT_ARGUMENT_SERVER              "Server"
 #define MANAGEMENT_ARGUMENT_SERVERS             "Servers"
@@ -131,7 +133,6 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_HEALTH              "Health"
 #define MANAGEMENT_ARGUMENT_STATUS              "Status"
 #define MANAGEMENT_ARGUMENT_TIME                "Time"
-#define MANAGEMENT_ARGUMENT_TIMESTAMP           "Timestamp"
 #define MANAGEMENT_ARGUMENT_TIMESTAMP           "Timestamp"
 #define MANAGEMENT_ARGUMENT_TOTAL_CONNECTIONS   "TotalConnections"
 #define MANAGEMENT_ARGUMENT_USERNAME            "Username"
--- a/src/libpgagroal/server.c
+++ b/src/libpgagroal/server.c
@@ -32,9 +32,9 @@
 #include <logging.h>
 #include <message.h>
 #include <pool.h>
+#include <queries.h>
 #include <security.h>
 #include <server.h>
-#include <message.h>
 #include <network.h>
 #include <utils.h>
 #include <value.h>
@@ -393,7 +393,6 @@ query_system_identifier(int server_idx, char* identifier, size_t id_size, int* p
    int status;
    bool query_ready = false;
    int offset_q;
-   char buffer[8];
 
    config = (struct main_configuration*)shmem;
    srv = &config->servers[server_idx];
@@ -401,7 +400,7 @@ query_system_identifier(int server_idx, char* identifier, size_t id_size, int* p
    if (pgagroal_server_query_execute(server_idx,
                                      config->health_check_user,
                                      config->health_check_user,
-                                     "SELECT system_identifier, pg_control_version FROM pg_control_system()",
+                                     (char*)pgagroal_queries_system_identifier(),
                                      NULL, &fd) != 0)
    {
       return 1;
@@ -489,11 +488,7 @@ query_system_identifier(int server_idx, char* identifier, size_t id_size, int* p
       }
    }
 
-   /* Send Terminate */
-   buffer[0] = 'X';
-   pgagroal_write_int32(buffer + 1, 4);
-   pgagroal_write_socket(NULL, fd, buffer, 5);
-
+   (void)pgagroal_write_terminate(NULL, fd);
    pgagroal_disconnect(fd);
 
    pgagroal_log_debug("system_identifier: Server [%s] (%s:%d) = %s, pg_control_version = %d",
@@ -512,6 +507,102 @@ error:
       pgagroal_disconnect(fd);
    }
    return 1;
+}
+
+int
+pgagroal_server_get_connectivity_info(int server, char** status, char** primary, int64_t* behind_bytes)
+{
+   int fd = -1;
+   bool in_recovery = false;
+   char recovery_value[16];
+   char behind_value[64];
+   char* endptr = NULL;
+   struct main_configuration* config = (struct main_configuration*)shmem;
+
+   *status = "Down";
+   *primary = "Unknown";
+   *behind_bytes = -1;
+
+   if (server < 0 || server >= config->number_of_servers)
+   {
+      return 1;
+   }
+
+   if (strlen(config->health_check_user) == 0)
+   {
+      pgagroal_log_debug("ping/status: health_check_user is not configured, cannot perform complete server connectivity checks");
+      return 0;
+   }
+
+   if (pgagroal_server_query_execute(server,
+                                     config->health_check_user,
+                                     config->health_check_user,
+                                     (char*)pgagroal_queries_is_in_recovery(),
+                                     NULL, &fd))
+   {
+      goto done;
+   }
+
+   *status = "Running";
+
+   if (pgagroal_read_query_first_column_text(fd, recovery_value, sizeof(recovery_value)))
+   {
+      (void)pgagroal_write_terminate(NULL, fd);
+      pgagroal_disconnect(fd);
+      fd = -1;
+      goto done;
+   }
+
+   in_recovery = (recovery_value[0] == 't');
+   *primary = in_recovery ? "No" : "Yes";
+   (void)pgagroal_write_terminate(NULL, fd);
+   pgagroal_disconnect(fd);
+   fd = -1;
+
+   if (!in_recovery)
+   {
+      return 0;
+   }
+
+   if (pgagroal_server_query_execute(server,
+                                     config->health_check_user,
+                                     config->health_check_user,
+                                     (char*)pgagroal_queries_replication_lag_bytes(),
+                                     NULL, &fd))
+   {
+      return 0;
+   }
+
+   if (pgagroal_read_query_first_column_text(fd, behind_value, sizeof(behind_value)))
+   {
+      *behind_bytes = -1;
+   }
+   else if (behind_value[0] == '\0')
+   {
+      *behind_bytes = -1;
+   }
+   else
+   {
+      errno = 0;
+      *behind_bytes = (int64_t)strtoll(behind_value, &endptr, 10);
+      if (errno != 0 || endptr == behind_value || *endptr != '\0')
+      {
+         *behind_bytes = -1;
+      }
+   }
+
+   (void)pgagroal_write_terminate(NULL, fd);
+   pgagroal_disconnect(fd);
+   fd = -1;
+
+done:
+   if (fd != -1)
+   {
+      (void)pgagroal_write_terminate(NULL, fd);
+      pgagroal_disconnect(fd);
+   }
+
+   return 0;
 }
 
 int
--- a/doc/CLI.md
+++ b/doc/CLI.md
@@ -66,8 +66,7 @@ pgagroal-cli flush pgbench   # pgagroal-cli flush gracefully pgbench
 ```
 
 ### ping
-The `ping` command checks if [**pgagroal**](https://github.com/pgagroal/pgagroal) is running.
-In case of success, the command does not print anything on the standard output unless the `--verbose` flag is used.
+The `ping` command checks if [**pgagroal**](https://github.com/pgagroal/pgagroal) is running and verifies connectivity to configured PostgreSQL servers.
 
 Command
 
@@ -88,6 +87,10 @@ In the case [**pgagroal**](https://github.com/pgagroal/pgagroal) is not running,
 pgagroal-cli ping          # $? = 1
 Connection error on /tmp
 ```
+
+With **JSON** output (`-F json` / `--format json`), the response includes a **`Servers`** object: one entry per configured backend. For each server you get **`Host`**, **`Port`**, **`Status`** (`Running` if PostgreSQL accepted the connectivity check, else `Down`), and **`Primary`** (`Yes`, `No`, or `Unknown`).
+
+**`Behind`**: only on **standby** (replica) backends—replication lag in **bytes** (WAL received but not yet replayed there). **`0`** means caught up. **Omitted** for primaries, when the server is down, or when lag cannot be read (for example, authentication failure). For password methods (SCRAM, MD5), configure **`health_check_user`** and the matching entry in **`pgagroal_users.conf`** so the check can log in.
 
 ### enable
 Enables a database (or all databases).
@@ -163,6 +166,8 @@ pgagroal-cli status
 ```
 
 With the `details` subcommand, a more verbose output is printed with a detail about every connection.
+
+For each configured PostgreSQL server, **`status`** and **`status details`** include the same connectivity summary as **`ping`** (host, port, running/down, primary vs standby, and **`Behind`** on standbys—see [ping](#ping)).
 
 Example
 
--- a/doc/man/pgagroal-cli.1.rst
+++ b/doc/man/pgagroal-cli.1.rst
@@ -71,7 +71,8 @@ flush [mode] [database]
   If no [database] name is specified, applies to all databases.
 
 ping
-  Verifies if pgagroal is up and running
+  Verifies if pgagroal is up and running and checks connectivity to configured PostgreSQL servers.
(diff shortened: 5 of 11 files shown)

### CODE REVIEW

File: src/cli.c [hunk 1]
```diff
@@ -356,6 +365,7 @@ usage(void)
    printf("                           - 'all' to flush all connections. USE WITH CAUTION!\n");
    printf("                           If no [database] name is specified, applies to all databases.\n");
    printf("  ping                     Verifies if pgagroal is up and running\n");
+   printf("  ping server              Verifies if pgagroal is up and checks PostgreSQL server connectivity\n");
```
fluca1978: we should document as `"      server"`, i.e., subcommand of `ping` not a separated fully line

fluca1978 (changes requested):
This looks almost perfect to me.
One doubt I've is if the status should be returned as `true`/`false` (like `ping` does) and not `OK/KO`, I may have misleded you using such terms.
Second to that, assuming `ping server` reports a failure, should we report an exit status of failure too (I means `$?` not zero)?

jesperpedersen:
Why can't we have it in the `ping` command ?

fluca1978:
> Why can't we have it in the `ping` command ?

I guess it is just a matter of semantic: assume pgagroal is up but postgres is down, what is the overall status?
How can we monitor the two different problems (pgagroal/postgres) if we merge the commands? The only way is to extend `ping` to include information about every postgresql host too, and have every monitoring to care about such result.

jesperpedersen:
Yes, have a `pgagroal` status, and then the list of configured PostgreSQL server each with their status. It is up to the client how it will consider the status from all systems

jesperpedersen:
We can have this information in `status details` too

shashidhar-bm:
Should I extend the `ping` command itself to include both pgagroal health and PostgreSQL server connectivity status and remove the `ping server` subcommand?

jesperpedersen:
@shashidhar-bm Yes, I think so

shashidhar-bm:
i made `ping` command itself to contain server connectivity status , and updated the pr description

jesperpedersen:
What about

```
   primary:
      Host: localhost
      Port: 5432
      Status: true
   Status: true
```

and have the same in `status details`

fluca1978:
Looks good to me, I would only refactor the `pgagroal_test_connectivity` with an `error` path as specified in my comments.

jesperpedersen:
`pgagroal_disconnect` on `-1` will fail, so you can ignore return value and no guard

shashidhar-bm:
now `status details` looks like this
```
       . . . 
      StartTime: -1
      Timestamp: -1
      Username: ''
  Databases:
    - Database: *
      Enabled: true
  Limits: {}
  MaxConnections: 100
  NumberOfServers: 1
  ServerVersion: 2.0.0
  Servers:
    primary:
      Host: localhost
      Port: 5432
      Status: true
  Status: Running
```
@fluca1978 , I've also updated the `error` path

fluca1978:
Looks fine for me, I've only the doubt that the `status` label can be somehow confusing:

```
  Servers:
    primary:
      Host: localhost
      Port: 5432
      Status: true
  Status: Running
```

we have `Status : true` to indicate PostgreSQL is up and running (well, reachable), and `Status : Running` to indicate `pgagroal` is running. Also `pgagroal-cli ping` reports `Status : true` for `pgagroal` itself, and this is somehow inconsistent

@jesperpedersen what do you think? I think we should report `Status: Running` always for `pgagroal` (i.e.,fix `pgagroal-cli ping`). Moreover, do you believe the Servers->primary->status is confusing with regard to the `pgagroal` status? On the other hand it makes sense to have the same "status" label to check for a service status (either PostgreSQL or pgagroal)..

jesperpedersen:
Yeah, it should be the same across naming and values - likely a `char*`. So, `Running` or `Down` for PostgreSQL. Ideally, there should be a `Recovering` flag too.

fluca1978:
Please @shashidhar-bm change the `Status` value for each server to `Running`, the same as the `pgagroal-cli status` for `pgagroal`.
Also fix the `pgagroal-cli ping` output to report the `Status: Running` for `pgagroal`.

shashidhar-bm:
> Please @shashidhar-bm change the `Status` value for each server to `Running`, the same as the `pgagroal-cli status` for `pgagroal`. Also fix the `pgagroal-cli ping` output to report the `Status: Running` for `pgagroal`.

sure! i'll update the pr shortly

shashidhar-bm:
> Yeah, it should be the same across naming and values - likely a `char*`. So, `Running` or `Down` for PostgreSQL. Ideally, there should be a `Recovering` flag too.

I wonder in what cases we can flag `Recovering`, @jesperpedersen can you give me an idea?

jesperpedersen:
@shashidhar-bm Primary function is `SELECT pg_is_in_recovery();` - but that is also used for a replica instance. So, you have to other ways as well. I see it as an improvement to this feature

shashidhar-bm:
> @shashidhar-bm Primary function is `SELECT pg_is_in_recovery();` - but that is also used for a replica instance. So, you have to other ways as well. I see it as an improvement to this feature

i'll explore on this, this looks interesting for me!

fluca1978:
> > @shashidhar-bm Primary function is `SELECT pg_is_in_recovery();` - but that is also used for a replica instance. So, you have to other ways as well. I see it as an improvement to this feature
> 
> i'll explore on this, this looks interesting for me!

We need to consider the streaming replication info, but I'm wondering if at this stage a "simple" `pg_is_in_recovery` does suffice, leaving the room for improvement in another patch.

shashidhar-bm:
@jesperpedersen , @fluca1978 ,I've  implemented recovery detection using `pg_is_in_recovery()` for now, with a TODO comment about streaming replication info for future enhancement. Happy to improve it later if needed.

fluca1978:
> @jesperpedersen , @fluca1978 ,I've implemented recovery detection using `pg_is_in_recovery()` for now, with a TODO comment about streaming replication info for future enhancement. Happy to improve it later if needed.

Please see the `clang-format` issues and my inline comments.
Also, when and where we decided to switch the strings from `Primary` to `Running`? At this point it could be worth having a `role` value (i.e., `Primary`, `Primary not init`, and so on) and a `status` one (`running`, `down`). What do you think?

shashidhar-bm:
> Please see the `clang-format` issues and my inline comments. Also, when and where we decided to switch the strings from `Primary` to `Running`? At this point it could be worth having a `role` value (i.e., `Primary`, `Primary not init`, and so on) and a `status` one (`running`, `down`). What do you think?

@fluca1978, Agreed — I'll update the PR (shortly) to expose role and status as two separate fields.

shashidhar-bm:
@fluca1978, I've added Role along with Status per server,

```
  ServerVersion: 2.1.0
  Servers:
    primary:
      Host: localhost
      Port: 5432
      Role: Not init
      Status: Running
```

fluca1978:
@jesperpedersen PTAL, it looks fine to me.
Please note that we break backward compability in the `status` output since `state` has been changed by `role`, to support the addition of `status`. I believe `role` is much explanatory (e.g., `role : primary`) as well as `status` (e.g., `status : running`).

jesperpedersen:
I don't understand it, what about

```
ServerVersion: 2.1.0
  Servers:
    primary:
      Host: localhost
      Port: 5432
      Primary: Yes
      Status: Running
```

If `Status` is `Running` then we know the state so either `true` or `false` from `SELECT pg_is_in_recovery();`. If `Status` is `Down`, then `Primary` is `Unknown`.

The call has to goto PostgreSQL - we have functionality for that; we use it during startup.

The key is based on the output to find the primary server, which are replicas, and which are down

fluca1978:
> I don't understand it, what about
> 
> ```
> ServerVersion: 2.1.0
>   Servers:
>     primary:
>       Host: localhost
>       Port: 5432
>       Primary: Yes
>       Status: Running
> ```
> 
> If `Status` is `Running` then we know the state so either `true` or `false` from `SELECT pg_is_in_recovery();`. If `Status` is `Down`, then `Primary` is `Unknown`.

It was a previous implementation, but I asked to not throw away the `primary`, `primary not init` and so on information we have right now. Thoughts?

jesperpedersen:
@fluca1978 But, why do we need `not init` ? If a node goes down we have to clear its status, and look at it again once it is back online. So, `Status: Running | Down` (or Yes/No) and `Primary: Yes | No | Unknown`. Plus `Behind: NUMBER` for replication lag.

Or am I missing something ?

fluca1978:
> @fluca1978 But, why do we need `not init` ? If a node goes down we have to clear its status, and look at it again once it is back online. So, `Status: Running | Down` (or Yes/No) and `Primary: Yes | No | Unknown`. Plus `Behind: NUMBER` for replication lag.
> 
> Or am I missing something ?

I'm just reluctant to throw away information about the internal status of `pgagroal`, but I agree we should keep the user experience simple enough.

Please @shashidhar-bm fix and ensure that:
- `Role` is no more present (my fault)
- `Status` is either `Running` or `Down`
- `Primary` is either `Yes` or `No` or `Unknown`

jesperpedersen:
@fluca1978 

> internal status of pgagroal

We can stick it in the Prometheus interface. WDYT ?

fluca1978:
> @fluca1978
> 
> > internal status of pgagroal
> 
> We can stick it in the Prometheus interface. WDYT ?

Well, this is fine. Also this sounds to me like a good difference between the text output and the JSON output. The latter should have more things, the textual output could not have the `Role` attribute.
So far we are keeping JSON/text outputs in sync, this is not mandatory according to me.

shashidhar-bm:
```
primary:
Host: localhost
Port: 5432
Status: Running
Primary: Yes
```
Should I add replication lag here?

jesperpedersen:
@shashidhar-bm I would say yes

fluca1978:
> ```
> primary:
> Host: localhost
> Port: 5432
> Status: Running
> Primary: Yes
> ```
> 
> Should I add replication lag here?

Yes please, but in the standby section.

shashidhar-bm:
@jesperpedersen , @fluca1978 , Apologies for the delay, I forgot about this work 😅, noww I've updated the code as we discussed.

shashidhar-bm:
@jesperpedersen , @fluca1978,  PTAL !
The new output looks like this:
```
  ServerVersion: 2.1.0
  Servers:
    primary:
      Host: 127.0.0.1
      Port: 5433
      Primary: Yes
      Status: Running
    standby:
      Behind: 5041040
      Host: 127.0.0.1
      Port: 5434
      Primary: No
      Status: Running    
```

fluca1978:
@shashidhar-bm I've tested, but apparently there is something not working with `behind_bytes` and the lag detection:

```
% pgagroal-cli status details && psql -h localhost -U postgres -p 9432 -c 'SELECT COALESCE(pg_wal_lsn_diff(pg_last_wal_receive_lsn(), pg_last_wal_replay_lsn()), 0)::bigint;'
Header:
  ClientVersion: 2.1.0
  Command: status-details
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260330112625
Outcome:
  Status: true
  Time: 00:00:01
Request: {}
Response:
  ActiveConnections: 0
  Connections:
    - AppName: ''
      Database: pgbench
      FD: 26
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 27
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 28
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 29
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 30
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 31
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 32
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 33
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 34
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: pgbench
      FD: 35
      PID: -1
      StartTime: 1774862615
      Timestamp: 1774862615
      Username: pgbench
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
    - AppName: ''
      Database: ''
      FD: -1
      PID: -1
      StartTime: -1
      Timestamp: -1
      Username: ''
  Databases:
    - Database: *
      Enabled: true
  Limits:
    - ActiveConnections: 0
      Database: pgbench
      InitialConnections: 10
      MaxConnections: 20
      MinConnections: 5
      Username: pgbench
  MaxConnections: 40
  NumberOfServers: 2
  ServerVersion: 2.1.0
  Servers:
    rachel:
      Host: 127.0.0.1
      Port: 5432
      Primary: Yes
      Status: Running
    ross:
      Host: localhost
      Port: 9432
      Primary: No
      Status: Running
  Status: Running
  TotalConnections: 10
 coalesce  
-----------
 328247360
(1 row)
```

Please test and verify.

fluca1978:
Commit d239d129aac6faf684600b08132b01b45251b0b9 is almost fine to me, can you document better the meaning of the `behind` value in command output (users should know what it means)?
Also it would be great to have tests for the new value.

shashidhar-bm:
@fluca1978 , also is the current ping auth flow good enough?, and if so, should we document setting `health_check_user` before using ping?

fluca1978:
> @fluca1978 , also is the current ping auth flow good enough?, and if so, should we document setting `health_check_user` before using ping?

Commit https://github.com/pgagroal/pgagroal/commit/ecd594f891d9ab0de2f6f547e969220f79a596f4 introduces a more generic infrastructure for querying the server, please check and adapt.
And yes, we should document that `health_check_user` is required for a complete `ping` command to work.

fluca1978:
@jesperpedersen LGTM, PTAL.

src/include/management.h
```diff
@@ -130,9 +130,10 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_SYSTEM_IDENTIFIER   "SystemIdentifier"
 #define MANAGEMENT_ARGUMENT_HEALTH              "Health"
 #define MANAGEMENT_ARGUMENT_STATUS              "Status"
+#define MANAGEMENT_ARGUMENT_PRIMARY             "Primary"
+#define MANAGEMENT_ARGUMENT_BEHIND              "Behind"
```
jesperpedersen: Sorting

src/libpgagroal/server.c
```diff
@@ -508,6 +509,204 @@ query_system_identifier(int server_idx, char* identifier, size_t id_size, int* p
    return 1;
 }
 
+static void
```
jesperpedersen: This belongs in message.h

src/libpgagroal/server.c
```diff
@@ -508,6 +509,204 @@ query_system_identifier(int server_idx, char* identifier, size_t id_size, int* p
...
+      return;
+   }
+
+   buffer[0] = 'X';
+   pgagroal_write_int32(buffer + 1, 4);
+   pgagroal_write_socket(NULL, fd, buffer, 5);
+   pgagroal_disconnect(fd);
+}
+
+static int
```
jesperpedersen: Look to port queries.h from pgmoneta

src/libpgagroal/server.c
```diff
@@ -508,6 +509,204 @@ query_system_identifier(int server_idx, char* identifier, size_t id_size, int* p
...
+      msg = NULL;
+   }
+
+error:
+   pgagroal_clear_message(msg);
+   return 1;
+}
+
+int
+pgagroal_server_get_connectivity_info(int server, char** status, char** primary, int64_t* behind_bytes)
```
jesperpedersen, line 513: Use queries.h

jesperpedersen (changes requested):

fluca1978:
Merged, thanks @shashidhar-bm .
Rewritten commit message.

Issue #703, opened 2026-02-02
memory leak in encrypt_decrypt_buffer (aes.c)


trxvorr:
a memory leak exists in aes.c:encrypt_decrypt_buffer. the *res_buffer pointer is allocated via malloc but is never freed if an error occurs and the function jumps to the error label. this can happen if pgagroal_get_master_key, dervive_key_iv, EVP_CIPHER_CTX or any of the EVP_CIPHER... functions fail. found this during the implementation of health check probes.

PR #704, opened 2026-02-03
[#703] fix memory leak in encrypt_decrypt_buffer error path

trxvorr:
when encryption/decryption fails, the allocated res_buffer was not being freed, causing a memory leak. this adds proper cleanup in the error handling section.

fixes: #703

### DIFF
--- a/src/libpgagroal/aes.c
+++ b/src/libpgagroal/aes.c
@@ -353,10 +353,10 @@ error:
       EVP_CIPHER_CTX_free(ctx);
    }
 
-   if (master_key)
-   {
-      free(master_key);
-   }
+   free(master_key);
+
+   free(*res_buffer);
+   *res_buffer = NULL;
 
    return 1;
 }


jesperpedersen:
@trxvorr Can you compare this with the pgmoneta implementation ?

trxvorr:
@jesperpedersen i compared this with the pgmoneta implementation of encrypt_decrypt_buffer.

pgmoneta correctly performs the cleanup in the error path free(*res_buffer) whereas the current pgagroal code was missing it which caused the leak.

i noticed pgmoneta calls free() directly whereas my PR currently includes an explicit if (*res_buffer) check.

would you prefer i remove the if check to make the implementation consistent with pgmoneta or is the current version acceptable?

fluca1978:
> @jesperpedersen i compared this with the pgmoneta implementation of encrypt_decrypt_buffer.
> 
> pgmoneta correctly performs the cleanup in the error path free(*res_buffer) whereas the current pgagroal code was missing it which caused the leak.
> 
> i noticed pgmoneta calls free() directly whereas my PR currently includes an explicit if (*res_buffer) check.
> 
> would you prefer i remove the if check to make the implementation consistent with pgmoneta or is the current version acceptable?

`free` handles correctly NULL pointers, so we can skip the defensive `if`.
Please remove it, then I'll merge.

trxvorr:
@fluca1978 @jesperpedersen i removed the if for both master key and *res_buffer just like how pgmoneta handles it
i hope it's fine now

jesperpedersen:
Rebased, and merged.

PR #723, opened 2026-02-16
[#722] Disable io_uring backend on older kernels lacking required features

Userfrom1995:
[#722] Disable io_uring backend on older kernels lacking required features


jesperpedersen:
Merged.

PR #726, opened 2026-02-18
[#725] Fix connection worker title truncation

### LINKED ISSUE #725: Process title truncated in connection workers when binary path is short
### Summary
In `pgagroal`, process titles on are set by overwriting the `argv[]` memory region. This region's size is calculated once during initialization and stored in a static variable. However, because connection workers are created via `fork()`, this static state is reset in the child, and subsequent attempts to set the title in the worker recalculate the buffer size using a hardcoded `argc=1`.

This leads to silent truncation whenever the combined length of the desired process title exceeds the length of `argv[0]` plus a null terminator.

### Root Cause
changing a process title involves modifying the memory pointed to by the `argv` array passed to `main()`. The total space available for the title is the sum of the lengths of all original command-line arguments (including null terminators).

**Hardcoded argc:** All calls to `pgagroal_set_proc_title` in `src/libpgagroal/worker.c` hardcode `argc=1`. This forces the worker to recalculate the buffer size using *only* `argv[0]`.

```c
// src/libpgagroal/worker.c (BEFORE fix)
pgagroal_set_proc_title(1, argv, "authenticating", NULL);
```

If the binary is started as `/usr/bin/pgagroal`, the buffer is ~18 bytes. If it's started as `/tmp/p`, the buffer is only 7 bytes. Since a typical title like `pgagroal: authenticating` is 24 bytes, it gets truncated to the size of `argv[0]`.

### Reproduction

The truncation can be reliably observed by running the binary from a short path.

```bash
# Setup: Copy to a short path to minimize the argv[0] buffer
cp ./build/src/pgagroal /tmp/p

# Start the process from the short path
PGAGROAL_CONFIG_DIR=./build/config /tmp/p &

# Observe the process title using ps
ps -eo pid,args | grep -E "pgagroa|/tmp/p" | grep -v grep
# Output: 12345  pgagroa   ← Truncated at 7 bytes
```

### Affected Components

The following files were identified as contributing to the inability of the worker to use the correct `argc` value:

- `src/libpgagroal/worker.c`: Hardcoded `argc=1` in title setter calls.
- `src/include/worker.h`: `pgagroal_worker()` signature lacked `argc`.
- `src/include/utils.h`: `accept_io` struct lacked `argc` field for transport.
- `src/main.c`: `argc` was not captured globally or forwarded to `accept_io`.

trxvorr:
The connection worker hardcoded argc=1, causing truncation on short paths. 
Pass real argc to ensure correct buffer sizing.
fixes #725

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,3 +29,4 @@ Ankush Mondal <mondalankush9851@gmail.com>
 Shashank Singh <shashanksgh3@gmail.com>
 Mazen Kamal <mazenkamal212@gmail.com>
 Shashidhar B M <shashidhar.i.0119@gmail.com>
+Trevor Jacob Mathews <trevorjacobmathews@gmail.com>
--- a/doc/manual/en/97-acknowledgement.md
+++ b/doc/manual/en/97-acknowledgement.md
@@ -36,6 +36,7 @@ Ankush Mondal <mondalankush9851@gmail.com>
 Shashank Singh <shashanksgh3@gmail.com>
 Mazen Kamal <mazenkamal212@gmail.com>
 Shashidhar B M <shashidhar.i.0119@gmail.com>
+Trevor Jacob Mathews <trevorjacobmathews@gmail.com>
 
 ```
 
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -706,6 +706,34 @@ pgagroal_set_proc_title(int argc, char** argv, char* s1, char* s2)
       return;
    }
 
+   // compute how long was the command line
+   // when the application was started
+   if (max_process_title_size == 0)
+   {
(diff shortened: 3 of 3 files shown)


src/include/worker.h
```diff
@@ -64,12 +64,13 @@ extern volatile int exit_code;
...
 /**
  * Create a worker instance
+ * @param argc The argc
  * @param fd The client descriptor
  * @param address The client address
  * @param argv The argv
  */
 void
-pgagroal_worker(int fd, char* address, char** argv);
+pgagroal_worker(int argc, int fd, char* address, char** argv);
```
jesperpedersen: `int argc` should be before `char** argv`

fluca1978:
thanks @trxvorr

Issue #727, opened 2026-02-18
Memory leak in MCTF_FINISH() macro - mctf_errmsg not freed before reset


Abdelrhmansersawy:
**Describe the bug**

While investigating this CI run https://github.com/pgagroal/pgagroal/actions/runs/22102958574/job/63970227559?pr=714#step:10:614, I discovered a memory leak in MCTF.

The `MCTF_FINISH()` macro  sets `mctf_errmsg = NULL` without freeing previously allocated memory, causing a memory leak when `mctf_errmsg` is assigned by `mctf_format_error()` before reaching `cleanup`.

**Root Cause**

```c
#define MCTF_FINISH()     \
   do                     \
   {                      \
      mctf_errno = 0;     \
      mctf_errmsg = NULL; /* leaks allocated error message */
      return 0;           \
   }                      \
   while (0)
```

`mctf_format_error()` allocates memory via `malloc()`. If a test assertion fails, `mctf_errmsg` points to this allocation, which is then overwritten without `free()`.

Abdelrhmansersawy:
@jesperpedersen Please confirm - if that’s correct, I’ll open a PR to fix this first, then continue working on #714.

jesperpedersen:
@shashidhar-bm Cc:

shashidhar-bm:
@Abdelrhmansersawy , Thanks for reporting — it is a bug.
 There are global free statements for `mctf_errmsg` in the runner, but `MCTF_FINISH()` sets it to `NULL` before control returns there, so the runner never sees that allocation. 
We could either remove that line and let the runner free it, or free it in the macro before nulling

jesperpedersen:
@shashidhar-bm See what the best patch is

shashidhar-bm:
Oh no, this is a serious mistake that happened during porting — apologies. The initial design of `MCTF_FINISH()` is to just return `mctf_errno`, and all the memory freeing happens on the way to the next test run (i.e, in the runner’s `FAIL`/`SKIP` branches and the pre‑run cleanup).

shashidhar-bm:
> [@jesperpedersen](https://github.com/jesperpedersen) Please confirm - if that’s correct, I’ll open a PR to fix this first, then continue working on [#714](https://github.com/pgagroal/pgagroal/pull/714).

@Abdelrhmansersawy, would you like to fix this, or should I proceed?

Abdelrhmansersawy:
[...]

PR #728, opened 2026-02-19
[#727] test: redesign MCTF_FINISH to return mctf_errno

shashidhar-bm:
- fixes: #727 

Avoid nulling `mctf_errmsg` in `MCTF_FINISH()` so the test runner can still copy and free the error message in its `SKIP`/`FAIL` paths.


jesperpedersen:
Merged.

Issue #692, opened 2026-01-26
Add `status standbys` to `pgagroal-cli`


fluca1978:
The `pgagroal-cli status` command (and its friend `status details`) show details about status and configuration of pooler and servers.
I suggest adding a new `pgagroal-cli status standbys` command that will report, for every standby (if any), if it is correctly streaming from the currently selected primary.

jesperpedersen:
I think we can just have it in `status details`

ankushT369:
@jesperpedersen @fluca1978 Assign this to me

fluca1978:
> [@jesperpedersen](https://github.com/jesperpedersen) [@fluca1978](https://github.com/fluca1978) Assign this to me

go for it!

ankushT369:
@fluca1978  
ok the possible solution i want to discuss first which came to my mind is 
- we have to get the information of currently streaming dbs from primary db
- then check those particular dbs if they are good
is this ok ?

also do i have to add a new ```State``` field ??

fluca1978:
> [@fluca1978](https://github.com/fluca1978) ok the possible solution i want to discuss first which came to my mind is
> 
>     * we have to get the information of currently streaming dbs from primary db
> 

No, we have a list of servers configured in `pgagroal.conf`, hence in shared memory. We know one of them is primary, the others are standbys, we want to check the latter. If the user has configured other streaming replication servers that are not listed in pgagroal, we don't care.


>     * then check those particular dbs if they are good
>       is this ok ?

We can add any level of detail in here: are they good? Are they streaming? Are they lagging?
First task is: are they alive?


> 
> also do i have to add a new `State` field ??

You can use what state field is already there for each server, and if you believe it is worth, add other fields for details. But keep it simple first.

ankushT369:
@fluca1978 there is another condition i am facing like if ```pgagroal``` is connected with 2 database one is ```PRIMARY``` and another is let's say ```SECONDARY``` both are exposed in different ports  (```5432```, ```5433``` respectively) but the ```SECONDAY``` is not streaming from ```PRIMARY``` like I made it like a standalone DB(```SECONDAY```) should i care for it ?? because according to the code : 
```c
   if (state == 'f')
   {
      atomic_store(&config->servers[server].state, SERVER_PRIMARY);
   }
   else
   {
// if its streaming then its REPLICA. 
      atomic_store(&config->servers[server].state, SERVER_REPLICA);
   }
```

[...]

PR #735, opened 2026-02-22
[#692] add isalive to pgagroal-cli

ankushT369:
#692 I only added isalive is this approach ok ?? like i didnt stored anything like is_alive variable in the main_configuration @fluca1978


fluca1978:
I don't understand the aim of this, we already have `ping`, so what is the purpose and how is it related to #692 ?
Also please consider that #692 should be probably be implemented alike <https://github.com/pgagroal/pgagroal/pull/693> or in conjunction with it.

ankushT369:
@fluca1978 ok i am taking another approach like it will contains another section like it will have list of streaming dbs and the attributes will streaming, lagging, alive??

is this ok ??

fluca1978:
> @fluca1978 ok i am taking another approach like it will contains another section like it will have list of streaming dbs and the attributes will streaming, lagging, alive??
> 
> is this ok ??


It could be. Please note there is already an implementation-WIP about the "is replica running" into #693. The value you could add is about the replication information, like `primary_conninfo` and `primary_slot_:name` into another section of `status details`. How does it sound?

ankushT369:
> > @fluca1978 ok i am taking another approach like it will contains another section like it will have list of streaming dbs and the attributes will streaming, lagging, alive??
> > is this ok ??
> 
> It could be. Please note there is already an implementation-WIP about the "is replica running" into #693. The value you could add is about the replication information, like `primary_conninfo` and `primary_slot_:name` into another section of `status details`. How does it sound?

good

ankushT369:
@fluca1978 hey i am facing a problem the function ```pgagroal_prefill_auth``` always returns find_best_rule thing which is always primary. Instead i am trying to make it like i will have list of server which i will iterate first i will do a minimal auth with my own func then execute query ?? what do you think letme know!!

fluca1978:
> @fluca1978 hey i am facing a problem the function `pgagroal_prefill_auth` always returns find_best_rule thing which is always primary. Instead i am trying to make it like i will have list of server which i will iterate first i will do a minimal auth with my own func then execute query ?? what do you think letme know!!

It is not clear to me why are you stepping thru `pgagroal_prefill_auth`, which clearly has to find the `primary` server and has to resolve any alias (`find_best_rule`), but I don't understand why are you dealing with this.
You already know which servers are not primary, so you can go to them directly, or am I missing something?

ankushT369:
> > @fluca1978 hey i am facing a problem the function `pgagroal_prefill_auth` always returns find_best_rule thing which is always primary. Instead i am trying to make it like i will have list of server which i will iterate first i will do a minimal auth with my own func then execute query ?? what do you think letme know!!
> 
> It is not clear to me why are you stepping thru `pgagroal_prefill_auth`, which clearly has to find the `primary` server and has to resolve any alias (`find_best_rule`), but I don't understand why are you dealing with this. You already know which servers are not primary, so you can go to them directly, or am I missing something?

Yes, I understand which servers are non‑primary, but my goal is to check their state — for example, whether they are in recovery mode, whether they are streaming, etc.  
To get this information, I need to run queries on those **non‑primary servers**, and to run those queries I first need to **authenticate** to them.  
That’s why I’m trying to figure out the correct way to perform authentication for these non‑primary servers.

So my questions are:

1.  **What is the recommended way to authenticate to non‑primary servers?**  
    Should I create a separate function for this, or is there an existing function in pgagroal?

2.  **Regarding standby detection:**  
    After identifying all servers that are not primary, do I classify a server as a standby only if `pg_is_in_recovery()` returns `true`?  
    Or should all non‑primary servers automatically be considered standbys?

Please let me know if my understanding is correct or if I’m missing something.

fluca1978:
>     1. **What is the recommended way to authenticate to non‑primary servers?**
>        Should I create a separate function for this, or is there an existing function in pgagroal?
> 

`pgagroal_get_connection` does the work you need, only that it stores it into the pooled connections and, internally, resolves to a primary server.  I would write a `pgagroal_get_standby_connection` that does a similar job without storing the connection in the pool and without selecting a primary. Internally it should be able to use pagroal_authenticate` *without* the call to the `get_primaru`, so I think adding a new parameter `server` to the arguments of `pgagroal_authenticate` and leave it `NULL` for all the exiting calls (so, -> `get_primary`) and skip the `get_primary` if `server` is already pointing to a standby.
@jesperpedersen thoughts?

>     2. **Regarding standby detection:**
>        After identifying all servers that are not primary, do I classify a server as a standby only if `pg_is_in_recovery()` returns `true`?
>        Or should all non‑primary servers automatically be considered standbys?

So far, every server with `primary = off` is considered as a standby. Once we can get information from the PostgreSQL (i.e., `pg_is_in_recovery`) we can reason about how to manage the status of a server, so that we can detect configuration inconsistencies.

jesperpedersen:
@ankushT369 @fluca1978 Yes, create a `pgagroal_get_standby_connection`. Make sure that as much code is shared as possible

fluca1978:
> @ankushT369 @fluca1978 Yes, create a `pgagroal_get_standby_connection`. Make sure that as much code is shared as possible


Should we keep these *standby connections* around for efficiency? In such case, I would propose to make a `reserved_max_connections` alike tunable to control how many connections to standby keep around.
This can be implemented in a follow up.

ankushT369:
> > @ankushT369 @fluca1978 Yes, create a `pgagroal_get_standby_connection`. Make sure that as much code is shared as possible
> 
> Should we keep these _standby connections_ around for efficiency? In such case, I would propose to make a `reserved_max_connections` alike tunable to control how many connections to standby keep around. This can be implemented in a follow up.

sounds good

ankushT369:
@fluca1978 Ive been busy whole week I will complete this soon.

fheikens:
Triage note: this is currently conflicting with master. Could you rebase on the latest master? The checks were green otherwise, so a rebase should make it reviewable.

PR #738, opened 2026-02-23
[#504] Optimizing Salt by removing need to reinitialize HMAC

### LINKED ISSUE #504: Optimize SALT
Optimize the SALT implementation, and see how it affects iteration count.

https://www.postgresql.org/message-id/flat/Z6BsNAadbID%2Bk2L4%40bass1 can be used for ideas

Zeyad-Daowd:
**The Modification: HMAC Context Reuse**

The Issue: The original implementation re-initializes the HMAC context with the password in every iteration (4,096 times).
The Fix: Following OpenSSL's documentation, I modified the loop to use HMAC_Init_ex(ctx, NULL, 0, NULL, NULL). and removed the HMAC_CTX_reset from the loop since it is no longer necessary.

This was based on the following openssl’s documentation page https://docs.openssl.org/3.0/man3/HMAC/#description

image

Why it matters: Passing NULL as the key and message digest tells OpenSSL to reuse the pre-processed key state already in the context, skipping the expensive re-initialization and resetting context is no longer necessary.

**Benchmarking & Performance Analysis:**

I isolated the salted_password function in a standalone file to be able to benchmark it smoothly,
I benchmarked the change by calling salted_password 1,000 times (at 4,096 iterations each) to try to minimize noise of running from a time to time.

Correctness: I verified that the optimized version produces the exact same hex output as the original implementation.

CPU Profiling: I used perf to be able to gain an estimate of where the bottleneck is to try to optimize.

Before: HMAC_Init_ex was responsible for ~66.6% of total CPU overhead which was more than hashing itself.

image

After: Overhead dropped to ~14%, with the CPU now spending the majority of its time on hashing itself 
image

Total Speedup: The optimized function is nearly 5x faster (reducing local execution time from ~2600ms to 577ms for calling the function 1000 times).

before:
image

after:image

This ~4.5x improvement can allow us to increase iterations count by a factor of 4, which from my understanding results in better security, and still if this is wanted.

I have also tried to increase the iterations count and it resulted in less time than the original 4096 iteration without the modification
image
Note: I haven't increased the iteration count in the PR since I believe this requires approval
Closes #504

### DIFF
--- a/src/libpgagroal/security.c
+++ b/src/libpgagroal/security.c
@@ -4023,12 +4023,8 @@ salted_password(char* password, char* salt, int salt_length, int iterations, uns
 
    for (int i = 2; i <= iterations; i++)
    {
-      if (HMAC_CTX_reset(ctx) != 1)
-      {
-         goto error;
-      }
-
-      if (HMAC_Init_ex(ctx, password, password_length, EVP_sha256(), NULL) != 1)
+      /* passing nulls cause function to reuse the same key / password in the context */
+      if (HMAC_Init_ex(ctx, NULL, 0, NULL, NULL) != 1)
       {
          goto error;
       }
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,3 +31,4 @@ Mazen Kamal <mazenkamal212@gmail.com>
 Shashidhar B M <shashidhar.i.0119@gmail.com>
 Trevor Jacob Mathews <trevorjacobmathews@gmail.com>
 Abdelrhman Sersawy <abdelrhmansersawy@gmail.com>
+Zeyad Daowd <zeyaddaowd@yahoo.com>
--- a/doc/manual/en/97-acknowledgement.md
+++ b/doc/manual/en/97-acknowledgement.md
@@ -38,7 +38,7 @@ Mazen Kamal <mazenkamal212@gmail.com>
 Shashidhar B M <shashidhar.i.0119@gmail.com>
 Trevor Jacob Mathews <trevorjacobmathews@gmail.com>
 Abdelrhman Sersawy <abdelrhmansersawy@gmail.com>
-
+Zeyad Daowd <zeyaddaowd@yahoo.com>
 ```
 
 ## Committers


jesperpedersen:
@Zeyad-Daowd Remember, doc/manual/en/97-acknowledgement.md. See DEVELOPERS.md

Zeyad-Daowd:
@jesperpedersen I've updated doc/manual/en/97-acknowledgement.md and amended the commit. Thank you for the reminder.

jesperpedersen:
@Zeyad-Daowd Did you check if there are any OpenSSL parameters we can set based on run-time CPU feature detection ?

fluca1978:
Looks fine to me, but I'm guessing if it is worth to migrate to `EVP_MAC` (<https://docs.openssl.org/3.0/man3/EVP_MAC/#context-manipulation-functions>). @Zeyad-Daowd what do you think?

Zeyad-Daowd:
@jesperpedersen I understood by your question whether we can set a parameter in openssl to optimize hashing operations themselves.
I am not sure if I understood it correctly, please correct me if i am wrong
After searching, I couldn't find a parameter in HMAC functions that can use cpu detection but i found this
https://docs.openssl.org/3.0/man3/OPENSSL_ia32cap/
image
from my understanding openssl uses this to choose between different code paths for cryptographic implementations based on different cpu architectures, so I think that this is handled internally by openssl.

jesperpedersen:
@Zeyad-Daowd I just wanted to make sure that we are using the CPU

src/libpgagroal/security.c
```diff
@@ -4023,12 +4023,7 @@ salted_password(char* password, char* salt, int salt_length, int iterations, uns
 
    for (int i = 2; i <= iterations; i++)
    {
-      if (HMAC_CTX_reset(ctx) != 1)
-      {
-         goto error;
-      }
-
-      if (HMAC_Init_ex(ctx, password, password_length, EVP_sha256(), NULL) != 1)
+      if (HMAC_Init_ex(ctx, NULL, 0, NULL, NULL) != 1) // passing nulls cause function to reuse the same key / password in the context
```
jesperpedersen: Please, move the comment above the `if` and use `/* */`

Zeyad-Daowd:
@fluca1978 regarding migration to EVP_MAC,
Talking from the performance side, I am not sure if migration to EVP_MAC would result in better performance than the current implementation using HMAC functions. But I think benchmarking a new implementation using EVP_MAC may be benefetial to know if that is the case.

Regarding maintainability and being future proof, the currently used HMAC functions are deprecated in openssl 3
image
Also it is mentioned in openSSL wiki, https://wiki.openssl.org/index.php/OpenSSL_3.0#Providers_and_FIPS_support
that the deprecated functions may be removed from future versions
image
I believe this may be an important point to consider to evaluate the migration.

An important note: you have previously told me about the importance of backward compatibility, and thinking from this side.
I feel like a full migration to EVP_MAC may break compatability for users with openssl 1.1 and I think this is undesirable.
The first solution that came to mind is to check openssl version in the code and then using # if  to either use the new EVP_MAC implementation or the old HMAC based implementation for users having old openssl.
similar to this:
image
image

I feel like this might add some complexity to the code but do you think it is worth keeping backward compatibility?

jesperpedersen:
@Zeyad-Daowd You only need to worry about OpenSSL 3.x

Zeyad-Daowd:
@jesperpedersen Regarding CPU usage: from my understanding, openSSL handles that under the hood using openssl_ia32cap
regarding openssl 3.x: Great, I believe that simplifies it as we will just focus on the new EVP_MAC without worrying about old HMAC. I think in that case it is worth it, for the long-term maintainability, to migrate to EVP_MAC
In case you agree, should I start the migration?

src/libpgagroal/security.c
Zeyad-Daowd: Done, I've moved the comment above the if block & used /* */ format. I've just pushed the update to the branch

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Zeyad-Daowd You likely need to cross-port this to pgmoneta and pgexporter

fluca1978:
> @fluca1978 regarding migration to EVP_MAC, Talking from the performance side, I am not sure if migration to EVP_MAC would result in better performance than the current implementation using HMAC functions. But I think benchmarking a new implementation using EVP_MAC may be benefetial to know if that is the case.

Performances are important, but security matters the most.
Moreover, I would not like to see an `#ifdef` maze just to pick up the "most decent OpenSSL version", and as Jesper wrote, we can take into account OpenSSL 3, hence my question about migrating to EVP_xxx.

It could be interesting to fully evaluate this migration in the future.

Zeyad-Daowd:
Thank you for the merge, @jesperpedersen. I'm glad the optimization is in.
I'll start working on the cross-ports for pgmoneta and pgexporter right away

Zeyad-Daowd:
@fluca1978 I agree, avoiding the #ifdef will be result in a much cleaner codebase to read.
Since we are considering openssl 3 only, the EVP_ migration becomes much cleaner.

should we open a separate tracking issue for the full EVP migration as so we can evaluate it? we could add our current findings there to make the migration easier to evaluate and implement

jesperpedersen:
@Zeyad-Daowd Yes, open an issue, please

Zeyad-Daowd:
@jesperpedersen Done, I have opened issue #752

Issue #736, opened 2026-02-23
pgagroal.socket only works on the 2nd connection


df7cb:
systemd socket activation only works on the 2nd connection:

```
sudo systemctl stop pgagroal.socket
sudo systemctl stop pgagroal
sudo systemctl start pgagroal.socket
Feb 23 13:08:38 turing systemd[1]: Listening on pgagroal.socket - Sockets for pgagroal.

sudo PGPASSWORD=agroalpass psql -h /run/postgresql -p 2345 -U agroaltestuser postgres
Feb 23 13:08:53 turing systemd[1]: Starting pgagroal.service - High-performance connection pool for PostgreSQL...
Feb 23 13:08:53 turing systemd[1]: Started pgagroal.service - High-performance connection pool for PostgreSQL.
Feb 23 13:08:53 turing pgagroal[882559]: 2026-02-23 13:08:53 INFO  main.c:1290 pgagroal: 2.0.1 started on localhost:2345

(psql doesn't connect)
^C

sudo PGPASSWORD=agroalpass psql -h /run/postgresql -p 2345 -U agroaltestuser postgres
psql (19devel (Debian 19~~devel-3~20260223.0815.g2f2c9d83637.pgdg+1), Server 18.2 (Debian 18.2-1.pgdg+1))
Geben Sie »help« für Hilfe ein.

postgres=> 
\q

Feb 23 13:09:07 turing pgagroal[882559]: 2026-02-23 13:09:07 WARN  connection.c:98 pgagroal_management_transfer_connection: get_pid connect = -1, pid = 883176, f = [.s.pgagroal.883176]
```

There is also the issue that pgagroal.socket cannot be (re)started while pgagroal.service is already running:
```
Feb 23 13:15:12 turing systemd[1]: pgagroal.socket: Socket service pgagroal.service already active, refusing.
```
but I am not sure much much this is supposed to work.

jesperpedersen:
@df7cb Can you attach your `.service` files ?

df7cb:
It's the two files from doc/etc/ with these two patches on top: https://salsa.debian.org/postgresql/pgagroal/-/tree/master/debian/patches

```
 cat doc/etc/pgagroal.service 
# systemd service unit for pgagroal
#
# - Adjust the user running the service in User
# - Adjust the path in ExecStart
#
[Unit]
Description=High-performance connection pool for PostgreSQL
Documentation=man:pgagroal(1)
Documentation=https://pgagroal.github.io/
[...]

PR #740, opened 2026-02-25
[#739] Fix memory corruption in systemd socket initialization

### LINKED ISSUE #739: Memory corruption in systemd socket initialization due to incorrect pointer arithmetic
theer is a memory corruption bug during systemd socket initialization in `src/main.c`. 

when `pgagroal` processes inherited file descriptors from systemd, it attempts to store the file descriptor in the `main_fds` array using the following pointer arithmetic:
```c
*(main_fds + (m * sizeof(int))) = fd;

trxvorr:
fixes #739

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -896,7 +896,7 @@ read_superuser_path:
          }
          else if (sd_is_socket(fd, AF_INET, 0, -1) || sd_is_socket(fd, AF_INET6, 0, -1))
          {
-            *(main_fds + (m * sizeof(int))) = fd;
+            main_fds[m] = fd;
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@trxvorr Could you verify #736 with this ?

trxvorr:
@jesperpedersen while this PR definitely fixes the memory bounds issue with the file descriptor array, it doesn't fully resolve #736. The daemon still hangs on the first connection attempt and drops the socket

trxvorr:
@jesperpedersen i got it to work on the first attempt ill close this pr and issue and link a pr directly to #736

PR #744, opened 2026-02-25
[#743] Fix uninitialized client_addr struct in accept()

### LINKED ISSUE #743: Uninitialized client_addr struct in accept() reads garbage stack memory
**Describe the bug**
In the socket acceptance logic (e.g., `accept_main_cb`), a local stack variable `struct sockaddr_in6 client_addr;` is declared to hold the incoming client's IP address. However, the `accept()` system call is currently being passed `NULL` for the address and length pointers. 

Because local stack variables are not implicitly zeroed out in C, and `accept()` isn't filling it with data, `client_addr` retains whatever random binary garbage was left in that memory location by previous function calls. 

connected via TCP from 127.0.0.1 and checked journal logs at TRACE level.
result: TRACE main.c:1491 accept_main_cb: client address: 2100::
this confirms that accept() is not populating the address struct due to the NULL pointers.

trxvorr:
fixes #743

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1477,6 +1477,10 @@ accept_main_cb(struct io_watcher* watcher)
 
    pgagroal_prometheus_client_sockets_add();
 
+   memset(&client_addr, 0, sizeof(struct sockaddr_in6));
+   socklen_t client_addr_length = sizeof(struct sockaddr_in6);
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Rebased, and merged.

jesperpedersen:
@trxvorr This likely need to be cross-ported to pgmoneta and pgexporter

PR #745, opened 2026-02-25
[#742] Add O_NONBLOCK to systemd sockets

### LINKED ISSUE #742: Systemd inherited sockets are in blocking mode, causing potential event loop deadlocks
when `pgagroal` inherits file descriptors via systemd socket activation (`sd_listen_fds`), the sockets are passed in blocking mode by default. as `pgagroal` utilizes asynchronous event loops (`io_uring` / `epoll`), it expects all registered file descriptors to be non-blocking. 

inspecting the file descriptor flags of the running daemon via `/proc/<pid>/fdinfo/<fd>` shows `flags: 02` (standard `O_RDWR`). it is missing the `O_NONBLOCK` flag (`04000` octal).

during systemd socket initialization in `main.c`, the daemon should explicitly set the `O_NONBLOCK` flag using `fcntl()` before registering the file descriptor with the event loop:
```c
int flags = fcntl(fd, F_GETFL, 0);
fcntl(fd, F_SETFL, flags | O_NONBLOCK);

trxvorr:
systemd passes sockets in blocking mode by default. pgagroal uses an asynchronous event loop (epoll/io_uring), calling accept() on a blocking socket can permanently deadlock the main thread if a client drops the connection just before the syscall.

explicitly applying the O_NONBLOCK flag via fcntl to all inherited systemd sockets during initialization fixes this 

fixes #742.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -905,6 +905,13 @@ read_superuser_path:
          }
          else if (sd_is_socket(fd, AF_INET, 0, -1) || sd_is_socket(fd, AF_INET6, 0, -1))
          {
+            int flags = fcntl(fd, F_GETFL, 0);
+
+            if (flags != -1)
+            {
+               fcntl(fd, F_SETFL, flags | O_NONBLOCK);
+            }
+
             main_fds[m] = fd;
             has_main_sockets = true;
             m++;


jesperpedersen:
Rebased, and merged.

jesperpedersen:
@trxvorr This likely need cross-port

src/main.c
Userfrom1995, line 904: The AF_UNIX branch also needs non-blocking mode?

Also rather than repeating the code for this one, we can make a function and calling it from here by passing FD you can name it something like `pgagroal_socket_nonblocking(fd)` and define it in `network.c`?

It would also be nice to log if setting non-blocking fails.

jesperpedersen: @Userfrom1995 Yes, stick it in `network.h` - we had one in the old days

trxvorr: @Userfrom1995 yes missed it, thanks for reviewing it

PR #751, opened 2026-02-25
[#749] Introduce pgagroal_socket_nonblocking function

Userfrom1995:
Add pgagroal_socket_nonblocking() to network.c with error handling and logging. Use it for both AF_UNIX and AF_INET/AF_INET6 branches in systemd socket activation, replacing inline fcntl code and adding the missing non-blocking call for Unix sockets.

### DIFF
--- a/src/include/network.h
+++ b/src/include/network.h
@@ -141,6 +141,14 @@ int
 pgagroal_socket_has_error(int fd);
 
 /**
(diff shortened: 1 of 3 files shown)


jesperpedersen:
Merged.

trxvorr:
@Userfrom1995 thanks for covering the refactor

PR #754, opened 2026-02-26
[#753] Prevent addr2line failure on paths with spaces and rem…

### LINKED ISSUE #753: Prevent addr2line failure on paths with spaces and remove popen()
Currently, pgagroal_backtrace in utils.c uses popen() to invoke addr2line during a crash. Because popen() relies on /bin/sh -c to parse the command string, it is vulnerable to shell metacharacter expansion and word splitting.

If pgagroal is executed from a filepath containing spaces (e.g., /opt/custom install/pgagroal), the shell will incorrectly split the -e argument, causing addr2line to fail silently and truncating the crash backtrace.

We should harden the backtrace utility by replacing popen() with a standard pipe(), fork(), and execvp() pattern.

This allows us to pass arguments in an array directly to the addr2line executable, entirely bypassing shell parsing and fixing the path-space issue.

We can map stdout of the child process back to the parent via pipe() and dup2() to maintain the exact same logging output.

trxvorr:
Removed popen() and pclose().

Implemented a fork() and execvp() pattern to pass arguments directly to the addr2line executable, entirely bypassing shell interpretation.

Mapped stdout of the child process back to the parent via pipe() and dup2() to maintain the exact same logging behavior.
fixes #753

### DIFF
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -47,6 +47,7 @@
 #include <sys/types.h>
 #include <sys/utsname.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <unistd.h>
 #ifdef HAVE_EXECINFO_H
 #include <execinfo.h>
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #758, opened 2026-02-26
[#757] Fix conf get/set crashes by initializing memory in child processes

### LINKED ISSUE #757: fix conf get/set crashes by initializing memory in child processes
when the daemon handles these management commands, it forks a child process. the child process was not calling pgagroal_memory_init() before attempting to build the JSON response payloads. this caused a silent segfault in the child, prematurely closing the socket

> @jesperpedersen @fluca1978 hi, i'm getting `read_byte: Invalid argument` errors on `conf get` commands, even on a fresh clone of upstream master. `ping` works fine, but `conf get` returns empty with no output.
> i've verified my setup is correct. while testing my code with `pgagroal-cli conf get failover_notify_script`, i found that other `conf get` commands (like `conf get host` and `conf get failover`) also don't work. i did a fresh clone to confirm, and the issue persists even without my changes. can you help.
> thanks. 

 _Originally posted by @bhaweshpanwar in [#686](https://github.com/pgagroal/pgagroal/issues/686#issuecomment-3968748437)_

trxvorr:
adds the necessary memory initialization and teardown in configuration.c to properly send the payload and exit

and adds a safety fallback in logging.c to print to stderr if the log_file is NULL, preventing null pointer de references during early startup or after a crash

fixes #757

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -32,6 +32,7 @@
 #include <configuration.h>
 #include <logging.h>
 #include <management.h>
+#include <memory.h>
 #include <network.h>
 #include <pipeline.h>
 #include <security.h>
@@ -6527,6 +6528,7 @@ pgagroal_conf_get(SSL* ssl __attribute__((unused)), int client_fd, uint8_t compr
    int total_seconds;
 
    pgagroal_start_logging();
+   pgagroal_memory_init();
 
    start_time = time(NULL);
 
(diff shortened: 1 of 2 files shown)


trxvorr:
@bhaweshpanwar i think this fixes the issue

jesperpedersen:
Merged.

jesperpedersen:
@trxvorr This likely need cross-porting

PR #759, opened 2026-02-27
[#752] Migrating SCRAM-SHA-256 to EVP_MAC

### LINKED ISSUE #752: Migration of SCRAM-SHA-256 to EVP_MAC
Objective:
This issue tracks the transition of the project's SCRAM-SHA-256 implementation, specifically the salted password generation and server/client signatures, to the modern EVP_MAC interface.

Reasoning: 
The currently used HMAC is deprecated in OPENSSL 3.x and may be removed from future versions

Success Criteria:
1. Migrated functions should produce same output (bitwise) of old implementation
2. There should be no unnecessary degredation in performance

Components I believe may be migrated:
Salting Logic (salted_password & salted_password_key)
Signature Logic (server_signature)
Client Proof (client_proof & verify_client_proof)

Useful resources:
https://github.com/pgagroal/pgagroal/pull/738 : The following PR contained discussion regarding SALT performance and migration to EVP
note: We are only concerned with openSSL 3.x
https://docs.openssl.org/3.0/man3/HMAC/#description : Currently Used HMAC functions documentation
https://docs.openssl.org/3.1/man3/EVP_MAC/ : EVP_MAC documentation

Zeyad-Daowd:
### Overview
This PR replaces the legacy HMAC_ functions with OpenSSL 3.0 EVP_MAC and EVP_Q_mac APIs. I added changes to five functions that still used the HMAC_ functions and will list the changes per function because they might differ.
Firstly, There are two choices to hash a message using evp
1. Use a similar manual flow like hmac_init_ex -> hmac_update -> hmac_final (but use the corresponding evp functions)
2. Use EVP_Q_mac which handles the full flow and does a single-shot hashing
image
reference: 
https://docs.openssl.org/3.1/man3/EVP_MAC/#context-manipulation-functions

### Modifications
**Function 1: salted_password**
I migrated the SCRAM flow to use the specified EVP_ functions (Approach 1)
since we do a lot of iterations of hashing I didn't want to use evp_q_mac which would create context per iteration adding a big overhead
I tested the new implementation versus the old optimized implementation using a standalone c file (calls the function 1000 times) and the results were:
the same bitwise results & a minor degradation in performance (my assumption is it is because of the overhead of evp_ apis)
image
I tried to investigate the new implementation using perf and found the following
after: 
image

before: 
image
In both cases the main cpu usage is the hashing logic in evp_mac_final or hmac_final

**Function 2: salted_password_key**
Since we are using only 1 iteration of hashing in this function I found that i could use either approach 1 (evp_q) or approach 2
(manual iteration)
I made another standalone c file that had the legacy hmac implementation, an implementation using evp_q and an implementation using a manual approach
The file called the functions 1000000 times to measure the performance
the results were in favor of the one-shot approach (evp_q) so that is why i migrated this function to evp_q_mac
image
Note: the bitwise resutls of the 3 functions are the same

**Functions 3-5: client_proof, verify_client_proof, server_signature:**
We are also using 1 iteration of hashing but there is a catch in those functions, those functions use multiple hmac_updates to concatenate the input before hashing and I found that if i wanted to use evp_q in them i would. have to allocate new memory and copy the data to the new location to use the one-shot evp_q
I made another file to benchmark legacy vs manual evp vs evp_q (with concatenation in another memory location)
image
Results were correct (bitwise) but were in favor of using the manual evp functions instead of evp_q
So i migrated those functions to the corresponding evp_ functions

Final Note: In the following functions:
client_proof, server_signature, salted_password_key, salted_password
I also added free(r); in the error block to avoid memory leaks because i found out that it wasn't freed in case of errors
closes #752 

Note on Hardware Acceleration & Providers:
I am currently investigating if OSSL_PARAM configurations can be used to use specific hardware based on CPU feature detection

### DIFF
--- a/src/libpgagroal/security.c
+++ b/src/libpgagroal/security.c
@@ -59,6 +59,8 @@
 #include <openssl/md5.h>
 #include <openssl/rand.h>
 #include <openssl/ssl.h>
+#include <openssl/params.h>
+#include <openssl/core_names.h>
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
@@ -3843,9 +3845,23 @@ client_proof(char* password, char* salt, int salt_length, int iterations,
    unsigned char* s_k = NULL;
    int s_k_length;
    unsigned char* c_s = NULL;
-   unsigned int length;
+   size_t length;
    unsigned char* r = NULL;
-   HMAC_CTX* ctx = HMAC_CTX_new();
+   EVP_MAC_CTX* ctx = NULL;
+   EVP_MAC* mac = NULL;
+   mac = EVP_MAC_fetch(NULL, "HMAC", NULL);
+   if (mac == NULL)
+   {
+      goto error;
+   }
+   ctx = EVP_MAC_CTX_new(mac);
+   OSSL_PARAM params[2];
+   params[0] = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST, "SHA256", 0);
+   params[1] = OSSL_PARAM_construct_end();
+   if (ctx == NULL)
+   {
+      goto error;
+   }
 
    if (salted_password(password, salt, salt_length, iterations, &s_p, &s_p_length))
    {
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@trxvorr Cc:

jesperpedersen:
@Zeyad-Daowd CI

Zeyad-Daowd:
@jesperpedersen Apologies for the failing checks on the CI earlier, I had an uninitialized pointer warning that i fixed.

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Zeyad-Daowd Please, cross-port

Zeyad-Daowd:
@jesperpedersen Happy to help! Glad to see it merged.
I'll begin working on the cross-port.

PR #770, opened 2026-03-01
[#769] Fix memory leak in extract_limit

### LINKED ISSUE #769: Fix memory leak in extract_limit during configuration parsing errors
A memory leak occurs in extract_limit when parsing a malformed configuration line. If the function successfully allocates memory for the database and user string but encounters an error late resulting in offset == -1, the execution jumps to the cleanup block. 

the cleanup block does not free the *database or *user pointers, causing them to leak since the parent function discards the result of the failed parse

trxvorr:
fixes #769

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -3167,6 +3167,20 @@ cleanup:
       db_part = NULL;
    }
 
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #786, opened 2026-03-08
Fix buffer overflow in pgagroal_md5 snprintf

trxvorr:


jesperpedersen:
`pgagroal_snprintf`

fheikens:
Heads-up on overlap with #896 (Remove md5, same author): #896 deletes the `pgagroal_md5` function this PR patches, along with the rest of the md5 auth path. If #896 is the intended direction, this snprintf fix is superseded by it and the two will conflict in `security.c`. Could you confirm which should land — i.e. is this a standalone interim fix (md5 retained for now), or should it close in favour of #896?

fluca1978:
I'm closing this in favor of #896 , we don't have it landed so far, hence we can prepare for the next version.
@jesperpedersen if you believe this is still worth, reopen.

jesperpedersen:
We are removing MD5 support

PR #797, opened 2026-03-19
Add tests for configuration parsing (boolean and byte size)

fheikens:
## Summary

Adds unit test coverage for configuration parameter parsing in
`test/testcases/test_configuration.c`.

Covered cases:
- boolean values:
  - accepted: `true`, `false`, `on`, `off`, `yes`, `no`, `1`, `0`
  - case-insensitive variants such as `TRUE`, `True`, `ON`
  - rejected invalid values such as `2`, `maybe`, `enabled`, `disabled`, empty string, `y`, `n`, `t`, `f`

- byte size values:
  - accepted: `256k`, `1M`, `1G`, `1024B`, `1024`
  - rejected invalid values such as `-1`, `abc`, `1T`

## Motivation

Configuration parameters control important runtime behavior. The parsing logic
accepts multiple representations but was not explicitly covered by tests.

This change ensures that accepted values remain stable and invalid values are
consistently rejected.

Follow-up: #804

## Test plan

- [x] Configuration module tests pass
- [x] Full unit test suite passes
- [x] No production code modified

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,3 +34,4 @@ Abdelrhman Sersawy <abdelrhmansersawy@gmail.com>
 Zeyad Daowd <zeyaddaowd@yahoo.com>
 Somye Mahajan <mahajan.somye@gmail.com>
 Bhawesh Panwar <panwarbhawesh1112@gmail.com>
+Frank Heikens <frank@elevarq.com>
--- a/test/testcases/test_configuration.c
+++ b/test/testcases/test_configuration.c
@@ -172,3 +172,88 @@ cleanup:
    free(str);
    MCTF_FINISH();
 }
+
+MCTF_TEST(test_configuration_accept_bool)
+{
+   // Lowercase
+   pgagroal_test_assert_conf_set_ok(CONFIGURATION_ARGUMENT_KEEP_ALIVE, "true");
+   pgagroal_test_assert_conf_set_ok(CONFIGURATION_ARGUMENT_KEEP_ALIVE, "false");
+   pgagroal_test_assert_conf_set_ok(CONFIGURATION_ARGUMENT_KEEP_ALIVE, "on");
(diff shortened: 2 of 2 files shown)


AUTHORS
```diff
@@ -34,3 +34,4 @@ Abdelrhman Sersawy <abdelrhmansersawy@gmail.com>
 Zeyad Daowd <zeyaddaowd@yahoo.com>
 Somye Mahajan <mahajan.somye@gmail.com>
 Bhawesh Panwar <panwarbhawesh1112@gmail.com>
+Frank Heikens <9465045+fheikens@users.noreply.github.com>
```
fluca1978: is the email really the one you want to use?

jesperpedersen:
@fluca1978 Yes, we should test the other sections as well

AUTHORS
jesperpedersen: Its a real email or nothing.

Copyright is still signed over to the project, and any voting is done by maintainers instead

fluca1978:
> @fluca1978 Yes, we should test the other sections as well

Ok, in any case this does not belong to this PR, since we need to test other properties as well, not only booleans, in server section.

Therefore, my opinion is that, once the email in `AUTHORS` have been solved, we could merge this.
@fheikens please fix the email, create also an issue for this case and link the commit to it.

AUTHORS
fheikens: Updated my email in AUTHORS. Thanks for pointing that out.

fheikens:
Updated my email in AUTHORS. Thanks for pointing that out.

fheikens:
Follow-up issue opened for broader configuration parsing coverage across other sections and property types: #804

fluca1978:
Merged, refencing a generic issue in the commit message.
Thans @fheikens for your work.

PR #802, opened 2026-03-20
[#781] Add FIPS status reporting

### LINKED ISSUE #781: FIPS
For PostgreSQL 18+ if pgacrypto is installed then we need to report its status.

If < 18 see if we can figure out a way to report it.

Default is `false`

pavanmanishd:
Closes #781

## What this does

Exposes whether a PostgreSQL backend has FIPS enabled via two output locations:

- `pgagroal-cli status details` → `FipsEnabled: true|false` per server
- Prometheus `/metrics` → `pgagroal_fips_enabled{server="<name>"} 0|1`

## Detection logic

FIPS is considered enabled when the server is **PostgreSQL >= 18** and the `pgcrypto` extension is installed. For older versions the field is always `false` and no query is sent to the server.

Detection runs at the end of `pgagroal_update_server_state()`, after the server version has already been populated from the startup parameters, by sending:

```sql
SELECT 1 FROM pg_extension WHERE extname = 'pgcrypto';
```

The response is parsed by reading the `RowDescription` (`T`) frame length and checking whether a `DataRow` (`D`) follows, this avoids hardcoded byte offsets and works across all PostgreSQL versions.

## Testing

Tested locally against PostgreSQL 14 (returns `0`, early exit before any query) and PostgreSQL 18 with `pgcrypto` installed (returns `1`, full query path exercised).

```
# PostgreSQL 18 + pgcrypto
$ curl -s http://localhost:2347/metrics | grep fips
pgagroal_fips_enabled{server="primary"} 1

$ pgagroal-cli status details | grep Fips
FipsEnabled: true
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,4 +34,5 @@ Abdelrhman Sersawy <abdelrhmansersawy@gmail.com>
 Zeyad Daowd <zeyaddaowd@yahoo.com>
 Somye Mahajan <mahajan.somye@gmail.com>
 Bhawesh Panwar <panwarbhawesh1112@gmail.com>
+pavanmanishd <pavanmanishd@gmail.com>
 Frank Heikens <frank@elevarq.com>
--- a/doc/PROMETHEUS.md
+++ b/doc/PROMETHEUS.md
@@ -48,6 +48,10 @@ The number of FATAL logging statements
 
 The number of failed servers
 
+**pgagroal_fips_enabled**
+
+Whether FIPS is enabled for the server (1 = enabled, 0 = disabled)
+
 **pgagroal_wait_time**
 
 The waiting time of clients
--- a/src/include/fips.h
+++ b/src/include/fips.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2026 The pgagroal community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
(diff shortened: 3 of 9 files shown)


pavanmanishd:
@jesperpedersen can you review this and why is this workflow not running?

jesperpedersen:
@pavanmanishd You need to support PostgreSQL 14+
@ahmedmmordi Cc:

You need to work together on this

pavanmanishd:
Updated to support PostgreSQL 14+.

**PostgreSQL 18+** — unchanged: calls `pgcrypto.fips_mode()` for the actual runtime status.

**PostgreSQL 14–17** — `fips_mode()` is not available on these versions. Detection now uses a two-step approach:
1. Confirm `pgcrypto` is installed (`SELECT 1 FROM pg_extension WHERE extname = 'pgcrypto'`). If absent, `fips_enabled` stays `false`.
2. Probe MD5 via pgcrypto (`SELECT digest('test'::bytea, 'md5')`). In FIPS mode, MD5 is a prohibited algorithm and pgcrypto raises an error — that error response sets `fips_enabled = true`. A successful DataRow means FIPS is not active.

**PostgreSQL < 14** — `fips_enabled` is set to `false`, no query is sent.


@jesperpedersen is this works around okay? i have tested it locally.

jesperpedersen:
@ahmedmmordi You need to coordinate your implementations such that we get the best one across the projects

ahmedmmordi:
Hi @pavanmanishd , I am working on FIPS in pgexporter. I had a good discussion with Jesper and received some useful code reviews that may help. Check them: [PR Link](https://github.com/pgexporter/pgexporter/pull/482) .

It would be great to align on an approach. I would be happy to hear your thoughts.

pavanmanishd:
Hey @ahmedmmordi, for PG14–17 in pgagroal, I used a PostgreSQL-side fallback (pgcrypto check + MD5 probe) because it reflects PostgreSQL runtime behavior directly and keeps detection DB-version scoped in fips.c, independent of host OS.

I just had a quick look at your PR, as the doc says OS-level FIPS detection (Linux/FreeBSD, and OpenBSD currently returns 0). What made you choose OS-level detection over PostgreSQL-side behavior, and how are you planning to keep it consistent across all supported OS targets? Also, did you find any workaround for OpenBSD as it was mentioned that system-level detection is not available for this?

ahmedmmordi:
> What made you choose OS-level detection over PostgreSQL-side behavior,

Because OS level detection is more reliable since FIPS is enforced outside Postgres at the OS level and Postgres just inherits this.

> I used a PostgreSQL-side fallback (pgcrypto check + MD5 probe)

I think Postgres might still execute it even if FIPS is enabled.

So we should try to make a hybrid approach. There are also some things that should be handled manually.

WDYT @pavanmanishd @jesperpedersen ?

fluca1978:
> > What made you choose OS-level detection over PostgreSQL-side behavior,
> 
> Because OS level detection is more reliable since FIPS is enforced outside Postgres at the OS level and Postgres just inherits this.
> 

More reliable meaning that PostgreSQL could report bad results? We are interested in knowing the status within PostgreSQL, right? So I guess we should trust what PostgreSQL tells us.
My 2 cents.

jesperpedersen:
@ahmedmmordi We don't have an extension for pgagroal, so we need to keep working on this to make this work - the implementation has to be the same across pgagroal, pgmoneta and pgexporter

ahmedmmordi:
@jesperpedersen @fluca1978 Great, that makes sense. I am still a bit unsure which approach to follow, so if Pavan’s approach is the best, I can align with it. I would appreciate your guidance.

fheikens:
Triage note: this conflicts with master at the moment. A rebase on latest master would clear it and let the FIPS work (#781) be reviewed — the checks were passing before the conflict.

PR #803, opened 2026-03-20
[#798] Reject invalid enum configuration values instead of silently using defaults

### LINKED ISSUE #798: Enum configuration parsers silently fall back to defaults on invalid input
## Summary

Several enum configuration parsers appear to fall back to a default value
when given unrecognized input, instead of warning or returning an error.

This means a typo like `pipeline = perforance` is silently treated as
`pipeline = auto` with no feedback to the operator.

## Affected parsers

| Parser | Default on invalid input | Location |
|---|---|---|
| as_pipeline() | PIPELINE_AUTO | configuration.c:2927 |
| as_validation() | VALIDATION_OFF | configuration.c:2906 |
| as_hugepage() | HUGEPAGE_OFF | configuration.c:2953 |
| as_logging_type() | TYPE_CONSOLE | configuration.c:2786 |
| as_logging_mode() | MODE_APPEND | configuration.c:2890 |

Additionally, `as_update_process_title()` returns an error code for invalid input,
but the caller sets `unknown = false`, which appears to swallow the error.

## Impact

Operators who misspell an enum value (e.g. `hugepage = yes` instead of `on`)
receive no error or warning, and the system silently falls back to a default
that may not match their intent.

## Observations

- `as_bool()` and `as_int()` propagate errors correctly via `unknown = true`
- `to_backend_type()` uses a sentinel value that is caught later with a warning
- enum parsers assign defaults directly without signaling or validation

## Question

Is this behavior intentional, or would it make sense to:
- return an error (like `as_bool`), or
- use a sentinel value and emit a warning during validation

## Version

2.1.0 (master at b31fe20)

fheikens:
## Summary

Several enum configuration parsers silently fell back to default values
when given unrecognized input. For example, `pipeline = perforance`
(typo) was silently treated as `pipeline = auto` with no feedback to
the operator.

This changes the affected parsers to signal an error on invalid input,
consistent with how `as_bool()` and `as_int()` already behave. The
existing configuration loading framework reports the invalid key/value
with file and line number context.

## Affected parameters

| Parameter | Valid values | Previous behavior on invalid input | New behavior |
|---|---|---|---|
| `pipeline` | auto, performance, session, transaction | Silent fallback to `auto` | Configuration error |
| `validation` | off, foreground, background | Silent fallback to `off` | Configuration error |
| `hugepage` | off, try, on | Silent fallback to `off` | Configuration error |
| `log_type` | console, file, syslog | Silent fallback to `console` | Configuration error |
| `log_mode` | a, append, c, create | Silent fallback to `append` | Configuration error |
| `update_process_title` | never, off, strict, minimal, verbose, full | Error returned but swallowed | Configuration error |

Omitted parameters and valid values (including case-insensitive
variants) behave exactly as before.

`as_logging_level()` is not changed as it has its own fallback logic
with explicit warnings for out-of-range debug levels.

## Tests

Adds 12 test functions covering acceptance and rejection for each
affected parameter.

## Test plan
- [x] All 16 configuration module tests pass
- [x] Full unit test suite passes (47/47)
- [x] No change to omitted-parameter defaults

Closes #798

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -65,14 +65,14 @@ static int extract_key_value(char* str, char** key, char** value);
 static int extract_syskey_value(char* str, char** key, char** value);
 static int as_int(char* str, int* i);
 static int as_bool(char* str, bool* b);
-static int as_logging_type(char* str);
+static int as_logging_type(char* str, int* type);
 static int as_logging_level(char* str);
-static int as_logging_mode(char* str);
+static int as_logging_mode(char* str, int* mode);
 
 static int as_logging_rotation_size(char* str, unsigned int* size);
-static int as_validation(char* str);
-static int as_pipeline(char* str);
-static int as_hugepage(char* str);
+static int as_validation(char* str, int* val);
+static int as_pipeline(char* str, int* pipeline);
+static int as_hugepage(char* str, unsigned char* hp);
 static unsigned int as_update_process_title(char* str, unsigned int* policy, unsigned int default_policy);
 static int extract_value(char* str, int offset, char** value);
 static void extract_hba(char* str, char** type, char** database, char** user, char** address, char** method);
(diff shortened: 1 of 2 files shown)


fluca1978:
Rebased and merged into 9aef1b36e5875122d70fda7168bcb304bc229b5c

fluca1978:
Thanks @fheikens

PR #805, opened 2026-03-20
[#772] Recognize [prometheus] as a global configuration section

### LINKED ISSUE #772: Configuration parser throws Unknown key warnings for valid [prometheus] section
When defining a `[prometheus]` block in `pgagroal.conf`, the configuration parser fails to recognize it as a valid global section. Instead, the parser falls back to treating it as a standard server definition block. Because Prometheus keys are not valid server keys, this causes the parser to spam the logs with "Unknown key" warnings during startup and configuration reloads.

fheikens:
## Summary

The configuration parser treated `[prometheus]` as a server section,
causing "Unknown key" warnings for valid prometheus keys during startup
and reload.

## Root cause

Two hardcoded lists of known global sections (in `pgagroal_read_configuration`
and `key_in_section`) included `pgagroal`, `pgagroal-vault`, and
`health_check` but not `prometheus`. A `[prometheus]` section was
therefore parsed as a server definition, and its keys failed the
server-key lookup.

## Fix

Add `"prometheus"` to both global section checks, matching the existing
pattern for `"health_check"`. The fix covers both startup and reload
paths since both use `pgagroal_read_configuration`.

## Tests

Adds a regression test verifying that `metrics`, `metrics_cache_max_age`,
and `metrics_cache_max_size` are accepted in a `[prometheus]` section.
Also verifies the same keys still work in the main `[pgagroal]` section.

## Test plan
- [x] Configuration module tests pass (5/5)
- [x] All unit tests pass; pre-existing integration test failures unchanged
- [x] Regression test fails before fix, passes after

Closes #772

### DIFF
--- a/test/testcases/test_configuration.c
+++ b/test/testcases/test_configuration.c
@@ -375,3 +375,17 @@ MCTF_TEST(test_configuration_reject_invalid_update_process_title)
 
    MCTF_FINISH();
 }
+
+MCTF_TEST(test_configuration_prometheus_section_keys)
+{
+   // Metrics keys should be accepted in a [prometheus] section
+   pgagroal_test_assert_conf_section_set_ok("prometheus", CONFIGURATION_ARGUMENT_METRICS, "9187");
+   pgagroal_test_assert_conf_section_set_ok("prometheus", CONFIGURATION_ARGUMENT_METRICS_CACHE_MAX_AGE, "10s");
+   pgagroal_test_assert_conf_section_set_ok("prometheus", CONFIGURATION_ARGUMENT_METRICS_CACHE_MAX_SIZE, "256K");
+
+   // Same keys should still work in the main [pgagroal] section
+   pgagroal_test_assert_conf_set_ok(CONFIGURATION_ARGUMENT_METRICS, "9187");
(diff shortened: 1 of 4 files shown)


test/testcases/test_configuration.c
```diff
@@ -172,3 +172,31 @@ MCTF_TEST(test_configuration_time_format_output)
    free(str);
    MCTF_FINISH();
 }
+
+static void
+assert_conf_section_set_ok(char* section, char* key, char* value)
```
fluca1978: Is there a particular reason we don't expose it as `pgagroal_assert_conf_section_set_ok`?

fheikens: Thanks, addressed.

Rebased the branch and moved the helper into the shared test infrastructure so we can reuse it instead of keeping local copies in individual tests.

Exposed it as:
`pgagroal_test_assert_conf_section_set_ok(...)`

I intentionally kept the `test` prefix to make it clear this is test-only functionality, not part of the production API.

The prometheus regression test now uses the shared helper. All configuration tests are passing again (21/21).

fluca1978:
Merged, thanks @fheikens

PR #807, opened 2026-03-21
[#419] Add -k/--master-key flag to set a custom master key file path

### LINKED ISSUE #419: Allow `pgagroal-admin` to specify where to store the master key password file

pavanmanishd:
Closes #419

## Problem

`pgagroal-admin master-key` always writes the master key to `$HOME/.pgagroal/master.key` with no way to override the path. A previous attempt (PR #430) was rejected for being a superficial find-and-replace.

## Fix

Add `-k / --master-key FILE` to `pgagroal-admin`. When supplied, the key is written to that path; when omitted, the default path is used unchanged.

The parent directory of the custom path is derived and permission-checked (must be accessible). The default `~/.pgagroal` directory creation and permission checks are only executed when no custom path is provided.

## Test

```bash
# Custom path
pgagroal-admin -k /tmp/my.key master-key
# → Master Key stored into /tmp/my.key

# Default path unchanged
pgagroal-admin master-key
# → Master Key stored into ~/.pgagroal/master.key
```

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -53,9 +53,9 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
-static int master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_format);
-static int add_user(char* users_path, char* username, char* password, bool generate_pwd, int pwd_length, int32_t output_format);
-static int update_user(char* users_path, char* username, char* password, bool generate_pwd, int pwd_length, int32_t output_format);
+static int master_key(char* password, char* key_path, bool generate_pwd, int pwd_length, int32_t output_format);
+static int add_user(char* users_path, char* key_path, char* username, char* password, bool generate_pwd, int pwd_length, int32_t output_format);
+static int update_user(char* users_path, char* key_path, char* username, char* password, bool generate_pwd, int pwd_length, int32_t output_format);
 static int remove_user(char* users_path, char* username, int32_t output_format);
 static int list_users(char* users_path, int32_t output_format);
 static int create_response(char* users_path, struct json* json, struct json** response);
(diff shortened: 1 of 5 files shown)


src/admin.c
```diff
@@ -305,10 +312,11 @@ main(int argc, char** argv)
 }
 
 static int
-master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_format)
+master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_format, char* key_path)
```
fluca1978: `key_path` should become first or second argument

fluca1978 (changes requested):
@pavanmanishd one problem we have here is that we store the master key somewhere else and then we are no more able to get it:

``
% pgagroal-admin master-key -k /tmp/luca.key
Master key (will not echo): 
Header:
  ClientVersion: 2.1.0
  Command: 19
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20260321152239
Outcome:
  Status: true
  Time: 00:00:03
Master Key stored into /tmp/luca.key
luca@franzina ~/git/private/pgagroal/build
 % pgagroal-admin user add -k /tmp/luca.key  
pgagroal-admin: Invalid master key
Header:
  ClientVersion: 2.1.0
  Command: 20
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20260321152256
Outcome:
  Error: 1
  Status: false
pgagroal-admin: Error for <user add>

```



So I think the first step is to generalize the master key into the in-memory 
configuration, so that both `pgagroal-admin` and `pgagroal` are able to read it.

Have you tried it?

pavanmanishd:
@fluca1978 Thanks for the suggestions, made the required changes can you PTAL

src/admin.c
pavanmanishd: fixed

fluca1978:
We are still at a single side: `pgagroal-admin` can now work with `-k`, but `pgagroal` does not.

We need a way to inform `pgagroal` and all other pieces that requires the master key file location to know it.

fheikens:
Triage note: this has been quiet since March and currently both conflicts with master and has changes requested. Are you still planning to take it forward? If so, a rebase plus addressing the review comments would unblock it; otherwise we can close and revisit #419 separately. Just let us know.

PR #816, opened 2026-03-24
Fix incorrect documentation entries

fheikens:
This fixes three factual documentation issues:

- correct `max_connections` default and max value in the man page
- replace non-existent `stop` CLI command with `shutdown` in getting started docs
- replace obsolete `libev` parameter with `ev_backend` in the manual

All changes were verified against the current codebase.

PR #817, opened 2026-03-24
Fix TLS documentation config path

fheikens:
This fixes an incorrect configuration file reference in the TLS documentation.

The TLS parameters belong in `pgagroal.conf`, not `pgagroal_vault.conf`.

Verified against current configuration handling and existing documentation.

### DIFF
--- a/doc/manual/en/08-tls.md
+++ b/doc/manual/en/08-tls.md
@@ -30,7 +30,7 @@ For the purpose of this tutorial we will assume the client certificate and key s
 
 **Modify the `pgagroal` configuration**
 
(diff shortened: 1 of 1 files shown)


fheikens:
The cancelled check appears unrelated to this change.

`build-linux (clang, Debug)` hit the 30-minute workflow timeout during `Install Dependencies` while installing the docs/LaTeX toolchain, and did not reach build or test. The PR itself is a 1-line documentation fix, and the same job passed on #816 the same day.

All other checks passed, so this looks like CI infrastructure/mirror slowness rather than a problem with the branch.

fluca1978:
Merged with reference to a specific ticket.
Thanks @fheikens

PR #826, opened 2026-03-31
[#710] Detect duplicated and inconsistent servers

### LINKED ISSUE #710: Detect duplicated and inconsistent servers
While testing #693 I discovered that `pgagroal` is unable to detect the same server configured over and over.

Example configuration:

```
% head -n 20 /etc/pgagroal/pgagroal.conf

[rachel]
host = 127.0.0.1
port = 5432
primary = on
tls = on
tls_ca_file = /certs.pgagroal/postgres.crt

[ross]
host = localhost
port = 5432
primary = on
tls = on
tls_ca_file = /certs.pgagroal/postgres.crt

```

The `rachel` and `ross` servers are the same, with the same configuration, and both are set to be `primary` (this can be also one primary on e one off, it does not change).
The ending result is:

```
% pgagroal-cli status        
Header:
  ClientVersion: 2.1.0
  Command: status
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260211094155
Outcome:
  Status: true
  Time: 00:00:00
Request: {}
Response:
  ActiveConnections: 0
  MaxConnections: 100
  NumberOfServers: 2
  ServerVersion: 2.1.0
  Servers:
    rachel:
      Host: 127.0.0.1
      Port: 5432
      Status: Running
    ross:
      Host: localhost
      Port: 5432
      Status: Running
  Status: Running
  TotalConnections: 35
```

**both servers are reported up and running** (this is against commit ed143476daeaad880d465b8a121db9bea390df5e but the same happens on `HEAD`:

```
% pgagroal-cli status
Header:
  ClientVersion: 2.0.0
  Command: status
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260211094527
Outcome:
  Status: true
  Time: 00:00:00
Request: {}
Response:
  ActiveConnections: 0
  MaxConnections: 100
  NumberOfServers: 2
  ServerVersion: 2.0.0
  Servers:
    - Host: 127.0.0.1
      Port: 5432
      Server: rachel
      State: Primary
    - Host: localhost
      Port: 5432
      Server: ross
      State: Not init (primary)
  Status: Running
```

We need some way to detect that the system is connecting to the same server (it could be worth using the server identifier).

Abdelrhmansersawy:
resolved #710 

## Changes

- Enhanced `is_same_server()` to resolve hostnames using `getaddrinfo()`, allowing detection of duplicates such as `localhost` vs `127.0.0.1`.
- Add `pgagroal_check_server_identifiers()` to detect servers pointing to the same PostgreSQL cluster by comparing `system_identifier` via `pg_control_system()` at startup.
- Refactored `server_probe()` in `health.c` to use the shared `pgagroal_server_query_execute()` in `server.c`, eliminating duplicated connect-and-query logic.
- Added test cases that test (Multiple primary servers, Duplicate server section names, System identifier checks, Valid configuration acceptance)

### DIFF
--- a/src/libpgagroal/server.c
+++ b/src/libpgagroal/server.c
@@ -34,6 +34,8 @@
 #include <pool.h>
 #include <security.h>
 #include <server.h>
+#include <message.h>
+#include <network.h>
 #include <utils.h>
 #include <value.h>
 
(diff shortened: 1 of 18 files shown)


Abdelrhmansersawy:
@fluca1978 
Side note: While working on the implementation of `query_system_identifier`, I noticed duplicated code between this and `server_probe()`. Both perform the same logic but use different SQL queries. I think we can create a utility function, `pgagroal_server_query_execute`, that takes an SQL query as a parameter and returns the response, while each caller handles its own logic.

WDYT?

Since this is critical part, I will validate it in a real environment tomorrow to verify more that we correctly handle servers pointing to the same cluster on different ports.

fluca1978:
Looks fine to me, even if the refactoring of the generic query mechanism should have an issue by its own, so that you can reference it in here too.
The things I note are:
- we are tying the identification mechanism to the health check configuration parameters, and I'm not so sure about this. While having credentials sounds fine, we should clearly document this tying and allow the user to disable health check if she only want identifiers checking *EDIT* we already have `health_check = off` for this purpose, so this is fine.
- the query function should be more general and accept credentials, so that we can use it in the future with different configuration

src/libpgagroal/server.c
```diff
@@ -45,9 +46,369 @@
...
+   pgagroal_snprintf(buffer + offset, sizeof(buffer) - offset, "%s", config->health_check_user);
+   offset += strlen(config->health_check_user) + 1;
+
+   buffer[offset++] = 0;
+
+   if (pgagroal_write_socket(NULL, fd, buffer, offset) != offset)
+   {
+      pgagroal_log_debug("server_query: Failed to write startup packet");
+      goto error;
+   }
```
Abdelrhmansersawy: @fluca1978 
As part of refactoring `pgagroal_server_query_execute`, I noticed we can use `pgagroal_create_startup_message` which already does the same thing - this avoids repeated code, gets rid of the fixed `buffer[1024]` since it allocates dynamically (avoid overflow), and it also sets `application_name = pgagroal` in the startup packet which makes these connections identifiable in `pg_stat_activity`.

fluca1978: > @fluca1978 As part of refactoring `pgagroal_server_query_execute`, I noticed we can use `pgagroal_create_startup_message` which already does the same thing - this avoids repeated code, gets rid of the fixed `buffer[1024]` since it allocates dynamically (avoid overflow), and it also sets `application_name = pgagroal` in the startup packet which makes these connections identifiable in `pg_stat_activity`.

The more we can generalize and make clean, the better!

Abdelrhmansersawy:
I have addressed the comments, PTAL

fluca1978:
@Abdelrhmansersawy  it looks good to me, but I still have some semantic doubts:
- if the `health_user` is something that does not exists, the test for the same identifier fails, hence `pgagroal` starts regularly. Should we make mandatory the test on the identifier?
- the `health_user` requires the setup of a database/credentials used to check for duplicated servers. While setting up user/database is fine for an extra feature, like healt check, I don't think is fine to just start up `pgagroal` in a safe way. And this is the contrary of making this test mandatory.

What I think, either:
- if we can collect at least one identifier, than the test must be mandatory for all the servers. On the other hand, if we cannot collect any identifier at all, we emit a warning to inform the user that pgagroal is not able to check for duplicates;
- we add a `strict = on` mode (that could be used for other purposes too) that makes the above test on identifiers mandatory and refuses to start if even one identifier cannot be collected.

@jesperpedersen what do you think? Personally, I prefer using `strict = on` to tell pgagroal to do its best in checking for non-duplicates.

Abdelrhmansersawy:
I don’t think if `health_user` is missing on any server, we should still run server identifier checks on others, as this leads to partial consistency handling. Either query the server identifier on all servers or skip it entirely.

I see `strict = on` is a good idea, and we should document that `health_user` is required for all configured servers.

WDYT

jesperpedersen:
My thoughts are that we should add the `health_user` template in our sample configuration. If the `health_user` isn't configured then we point that out in an `INFO` upon startup.

Setting up a user/database for this basically takes a connection from the backend and require extra DBA resources.

However, we need to highlight that it is a best-practice to enable

Abdelrhmansersawy:
From what I understood, we should:
1. Document that it’s best practice to enable `health_user` on all configured servers.
2. On pgagroal startup, query the server identifier for all servers that have `health_user` - if not, log `INFO`.
3. Detect server duplication through the identifier.

My questions:
1. If `health_user` is missing on some servers, should we still run server identifier duplication checks on the others?
2. With this approach, we enforce server identification - what if the user doesn’t need this for any reason? or this scenarios shouldn't exist

jesperpedersen:
Maybe we need more policies to control this area...

Abdelrhmansersawy:
What if we introduce a dedicated config parameter just for startup validation, instead of tying it to the health check?

Something like:
`server_validation = strict | optional | off` (default = `optional`)

- `strict`: fail startup if identifiers can’t be fetched or if duplicates are detected
- `optional`: try to run the check if `health_check_user` exists; otherwise just log an INFO and continue
- `off`: skip identifier checks completely

This keeps it simple, but also gives DBAs clear control over how strict they want startup validation to be. What do you think?

jesperpedersen:
@Abdelrhmansersawy Yes, we can do `on`, `try` and `off`

Abdelrhmansersawy:
@fluca1978 
A few follow-up:
1. I have added a new config parameter `startup_validation = on | try | off` (that could be used for other startup validation purposes too)
2. Document that's best practice to config `heath_check_user` to all server.

fluca1978:
> @fluca1978 A few follow-up:
> 
>     1. I have added a new config parameter `startup_validation = on | try | off` (that could be used for other startup validation purposes too)
> 
>     2. Document that's best practice to config `heath_check_user` to all server.

Great.
However, there is a default configuration problem:

```
% grep startup_val /etc/pgagroal/pgagroal.conf
#startup_validation = on
```

so it should be set to `try`, right?

```
% pgagroal-cli conf get startup_validation
Error
```

It works fine when the value is explicitly set.

Another thing I noted, that I believe should be managed as a separate issue, is that if we statt with one server and `startup_validation = on`, then add another server and `pgagroal-cli conf reload` the validation is not run again. This is ok (should we stop `pgagroal`? I don't think so), but we should consider when `conf reload` what can happen changing either `startup_validation` or the servers. Again, this is probably another issue with a different workflow to detect runtime (not startup) inconsistencies.

Abdelrhmansersawy:
> Great. However, there is a default configuration problem:
> 
> ```
> % grep startup_val /etc/pgagroal/pgagroal.conf
> #startup_validation = on
> ```
> 
> so it should be set to `try`, right?

it's already set to `try` by default, I tried it and it's works. really I don't know what's the problem


> ```
> % pgagroal-cli conf get startup_validation
> Error
> ```
> 
> It works fine when the value is explicitly set.

I have added `startup_validation` to pgagroal_cli

```
pgagroal-cli conf get startup_validation
1
```

But the question is: I followed the same pattern as `validation`. Should we print `on | try | off` instead of `2 | 1 | 0`, since users won’t know what these numbers represent? (as the same of `validation` be `off | foreground |background`)

 
> Another thing I noted, that I believe should be managed as a separate issue, is that if we statt with one server and `startup_validation = on`, then add another server and `pgagroal-cli conf reload` the validation is not run again. This is ok (should we stop `pgagroal`? I don't think so), but we should consider when `conf reload` what can happen changing either `startup_validation` or the servers. Again, this is probably another issue with a different workflow to detect runtime (not startup) inconsistencies.

Once we done from this PR, I follow-up open a separate issue for this - discuss what's best to deal with this.

fluca1978:
> But the question is: I followed the same pattern as `validation`. Should we print `on | try | off` instead of `2 | 1 | 0`, since users won’t know what these numbers represent? (as the same of `validation` be `off | foreground |background`)


It would be better, and I guess we should also start differentiating the cli UI from the JSON one, so print `on | try | off` in the cli and print **also** `0 | 1 | 2` in the JSON one. This is not mandatory, so at least print a stringigified version in both.  We could start a refactoring/improvement of the JSON output as a separate and more organized PR.

> 
> > Another thing I noted, that I believe should be managed as a separate issue, is that if we statt with one server and `startup_validation = on`, then add another server and `pgagroal-cli conf reload` the validation is not run again. This is ok (should we stop `pgagroal`? I don't think so), but we should consider when `conf reload` what can happen changing either `startup_validation` or the servers. Again, this is probably another issue with a different workflow to detect runtime (not startup) inconsistencies.
> 
> Once we done from this PR, I follow-up open a separate issue for this - discuss what's best to deal with this.

Great!

Abdelrhmansersawy:
Once we have done from this PR, I start working on follow-up opened issue as this PR is blocked them

Abdelrhmansersawy:
@fluca1978 
as follow-up, I have updated `startup_validation` to use new `pgagroal_json_put_enum_value`
```
% pgagroal-cli  conf get startup_validation
try
```
Let me know if there is anything I missed to be able to merge this PR.

fluca1978:
merged, with a commit message.
Thanks @Abdelrhmansersawy 

Please, in the future, use the commit message to indicate the changes, so that it documents the PR.

PR #827, opened 2026-03-31
Add .dockerignore for container builds

fheikens:
The Dockerfiles in contrib/docker/ are built with the repository root as
context (e.g. docker build -f ./contrib/docker/Dockerfile.alpine .). Without
a .dockerignore, the full repository is sent to the build context, including
files not required for the cmake build or runtime image.

Add a .dockerignore to exclude version control, CI configuration, tests,
monitoring tooling, developer utilities, packaging files, and build artifacts.

All required paths remain included:

- src/, cmake/, CMakeLists.txt (cmake build)
- doc/ (man page generation)
- contrib/docker/pgagroal.conf, pgagroal_hba.conf (runtime configuration)

Excluding test/ is safe, as cmake already guards for its presence.

No change in build output or runtime behavior. Compatible with Docker and Podman.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,42 @@
+# Reduce Docker build context for container image builds.
+# The Dockerfiles in contrib/docker/ use COPY . . in the builder stage,
+# so only files needed by cmake and the runtime COPY are required.
+
+# Version control
+.git
+.gitignore
+
+# CI/CD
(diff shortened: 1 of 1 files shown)


fluca1978:
Merged with issue reference and commit message.
Thanks @fheikens

Issue #809, opened 2026-03-21
Improve test suite


fluca1978:
Improve the test suite to check for configuration parsing errors.

fluca1978:
@fheikens if this work is done, please comment so we can close the issue.
Otherwise, please reference this issue in subsequent PRs.

fheikens:
Covered now. The test suite checks configuration parsing/validation errors in:

- `test/testcases/test_configuration.c` — rejects invalid values across parameter types (time, bool, pipeline, bytes, validation, hugepage, `log_type`/`log_mode`, int, log_rotation, …) alongside the accept cases.
- `test/testcases/test_startup_validation.c` — config-read failure paths (duplicate server names, multiple primaries, …).

That covers the configuration-parsing-error checks this issue asked for, so closing as done.

PR #828, opened 2026-03-31
Add HEALTHCHECK to container Dockerfiles

fheikens:
The container Dockerfiles in contrib/docker/ have no HEALTHCHECK instruction.
This means Docker and Podman cannot report whether the pgagroal daemon inside
the container is actually running.

Add a HEALTHCHECK using pgagroal-cli ping, which checks daemon liveness via
the Unix socket. The check runs as the pgagroal user with the container's own
configuration file.

Parameters: interval 10s, timeout 5s, start-period 5s, retries 3.

pgagroal-cli ping checks the daemon process only, not backend connectivity.
This is intentional: the container is healthy when its daemon is running,
regardless of backend state.

No change to runtime behavior. Both Alpine and Rocky 9 variants are updated
identically.

### DIFF
--- a/contrib/docker/Dockerfile.alpine
+++ b/contrib/docker/Dockerfile.alpine
@@ -95,4 +95,7 @@ EXPOSE 2345 2346
 
 USER pgagroal
 
+HEALTHCHECK --interval=10s --timeout=5s --start-period=5s --retries=3 \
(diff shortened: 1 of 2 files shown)


fluca1978:
merged, referencing the catch-all #809.

Thanks @fheikens

PR #829, opened 2026-03-31
Remove development packages from container runtime images

fheikens:
The runtime stages in both container Dockerfiles install packages that are only
needed at compile time. The pgagroal binaries are already compiled in the builder
stage and copied to the runtime image, so development headers, sanitizer
libraries, and unused dependencies serve no purpose at runtime.

Rocky 9: remove libpq-devel, zlib-devel, libzstd-devel, lz4-devel, bzip2-devel,
libyaml-devel, systemd-devel, libasan, libubsan, and a duplicate libatomic entry.
Replace liburing-devel with liburing (runtime library only).

Alpine: remove yaml, bzip2-dev, linux-headers. Replace liburing-dev with liburing.

All runtime library dependencies remain satisfied:
- libpq, zlib, zstd, lz4, bzip2, liburing, libatomic (explicit)
- openssl-libs, systemd-libs, glibc/musl, pthread (base image)

Builder stages are unchanged. No change to runtime behavior.

### DIFF
--- a/contrib/docker/Dockerfile.alpine
+++ b/contrib/docker/Dockerfile.alpine
@@ -69,14 +69,11 @@ RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositori
     apk add --no-cache \
     bash \
     ca-certificates \
-    yaml \
     bzip2 \
-    bzip2-dev \
     libpq \
     libatomic \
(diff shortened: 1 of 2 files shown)


fluca1978:
merged, with issue link and commit message.
Thanks @fheikens 

Please, for the future, add an issue and reference it when doing changes.

PR #815, opened 2026-03-24
[#813] Add pool saturation test configuration and test case

fheikens:
## Summary

Adds a test configuration and test case that exercises pool acquisition
under saturation — when more clients connect than `max_connections` allows.

This is the first test that explicitly exercises `blocking_timeout` and the
retry path in `pgagroal_get_connection`.

## Changes

**test/conf/07/pgagroal.conf** — new configuration with:
- `max_connections = 2`
- `blocking_timeout = 10s`
- `pipeline = session`

**test/testcases/test_connection.c** — new test:
- `test_pgagroal_connection_saturation`: runs 4 concurrent pgbench clients
  (select-only, 100 transactions each) against the 2-connection pool
- Validates that all clients complete successfully within the blocking
  timeout, confirming that the retry mechanism works under saturation

## How it works

The existing `run_multiple_config_tests` in `check.sh` automatically picks
up `test/conf/07/` and runs the full test suite against it. With only 2
pool slots and 4+ concurrent clients, the retry path is exercised on every
test run.

## Test plan
- [x] Compiles cleanly
- [x] Existing unit tests unaffected (25/25 configuration tests pass)
- [x] No production code modified

Ref #813

Issue #813, opened 2026-03-22
Revisit hardcoded 500ms retry delay in pool acquisition under blocking_timeout


fheikens:
## Summary

When `blocking_timeout` is configured and the pool cannot satisfy a connection request, `pgagroal_get_connection()` retries in a polling loop with a hardcoded 500ms sleep between attempts. This may introduce unnecessary latency under brief pool saturation.

## Current behavior

In `src/libpgagroal/pool.c`, the retry path at label `retry2` sleeps for a fixed 500ms before re-scanning the pool:

```c
retry2:
   if (pgagroal_time_is_valid(config->blocking_timeout))
   {
      /* Sleep for 500ms */
      SLEEP(500000000L)
      ...
      goto start;
   }
```

Each retry re-increments `active_connections`, re-scans all slots, and sleeps again if no connection is available.

## History

- Commit `96361f3` (2019-09-16, @jesperpedersen): introduced the `blocking_timeout` feature with `sleep(1)` — a full 1-second delay.
- Commit `f39c906` (2019-10-02, @jesperpedersen): changed to 500ms nanosleep as part of a broader refactoring ("number_of_connections → active_connections"). The commit message does not mention the sleep change.

I could not find a documented rationale for either the original 1-second value or the subsequent 500ms value.

## Observation

There is no wakeup mechanism between `pgagroal_return_connection()` and `pgagroal_get_connection()`. The retry model is pure polling, which makes the sleep interval a direct tradeoff between CPU usage and connection acquisition latency.

With the default `blocking_timeout = 30s`, the current behavior allows a maximum of 60 retries. If a connection frees 1ms after a sleep starts, the requester waits the remaining 499ms unnecessarily. Under brief saturation, the median unnecessary wait is ~250ms.

For comparison, the transaction pipeline retry path (line 371) uses `SLEEP_AND_GOTO(1000L, start)` — a 1μs delay, which is several orders of magnitude shorter.

## Why this issue first

I wanted to check whether the 500ms value is intentional or simply a historical default that was never revisited. If there is a specific reason for this value (e.g., avoiding excessive pool scanning under sustained saturation, fairness between waiting processes), that would be useful to document.

## Possible improvement

If the value is not load-bearing, a bounded exponential backoff could reduce latency under brief saturation while remaining CPU-friendly under sustained saturation:

```
1ms → 2ms → 4ms → 8ms → 16ms → 32ms → 64ms → 100ms (cap)
```

[...]

PR #832, opened 2026-04-01
[#813] Add pool saturation test configuration and test case

fheikens:
## Summary

Adds the test configuration and test case for pool saturation that was
described in #815 but not included in the merged diff.

**Config files:**
- **test/conf/07/pgagroal.conf** — session pipeline with
  `max_connections = 6` and `blocking_timeout = 10`
- **test/conf/07/pgagroal_databases.conf** — limit overrides that
  reduce the total from 8 to 6, matching the lower `max_connections`
- **test/conf/07/pgagroal_hba.conf** — standard trust-all HBA

**Test:**
- `test_pgagroal_connection_saturation`: 8 pgbench clients (1 thread
  each, select-only, 100 transactions)

Against config 07, the first 6 clients connect immediately, the
remaining 2 enter the `blocking_timeout` retry loop and acquire
connections as earlier clients finish and disconnect.

### Skip condition

The test skips when `max_connections >= 8` (the client count).
Saturation only occurs when the pool is smaller than the number of
clients. Configs 01–06 have `max_connections` of 100–200 (capped to 8
in Debug builds), so the test only runs against config 07
(`max_connections = 6`).

### Databases config override

The CI base `pgagroal_databases.conf` defines limits totaling 8
connections (6 + 2). With `max_connections = 6`, pgagroal rejects this
as invalid (`total_connections > max_connections`). The override in
`test/conf/07/` reduces the limits to 4 + 2 = 6, matching
`max_connections`. The existing `run_multiple_config_tests`
infrastructure (merged in #815) already supports per-config
`pgagroal_databases.conf` overrides.

## What this does NOT change

- No production code modified
- No retry algorithm changes
- No new configuration parameters

This is a test-only baseline for #813.

## Test plan
- [x] Compiles cleanly
- [x] Config 07 limit totals (6) <= max_connections (6)
- [x] Existing tests unaffected — `test_pgagroal_connection_load`
      (6 clients) fits in max_connections = 6
- [x] Skips on configs 01–06 (max_connections >= 8)
- [x] No production source files modified
- [ ] CI validation (pending)

Ref #813

### DIFF
--- a/test/conf/07/pgagroal.conf
+++ b/test/conf/07/pgagroal.conf
@@ -1,0 +1,19 @@
+[pgagroal]
+host = localhost
+port = 2345
+
+log_type = file
+log_level = debug5
+log_path = test.log
+
+max_connections = 6
+blocking_timeout = 10
+idle_timeout = 600
+validation = off
+unix_socket_dir = /tmp/
+pipeline = session
+ev_backend = epoll
+
+[primary]
+host = localhost
+port = 5432
--- a/test/conf/07/pgagroal_databases.conf
+++ b/test/conf/07/pgagroal_databases.conf
@@ -1,0 +1,4 @@
+#
+# DATABASE=ALIAS1,ALIAS2 USER MAX_SIZE INITIAL_SIZE MIN_SIZE
+#
+mydb=pgalias1,pgalias2 myuser 6 6 1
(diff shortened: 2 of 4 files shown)


fheikens:
Closing this PR. The saturation test consistently exposes a retry/acquisition issue in `pgagroal_get_connection()` that prevents blocked clients from picking up freed pool slots under concurrent contention — the same problem described in #813.

Posting the full findings as a comment on #813 instead, since the test belongs after the retry mechanism is improved, not before.

Issue #834, opened 2026-04-02
Segmentation fault


ankushT369:
**Describe the bug**
```sh
[ankush@cognitive src]$  ./pgagroal-cli -h localhost -p 2345 status details
User name: ankush
Password : 
Segmentation fault (core dumped)
```

After searching out I found: 
```sasl_continue->length``` is less that 9
```c
get_scram_attribute('r', (char*)(sasl_continue->data + 9), sasl_continue->length - 9, &combined_nounce);
```

jesperpedersen:
@ankushT369 Can you reproduce that ?

ankushT369:
> [@ankushT369](https://github.com/ankushT369) Can you reproduce that ?

on running this  
```
./pgagroal-cli -h localhost -p 2345 status details 
```
then entering username then password then hitting enter it shows segfault

ankushT369:
@jesperpedersen 

After searching out I found:
```sasl_continue->length``` is less that 9

```
                                                      this part becomes negative
                                                          |------------------------|
get_scram_attribute('r', (char*)(sasl_continue->data + 9), sasl_continue->length - 9, &combined_nounce);
```

which endup calling : 
```c
static int
get_scram_attribute(char attribute, char* input, size_t size, char** value)
{
   char* dup = NULL;
   char* result = NULL;
   char* ptr = NULL;
   size_t token_size;
   char match[2];

   match[0] = attribute;
   match[1] = '=';
   // Since 'size' is of type size_t, a negative value becomes a very large positive number,
// which can lead to calling calloc() with an excessively large memory size.
   dup = (char*)calloc(1, size + 1);
   memcpy(dup, input, size);

```

jesperpedersen:
@ankushT369 Yes, but that means a wrong message - size, format, something... We start from there.

Was the entire message received ?

ankushT369:
> [@ankushT369](https://github.com/ankushT369) Yes, but that means a wrong message - size, format, something... We start from there.
> 
> Was the entire message received ?

heres the thing : 
adding    pgagroal_log_message(msg); in the code i see this its not complete I have to inspect it more!!

```sh
[ankush@cognitive src]$ ./pgagroal-cli -h localhost -p 2345 status
User name: ankush
Password : 
00 ?
Segmentation fault (core dumped)
```

jesperpedersen:
@ankushT369 So, we need to keep loading until the message is complete

fheikens:
[...]

PR #835, opened 2026-04-02
[#834] fixed: segmentation fault due to unhandled sasl_continue length

ankushT369:
#834 fixed

### DIFF
--- a/src/libpgagroal/security.c
+++ b/src/libpgagroal/security.c
@@ -1051,6 +1051,11 @@ pgagroal_remote_management_scram_sha256(char* username, char* password, int serv
 
    sasl_continue = pgagroal_copy_message(msg);
 
+   if (sasl_continue->length <= 9)
+   {
+      goto error;
+   }
+
    get_scram_attribute('r', (char*)(sasl_continue->data + 9), sasl_continue->length - 9, &combined_nounce);
    get_scram_attribute('s', (char*)(sasl_continue->data + 9), sasl_continue->length - 9, &base64_salt);
    get_scram_attribute('i', (char*)(sasl_continue->data + 9), sasl_continue->length - 9, &iteration_string);


jesperpedersen:
@trxvorr Can you take a look as well ?

jesperpedersen:
This isn't the correct fix

jesperpedersen:
We need to investigate what is happening - use _log_message

ankushT369:
> We need to investigate what is happening - use _log_message

got it!!

trxvorr:
@jesperpedersen @ankushT369 read_message() and ssl_read_message() currently lack reassembly logic, making them susceptible to TCP fragmentation

the fix requires updating the read layer in message.c to parse the type + length header and loop until the full payload is recieved, ensuring it receives complete messages (matching the logic present in write_messafe())

trxvorr:
read_message_from_buffer() is missing that loop aswell

ankushT369:
> @jesperpedersen @ankushT369 read_message() and ssl_read_message() currently lack reassembly logic, making them susceptible to TCP fragmentation
> 
> the fix requires updating the read layer in message.c to parse the type + length header and loop until the full payload is recieved, ensuring it receives complete messages (matching the logic present in write_messafe())

thanks @trxvorr

fheikens:
Taking a look at this to help unstick the PR — sharing a scoping note.

## What the bug actually is

The crash in #834 is a symptom. The real problem, as @trxvorr pointed out, is in `src/libpgagroal/message.c`: the three read functions don't reassemble PostgreSQL frames under TCP fragmentation.

- `read_message` (raw socket)
- `ssl_read_message` (SSL)
- `read_message_from_buffer` (io_uring, called via `pgagroal_recv_message`)

Each returns after the first `read()` / `SSL_read()` succeeds, with `msg->length` set to whatever bytes happened to arrive — not the true on-wire frame length. `pgagroal_write_message` in the same file already has the correct loop-until-done pattern; reads and writes are asymmetric.

## Why a call-site fix isn't enough

Callers above the read layer dereference `msg->data + fixed_offset` without validating `msg->length`. Today that's at least:

- `security.c:1054` — SCRAM server-first (the one #834 crashes in): reads `data + 9` with size `length - 9`, which wraps when `length < 9`
- `security.c:1979` — backend auth response: reads `data + 9`
- `security.c:2084` — SCRAM client-final: reads `data + 5`
- `security.c:2281` — backend key data: reads `data + 5` and `data + 9`
- `security.c:5268` — server-side SCRAM: reads `data + 9`

Patching each site one-by-one leaves the layer's contract broken. Every future caller inherits the same trap.

## What the read layer should do

The contract every caller assumes but the code doesn't enforce:

> If the function returns `OK`, `msg->length` is the true on-wire frame length, and `msg->data[0..length-1]` is a complete PostgreSQL frame.

Per function, the logic:

1. Read until the frame header is available (1 byte kind + 4-byte length for typed frames; startup-message form for un-typed).
2. Parse declared length from the header.
3. Continue reading until the full payload has arrived.
4. Only then return `OK`.

Error cases the layer should surface distinctly (today they either truncate silently or return a partial `OK`):

| Condition | Today | Should be |
|---|---|---|
| Peer closes mid-frame | `OK` with short length | `ERROR` |
| Declared length > buffer | silent truncation | `ERROR` |
| Timeout with partial frame received | `OK` with partial, or `TIMEOUT` with partial | `TIMEOUT`, no partial |
| Full frame in one read (happy path) | `OK` after one read | unchanged — no extra syscall |

Public dispatcher signatures don't need to change — the fix is entirely below them, so the existing call sites don't need auditing.

## Suggested tests

Socket-pair based, one per scenario, all currently failing on master:

- **normal** — complete frame in one segment → `OK` with full frame (passes today)
- **fragmented** — 10 bytes, gap, remaining bytes → must wait for all bytes before returning `OK`
- **oversized** — header declares length > buffer → `ERROR`, not truncation
- **peer-close-mid-frame** — partial write + `shutdown(SHUT_WR)` → `ERROR`, not short-length `OK`
- **timeout-mid-frame** — partial then idle, timeout variant → `TIMEOUT`, no partial

Each should be exercised against all three transports (socket, SSL via BIO pair, io_uring watcher buffer). Nothing in `test/` covers fragmentation today. The io_uring variant may additionally need changes to how the watcher buffer is populated — worth scoping separately if that turns out to be architectural.

## On this PR

@ankushT369 — is this one you'd like to extend to the read-layer change, or would you prefer someone else pick it up from here? Happy to help scaffold the tests either way.

ankushT369:
@fheikens i got the problem what @trxvorr  said I will complete it soon I just need a time I am very much busy right now by the end of this month I will solve all the issues or PR i have created.

thanks
Ankush

@fheikens apart from this are you using some-kind of AI tool ?? just curious !!

fheikens:
@ankushT369 Yes — Claude, mostly as a writing assistant. English isn't my first language and an LLM types faster than I do. I write the prompts, review the output, cross-check against the code, and lean heavily on testing — my background is in software testing (professional training, former test coordinator), so the test focus is where my background is, not borrowed from the tooling. The reproducers on #848 are recent examples. No AI co-author trailers on commits; everything I submit is work I've understood and verified.

ankushT369:
> @ankushT369 Yes — Claude, mostly as a writing assistant. English isn't my first language and an LLM types faster than I do. I write the prompts, review the output, cross-check against the code, and lean heavily on testing — my background is in software testing (professional training, former test coordinator), so the test focus is where my background is, not borrowed from the tooling. The reproducers on #848 are recent examples. No AI co-author trailers on commits; everything I submit is work I've understood and verified.

@fheikens you can take this issue !!! and solve it !

thanks,
Ankush

fheikens:
Triage note: the fix is on-topic for #834. CI is red on the two clang jobs — build-linux (clang, Debug) and (clang, Release) — while the gcc jobs pass, so it looks like a clang-specific compile issue. Could you check those job logs and push a fix? Once it is green this is a good review candidate, being a crash fix.

PR #847, opened 2026-04-10
[#841] Refactor pgagroal conf get output to decouple CLI and JSON for…

### LINKED ISSUE #841: Separate CLI and JSON output in conf get
Right now, `conf get` returns the same structure for both CLI output and JSON

Running something like `pgagroal-cli conf get startup_validation` currently returns `1` instead of `try`, which isn’t very helpful - no one memorizes what each number represents.

The same issue applies to other enum parameters like `validation`, `hugepage`, `pipeline`, `log_type`, `log_mode`, and `update_process_title`.

The idea is to tailor each format to its use case:

* CLI - should stay simple and human-friendly, showing values like `try`, `off`, `foreground`.
* JSON - should be more explicit by including both the numeric value and its label. For example:

```json
"startup_validation": { "value": 1, "label": "try" },
"validation": { "value": 0, "label": "off" },
"hugepage": { "value": 1, "label": "try" }
```

This way, scripts can rely on stable numeric values, while still having readable labels for logging or display.

Abdelrhmansersawy:
resolved #841

## Changes
1. Add `pgagroal_json_put_enum_value()`, `pgagroal_json_put_time_value()`, and `pgagroal_json_put_size_value()` to format Enum, Time, and Size parameters as structured objects (`{"value": N, "string_value": "..."}`)
2. Update the configuration response to emit these enriched structured objects instead of raw numeric values
3. Modify the CLI parser to extract the human-readable string for standard text output, while preserving the full structured objects for the management JSON output

## Test
```
% pgagroal-cli conf get validation
off

% pgagroal-cli conf get idle_timeout
600s

% pgagroal-cli json conf get hugepage
{
  "Header": {
    "ClientVersion": "2.1.0",
    "Command": 3,
    "Compression": 0,
    "Encryption": 0,
    "Output": 1,
    "Timestamp": "20260410092744"
  },
  "Outcome": {
    "Status": true,
    "Time": "00:00:00"
  },
  "Request": {},
  "Response": {
    "hugepage": {
      "string_value": "try",
      "value": 1
    }
  }
}
```

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -1729,15 +1729,36 @@ get_config_key_result(char* config_key, struct json* j, uintptr_t* r, int32_t ou
       else if (!strcmp(key, iter->key))
       {
          // Handle single or two-part keys
-         config_value = pgagroal_value_to_string(iter->value, FORMAT_TEXT, NULL, 0);
          if (iter->value->type == ValueJSON)
          {
-            struct json* server_data = NULL;
-            pgagroal_json_clone((struct json*)iter->value->data, &server_data);
-            pgagroal_json_put(filtered_response, key, (uintptr_t)server_data, iter->value->type);
+            struct json* nested_obj = (struct json*)iter->value->data;
+
+            /* Handle enriched values formatted as { "value": N, "string_value": "..." } */
+            if (pgagroal_json_contains_key(nested_obj, "string_value"))
+            {
+               /* Extract the string value for text output */
+               config_value = strdup((char*)pgagroal_json_get(nested_obj, "string_value"));
+
+               if (output_format == MANAGEMENT_OUTPUT_FORMAT_JSON)
+               {
+                  /* Emit the full nested object in JSON mode */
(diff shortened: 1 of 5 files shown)


fluca1978:
This looks great to me, I'm wondering if you can also add a few tests so that we ensure that the affected values are now displayed as enumeration, strings, or time and that json has now two indexes `value` and `string_value`.
<https://github.com/pgagroal/pgagroal/blob/master/test/testcases/test_configuration.c>

Abdelrhmansersawy:
@fluca1978 PTAL

fluca1978:
Merged with a commit message.

Thanks @Abdelrhmansersawy

Abdelrhmansersawy:
@fluca1978 
I think this enhancement we could cross-port to other project

@jesperpedersen WDYT?

jesperpedersen:
@Abdelrhmansersawy Yes, this is a cross-port

Issue #851, opened 2026-04-13
blocking_timeout default interacts poorly with transaction pipeline


fheikens:
The default value of `blocking_timeout` is 30 seconds (`src/include/pgagroal.h`, `DEFAULT_BLOCKING_TIMEOUT`).

`doc/PIPELINES.md` recommends setting `blocking_timeout` to 0 when using the transaction pipeline, because clients may time out while waiting for a backend when the pool is saturated.

There is currently no startup warning or validation that surfaces this interaction, and the default 30s remains in effect unless the operator reads the pipeline documentation and adjusts it manually.

Questions for the maintainers:

1. Is the current default intentional for the transaction pipeline?
2. Would it make sense to adjust the default when `pipeline = transaction` is selected, or to emit a startup warning when both are set together?
3. Is a configuration validation rule (e.g. logging a WARN at startup) acceptable for this case?

Happy to follow up with a patch once the preferred direction is clear.

fheikens:
Fixed by #852, which kept the short `blocking_timeout is not recommended` warning for the transaction pipeline and added a second warning pointing to `doc/PIPELINES.md` for configuration guidance — instead of cramming the explanation into the log message. Closing.

PR #852, opened 2026-04-13
[#851] Clarify blocking_timeout warning for the transaction pipeline

fheikens:
Addresses #851.

## Summary

- Extend the existing startup warning emitted when `blocking_timeout` is set together with `pipeline = transaction` so operators see the concrete consequence and the recommended value.

Before:

```
pgagroal: Using blocking_timeout for the transaction pipeline is not recommended
```

After:

```
pgagroal: Using blocking_timeout for the transaction pipeline is not recommended; clients may time out while waiting for a backend when the pool is saturated. Set blocking_timeout to 0 for the transaction pipeline.
```

No default values and no behavior are changed. If the maintainers prefer a different direction (adjusting the default, emitting a FATAL, or a dedicated validation path), happy to follow up.

## Test plan

- [ ] Start pgagroal with `pipeline = transaction` and the default `blocking_timeout = 30s`; confirm the updated warning is logged.
- [ ] Start pgagroal with `pipeline = transaction` and `blocking_timeout = 0`; confirm no warning is logged.

### DIFF
--- a/src/libpgagroal/configuration.c
+++ b/src/libpgagroal/configuration.c
@@ -766,6 +766,7 @@ pgagroal_validate_configuration(void* shm, bool has_unix_socket, bool has_main_s
       if (pgagroal_time_is_valid(config->blocking_timeout))
       {
          pgagroal_log_warn("pgagroal: Using blocking_timeout for the transaction pipeline is not recommended");
+         pgagroal_log_warn("pgagroal: See doc/PIPELINES.md for transaction pipeline configuration guidance");
       }
 
       if (pgagroal_time_is_valid(config->idle_timeout))


src/libpgagroal/configuration.c
```diff
@@ -790,7 +765,7 @@ pgagroal_validate_configuration(void* shm, bool has_unix_socket, bool has_main_s
 
       if (pgagroal_time_is_valid(config->blocking_timeout))
       {
-         pgagroal_log_warn("pgagroal: Using blocking_timeout for the transaction pipeline is not recommended");
+         pgagroal_log_warn("pgagroal: Using blocking_timeout for the transaction pipeline is not recommended; clients may time out while waiting for a backend when the pool is saturated. Set blocking_timeout to 0 for the transaction pipeline.");
```
fluca1978: While I'm a huge fan of making logs verbose, I don't think we need to explain such a behavior in the logs.
Rather add another log warning to reference the documentation?

fluca1978 (changes requested):

fheikens:
Reverted the expanded warning to the original short form and added a second warning line pointing to doc/PIPELINES.md.

PR #855, opened 2026-04-15
[#842] Detect PostgreSQL version mismatches across servers

### LINKED ISSUE #842: Store PostgreSQL version from pg_control_system() for server compatibility checks
followup with: https://github.com/pgagroal/pgagroal/commit/8a8ba4f990d1aef36828802cc8c9a7531dfb599a#r181348204

During startup validation, we already query `pg_control_system()` to fetch the `system_identifier` for each configured server. This same function also returns the PostgreSQL version, but we currently discard it.

Storing the version alongside the identifier would let us add useful checks down the line, such as:

- Detecting servers running a major version too old to support streaming replication

- Warning when configured servers are on different major versions

- Skipping or failing gracefully on servers where `pg_control_system()` doesn't exist (PG < 9.6)

Abdelrhmansersawy:
resolved #842 

## Changes
1. Queries `pg_control_version` from `pg_control_system()` during startup validation to detect when configured servers are running different major PostgreSQL versions. Emits a warning without blocking startup.

2. Stores the major version in the `server` struct version attribute for later use. Exposes it in `status details` output as MajorVersion.

3. Adds a test case to verify `pg_control_version` is fetched and stored.


Abdelrhmansersawy:
side note: I see that it would be good to store `pg_control_version` in the `version` attribute inside the `server` struct during the startup check, and expose it in `pgagroal-cli status details` like this:

```
Servers:
    - Health: UNKNOWN
      Host: localhost
      MajorVersion: 18
      Port: 5432
      Server: primary
      State: Not init
```

as it would help DBAs quickly detect inconsistent PostgreSQL versions alongside `_WARN` logs.

src/libpgagroal/server.c
```diff
@@ -141,6 +143,33 @@ pgagroal_check_server_identifiers(void)
       }
    }
 
+   /* Check for version mismatches across servers */
+   for (int i = 0; i < num_servers; i++)
```
fluca1978, line 153: I think we should get the primary server version as a baseline, and then check all the remaining servers ( or all, it does not matter that much) against such version. If they are different, I believe it does suffice.
Imagine we have three servers: `17`, `18`, `16`. 
So far we are detecting `17 != 18` and `18 != 16`, that is fine. Assume `17` is the primary, so we should report `18 != 17` and `16 != 17`, then who cares of `18 != 16` (the replicas).
WDYT?

src/libpgagroal/status.c
```diff
@@ -211,6 +211,7 @@ status_details(bool details, struct json* response)
       pgagroal_json_put(js, MANAGEMENT_ARGUMENT_PORT, (uintptr_t)config->servers[i].port, ValueInt32);
       pgagroal_json_put(js, MANAGEMENT_ARGUMENT_STATE, (uintptr_t)pgagroal_server_state_as_string(config->servers[i].state), ValueString);
       pgagroal_json_put(js, MANAGEMENT_ARGUMENT_HEALTH, (uintptr_t)health_str, ValueString);
+      pgagroal_json_put(js, MANAGEMENT_ARGUMENT_MAJOR_VERSION, (uintptr_t)config->servers[i].version, ValueInt32);
```
fluca1978, line 214: at this point, I would expose the server identifier too, in the JSON only output.
WDYT?

fluca1978 (changes requested):
Great, I think we should expose the version number (and identifier?) also in prometheus. I would expose the identifier only in JSON format.

Abdelrhmansersawy:
@fluca1978 PTAL

Abdelrhmansersawy:
> I think we should expose the version number (and identifier?) also in Prometheus

Good idea, I just had a small question: how could a metric like “PG Version” be useful in Prometheus? Would it be something we could meaningfully use in Grafana panels, and if so, in what way?

From a DBA perspective, I’m trying to understand how this would improve visualization or monitoring. Since this attribute doesn’t change frequently (mostly only during PostgreSQL migrations), I’m curious about where it would add the most value.

I’d really like to better understand how pgagroal metrics are designed with DBAs in mind.

fluca1978:
> > I think we should expose the version number (and identifier?) also in Prometheus
> 
> Good idea, I just had a small question: how could a metric like “PG Version” be useful in Prometheus? Would it be something we could meaningfully use in Grafana panels, and if so, in what way?
> 

In grafana I usually show the version of a server to manage and recognize upgrades.
But this case is different: we are exposing the major version (e.g., 18), so effectively that is going to stay there for a long time. On the other hand, the minor could change quicker and can be different from primary and standbys.

@jesperpedersen WDYT? According to me this PR is ready, but now I've the doubt if it really make sense to expose the PostgreSQL major version number via prometheus. We could use it as the basis for exposing the full version later on.

jesperpedersen:
@fluca1978 We have to rely on people not doing something stupid. pgexporter exposes the major/minor version number. That can be turned into an alert if it is outdated or EOLed.

Good to go

fluca1978:

PR #857, opened 2026-04-20
[#710] Fix system_identifier fatal check for non-primary duplicates

Abdelrhmansersawy:
Make duplicate system_identifier fatal only for replica-replica conflicts (primary=off/off). Add coverage for on/off and off/off scenarios and update documentation for validation behavior.


fluca1978:

Issue #767, opened 2026-02-27
config set and reload operations fail or cause instability


Userfrom1995:
When executing a configuration change via `pgagroal-cli conf set` or `pgagroal-cli reload`, the daemon experiences network binding conflicts (e.g., Address already in use) and/or severe accept loop spam (e.g., repeated Bad file descriptor errors).

Shyam-123pandey:
<img width="1919" height="1019" alt="Image" src="https://github.com/user-attachments/assets/4b577936-58bb-4979-882a-8662cc665fb4" />

Hello, I would like to investigate this issue.
I have successfully built pgagroal locally and am currently exploring the configuration reload flow to understand how pgagroal-cli conf set and pgagroal-cli reload interact with the running daemon.

I will try to reproduce the issue and trace where the socket binding or file descriptor handling might be failing during reload.

I will update here once I have more findings.

Userfrom1995:
> Hello, I would like to investigate this issue. I have successfully built pgagroal locally and am currently exploring the configuration reload flow to understand how pgagroal-cli conf set and pgagroal-cli reload interact with the running daemon.
> 
> I will try to reproduce the issue and trace where the socket binding or file descriptor handling might be failing during reload.
> 
> I will update here once I have more findings.

If you're going to investigate it, a few things to look at are how `reload` and `set` are implemented.
The address binding issue should be easy to fix.
Also, please take a look at this comment:
[https://github.com/pgagroal/pgagroal/pull/765#issuecomment-3975194101](https://github.com/pgagroal/pgagroal/pull/765#issuecomment-3975194101)

Shyam-123pandey:
Okay thanks. I am exploring.

fluca1978:
@Shyam-123pandey I've assigned this issue to you, so please dig into the problem and ask here for help while progressing.
Also, please do not post images if not needed.

Shyam-123pandey:
Thanks to assign me in next few days I'll put my 💯

Shyam-123pandey:
<img width="1919" height="967" alt="Image" src="https://github.com/user-attachments/assets/e18cdb26-0ffb-46ce-9e1f-9fe220012434" />
Hello Sir Can you here, Just tell is this issue have been fixed or in which environment this have been occur please tell, because i tried lots to produce but not able to produce this issue. Always getting success Reload and client connected.

Userfrom1995:
[...]

PR #859, opened 2026-04-20
[#767] Refactored the configuration set and reload command remove the support for port reloads

Userfrom1995:
- Refactored 'conf set' and 'conf reload' logic to enforce restart requirements for architectural settings (host, port, hugepage, max_connections, etc.).
- Removed legacy code that attempted to reload listener sockets and memory-critical arrays live, which previously led to inconsistencies.
- Updated documentation and manuals to reflect the changes in conf `reload` and `set`.

#767

### DIFF
--- a/doc/CLI.md
+++ b/doc/CLI.md
@@ -416,7 +416,7 @@ pgagroal-cli conf set max_connections 1000 --format json
     "message": "Configuration change requires restart. Current values preserved.",
     "requested_value": "1000",
     "restart_required": true,
-    "status": "success_restart_required"
+    "status": "restart_required"
   }
 }
 ```
@@ -426,17 +426,11 @@ pgagroal-cli conf set max_connections 1000 --format json
 ### **Important**
 
 > **Warning:**  
-> When changing critical parameters such as the **main port**, **metrics port**, **management port**, or **unix_socket_dir**, you must choose values carefully.
+> When changing critical structural parameters such as the **main port (host, port)**, **metrics**, **management**, **console**, or **unix_socket_dir**, the changes cannot be applied to the active pooling engine without a restart.
 >
-> If you set a port or socket to a value that is already in use or unavailable, the reload will fail. In this case, the CLI may still report "success" because the configuration was accepted, but the server will **not** be listening on the new port or socket.  
+> In these cases, the CLI will return a **Restart Required** (`status: restart_required`) response and will **not** modify the configuration (neither on disk nor in memory). This ensures that the configuration correctly reflects the actually running state of the daemon. To apply these changes, you must manually edit the configuration files and restart the `pgagroal` service.
 >
-> **How to recover:**  
-> If you accidentally set a port or socket to an unavailable value, simply use the `conf set` command again to set it to a valid, available value. The server process will remain running and can be reconfigured as needed.
->
-> **Recommendation:**  
-> - Avoid setting ports or sockets to values that are already in use by other services.
-> - Always verify the availability of the desired port or socket before applying changes.
-> - Review server logs for detailed error messages if a configuration change does not take effect as expected.
+> Note: The `conf set` command is **in-memory only**; it does not persist changes to the configuration files on disk.
 >
 > **Multiple Entries in `limit` or `hba`:**  
 > If your configuration contains multiple entries with the **same database name** in the `limit` section or the **same username** in the `hba` section,  
--- a/doc/manual/en/14-cli-tools.md
+++ b/doc/manual/en/14-cli-tools.md
@@ -322,12 +322,11 @@ Configuration change failed
 ```
 
 > **Warning:**  
-> When changing critical parameters such as the **main port**, **metrics port**, **management port**, or **unix_socket_dir**, you must choose values carefully.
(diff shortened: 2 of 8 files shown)


doc/manual/en/14-cli-tools.md
```diff
@@ -317,12 +317,11 @@ Configuration change failed
 ```
 
 > **Warning:**  
-> When changing critical parameters such as the **main port**, **metrics port**, **management port**, or **unix_socket_dir**, you must choose values carefully.
+> When changing critical structural parameters such as the **main port (host, port)**, **metrics**, **management**, **console**, or **unix_socket_dir**, the changes cannot be applied to the active proxying engine without a restart.
```
fluca1978: `proxying` -> `pooling`?

doc/CLI.md
```diff
@@ -421,17 +421,11 @@ pgagroal-cli conf set max_connections 1000 --format json
 ### **Important**
 
 > **Warning:**  
-> When changing critical parameters such as the **main port**, **metrics port**, **management port**, or **unix_socket_dir**, you must choose values carefully.
+> When changing critical structural parameters such as the **main port (host, port)**, **metrics**, **management**, **console**, or **unix_socket_dir**, the changes cannot be applied to the active proxying engine without a restart.
```
fluca1978: `proxying` -> `pooling`?

src/libpgagroal/configuration.c
```diff
@@ -3552,55 +3550,137 @@ extract_alias_with_space(char* str, int offset, char** db_part)
...
+transfer_configuration(struct main_configuration* config, struct main_configuration* reload)
+{
 #ifdef HAVE_SYSTEMD
    sd_notify(0, "RELOADING=1");
 #endif
 
+   /* Check if any parameter requires a restart before applying any changes */
+   if (check_restart_required(config, reload))
+   {
+      pgagroal_log_warn("Configuration reload deferred: restart required for one or more parameters. Running state preserved.");
```
fluca1978: `deferred` -> `debied`?

src/include/configuration.h
```diff
@@ -101,6 +101,10 @@ extern "C" {
 #define CONFIGURATION_ARGUMENT_VALIDATION                       "validation"
 #define CONFIGURATION_ARGUMENT_STARTUP_VALIDATION               "startup_validation"
 #define CONFIGURATION_ARGUMENT_BACKGROUND_INTERVAL              "background_interval"
+#define CONFIGURATION_ARGUMENT_HEALTH_CHECK                     "health_check"
+#define CONFIGURATION_ARGUMENT_HEALTH_CHECK_PERIOD              "health_check_period"
+#define CONFIGURATION_ARGUMENT_HEALTH_CHECK_TIMEOUT             "health_check_timeout"
+#define CONFIGURATION_ARGUMENT_HEALTH_CHECK_USER                "health_check_user"
```
fluca1978, line 107: do these belong here?

fluca1978 (commented):
I like the approach, sounds interesting to avoid reloading partial configuration.

jesperpedersen:
@Userfrom1995 Remember squash

Userfrom1995:
> @Userfrom1995 Remember squash

We have four different commits, and each commit solves a different issue. So we would need 3 more issues, which I will open.

I was thinking we could review everything here since the other three are relatively small fixes, and then merge all four together afterward. However, if you prefer, I can also open 3 separate PRs for them instead. Let me know what you would prefer.

This one will need a bit more testing for the set and reload functionality. I have done some initial testing, and it looks good so far. Once I am fully satisfied with the testing, I will merge it.

@fluca1978 Could you please have another look and, if possible, please test the reload, set, and get changes as well?

Also, what are your thoughts on the other changes?

doc/CLI.md
fluca1978: still proxying, is this supposed to be pooling?

src/libpgagroal/health.c
```diff
@@ -105,13 +105,14 @@ pgagroal_health_check_stop(void)
...
-      for (int i = 0; i < 50; i++)
+      /* Wait up to 2s for exit - the worker checks its flags every 1s via sleep(1) */
+      for (int i = 0; i < 10; i++)
       {
          if (kill(config->health_check_pid, 0))
          {
             break;
          }
-         SLEEP(100000000L);
+         SLEEP(200000000L);
```
fluca1978, line 115: please document this change in the commit message

fluca1978 (commented):
Look good to me, but I'm not sure the changes in `main.c` about the periodic watcher do belong in here. Aren't they related to another refactoring round?

jesperpedersen:
@Userfrom1995 Remember to squash

Userfrom1995:
First we have to merge #900, #901, and #902, then this one can go in.
After those three are merged and this branch is rebased, only one commit should remain.

Userfrom1995:
> Look good to me, but I'm not sure the changes in `main.c` about the periodic watcher do belong in here. 
Aren't they related to another refactoring round?

These changes belong here because the new service_reload logic needs to refresh the periodic watchers. main.c already had this code, so rather than copy-pasting it, I extracted it into a static function so it could be cleanly called from both places. I believe the function extraction is justified here and will be useful in the future. However, if you'd prefer, I'd be happy to split this specific extraction into a separate commit or open a separate issue for it.

Userfrom1995:
@fluca1978 @jesperpedersen PTAL.
Once it looks good to you, I’ll merge it.

Issue #776, opened 2026-03-01
Automated configuration scaffolding via pgagroal-admin


trxvorr:
currently, new users setting up `pgagroal` can copy the boilerplate files from `doc/etc/`. however, they still have to manually edit multiple files (`pgagroal.conf`, `pgagroal_hba.conf`, user vaults, etc.) and understand the syntax to link the pooler to their backend database. for beginners or automated deployment scripts, this manual editing can be a source of friction and typos.

idea is to implement a native C interactive setup utility inside `pgagroal-admin`

running `pgagroal-admin init <directory>` would trigger standard `stdin` prompts in the terminal:
* "Listen port for pgagroal [2345]:"
* "Primary database IP [127.0.0.1]:"
* "Primary database port [5432]:"
* "Setup User Vault for authentication? (y/n):"
* "Generate systemd .service and .socket files? (y/n):"

the tool would then automatically generate the complete suite of configuration files (`.conf`, `.hba`, `vault.conf`, and `.service`/`.socket` if requested) in the target directory, fully linked and pre-filled with the user's specific answers.

jesperpedersen:
@trxvorr There are a lot of options though so maybe we should just do a "best practice" route and only ask for the basic information about the primary cluster

fluca1978:
This sounds to me like a good application of templates: we should provide a  basic catch-all configuration example template that can be adjusted via some minimal questions. I would not write a full configuration manager, and besides I will not insert it into `pgagroal-admin`, rather `pgagroal-cli conf <subcommand>`.

jesperpedersen:
`pgagroal-cli conf init`

fluca1978:
> `pgagroal-cli conf init`

Sounds great to me, but again, this is a template engine task according to me, I'm not sure it is a dependency we want to introduce.
On the other hand, why not provide a command to test that the basic configuration is in place?

jesperpedersen:
@fluca1978 We can create a new binary to handle all of this

fluca1978:
> [@fluca1978](https://github.com/fluca1978) We can create a new binary to handle all of this

Uhm.., I am against the proliferation of binaries, in fact I would like to see one day `pgagroal-admin` and `pgagroal-cli` merged together. Therefore, if we would like to add another executable (e.g., `pgagroal-conf`) I propose to:
- make it manage all the configuration (i.e., `pgagroal-cli conf` implementation moved to this new executable)
[...]

PR #865, opened 2026-04-25
[#864] Add Configuration Generator

### LINKED ISSUE #864: Configuration generator
Following the discussion in #1088, this issue tracks the cross-porting of the configuration generator.

## References
- Cross-port suggested in: pgmoneta/pgmoneta#1088
- Original feature in pgmoneta: pgmoneta/pgmoneta#516

Sakshii-27:
Closes #864, #776
This PR introduces the `pgagroal-config` utility, cross-ported from `pgmoneta`.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1,0 +1,1339 @@
+/*
+ * Copyright (C) 2026 The pgagroal community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+ * TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* pgagroal */
+#include <pgagroal.h>
+#include <configuration.h>
+#include <logging.h>
+#include <utils.h>
+
+/* system */
+#include <ctype.h>
+#include <err.h>
+#include <errno.h>
+#include <getopt.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <time.h>
+#include <unistd.h>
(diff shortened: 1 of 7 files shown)


trxvorr:
@Sakshii-27 hey Sakshi i haven't gone through the code or checked pgmoneta implementation 
but is it implementing #776, if it is do check the comments of that aswell and add it to the "closes: " line in description

Sakshii-27:
Hey @trxvorr! Yes, this is related to #776. I cross-ported the configuration generator from pgmoneta, which implements the approach discussed in your issue, a separate offline binary (`pgagroal-config`) that generates a best-practice `pgagroal.conf` skeleton. Thanks for flagging it.

src/config.c
```diff
@@ -0,0 +1,1303 @@
...
+   /* Open output file initially as .tmp to be safe */
+   file = fopen(tmp_path, "w");
+   if (file == NULL)
+   {
+      warn("Could not open temp file '%s'", tmp_path);
+      goto error;
+   }
+
+   /* Write [pgagroal] section */
+   write_section(file, "pgagroal");
```
fluca1978: should we write something like an header in the file?
Like
```
# generated by pgagroal-conf version x.y
# on date yyyy-mm-dd
```

src/config.c
```diff
@@ -0,0 +1,1303 @@
...
+
+   if (!quiet)
+   {
+      printf("--- [pgagroal] section ---\n\n");
+   }
+
+   if (quiet)
+   {
+      pgagroal_snprintf(host, sizeof(host), "*");
+      pgagroal_snprintf(port, sizeof(port), "2345");
```
fluca1978: We should use as much as possible `#define` for default values.
So I guess it is time to `#define` default values and use them in here.

src/config.c
```diff
@@ -0,0 +1,1303 @@
...
+   file = fopen(tmp_path, "w");
+   if (file == NULL)
+   {
+      warn("Could not open temp file '%s'", tmp_path);
+      goto error;
+   }
+
+   /* Write [pgagroal] section */
+   write_section(file, "pgagroal");
+   write_key_value(file, "host", host);
```
fluca1978: here we should use `@define` keys from `config.h`, like `CONFIGURATION_ARGUMENT_HOST`

fluca1978 (commented):
I like the idea, but I think we should use the same parsing from `libpgagroal/config.c` to avoid code to be out of sync.

Sakshii-27:
@fluca1978 Thanks for the review! I've pushed an update addressing all your points. PTAL

Sakshii-27:
The core parser in `libpgagroal/configuration.c` loads the file into a struct and discards comments and formatting. Since `set/del` need to modify files while preserving the user's comments, a line-by-line approach is required here. To keep things in sync, I've moved all default values and config keys to `#define` in `configuration.h` so both paths share the same constants.

jesperpedersen:
I think the important part is precisely that. Users add comments to their configuration in order to justify why they did that way. So, we can't remove comments from existing files.

If it is a default we are adding then we probably don't need to add a comment, however if the user is adding something non-default we should take an optional comment as well

fluca1978:
> The core parser in `libpgagroal/configuration.c` loads the file into a struct and discards comments and formatting. Since `set/del` need to modify files while preserving the user's comments, a line-by-line approach is required here. To keep things in sync, I've moved all default values and config keys to `#define` in `configuration.h` so both paths share the same constants.

Right, but so far we are nuking inline comments in any case, unless you explicitly specifiy it as argument.

```
% pgagroal-config get pgagroal.conf pgagroal ev_backend
kqueue   # this is wrong!

% pgagroal-config set pgagroal.conf pgagroal ev_backend io_uring

% pgagroal-config get pgagroal.conf pgagroal ev_backend         
io_uring

% pgagroal-config set pgagroal.conf pgagroal ev_backend "epoll #foo"
% pgagroal-config get pgagroal.conf pgagroal ev_backend             
epoll #foo
```

Therefore, while I agree with the current implementation, I think in the long term we should try to unify the configuration parsing.
As an example of "code out of sync", see <https://github.com/pgagroal/pgagroal/commit/bf97eee671b2a6f8573e20ff2855164e085933fb> where the options are not what `config.c` expects later on (most notably `liburing` is `io_uring`).


Another thing I see is that every command requires a file (except for `init`), so probably the file argument should be specified via an option, something like `-f <file>`. This simplifies the command line, according to me, and seems more consistent with the other tools (e.g., `pgagroal -f /etc/pgagroal/pgagroal.conf`). This can be an issue for a later improvement.

Sakshii-27:
Thanks for the feedback @jesperpedersen and @fluca1978! I've updated the PR with the following fixes:

1. **Optional Comments**: Added an optional 5th argument to `set` to allow users to add/preserve comments.
2. **Fixed `get` Output**: `get` now correctly strips inline `#` and `;` comments before returning the value.
3. **Synced Options**: Updated the prompt hints to match the core parser (`poll` -> `kqueue`, added `trace`). 

@fluca1978 I completely agree that a unified parser is the right long-term solution to prevent drift, and that using a `-f` flag instead of a positional argument is much cleaner. Should I open follow-up issues for the unified parser and the `-f` flag CLI redesign?

fluca1978:
> @fluca1978 I completely agree that a unified parser is the right long-term solution to prevent drift, and that using a `-f` flag instead of a positional argument is much cleaner. Should I open follow-up issues for the unified parser and the `-f` flag CLI redesign?

Yes, the above is future work, so open issues about those.

fluca1978:
Merged, with a commit message,.
Thanks @Sakshii-27

Issue #876, opened 2026-04-28
Guard `backend_secret` trace logging behind `#ifdef DEBUG` and document trace-level data sensitivity


fheikens:
**Goal**

Restore the `Backend Secret: %d` trace logs in `pgagroal_pool_status` — the eight per-state diagnostic blocks in `connection_details()` (`pool.c`) — but wrap them with `#ifdef DEBUG` so they only compile into debug builds. Then document that `log_level = trace` may include sensitive data, so operators who enable it understand the implication.

**Background**

PR #868 removed the `backend_secret` from a debug log and from these eight trace logs to avoid leaking cancel-protocol secrets through captured logs. On merge, [@fluca1978 specified the follow-up direction](https://github.com/pgagroal/pgagroal/pull/868#issuecomment-4333418760):

> With regard to logging again the secret, I would trace it with a guard against production at compile time. Then we should document that `log_level = trace` could log "sensible" data.

**Scope**

1. **Code.** Restore the eight `pgagroal_log_trace("                      Backend Secret: %d", connection.backend_secret);` lines in `pool.c` (in `connection_details()`, one per state branch). Wrap each in `#ifdef DEBUG` / `#endif`. The existing pattern is already used across the codebase: `src/include/logging.h:68`, `src/include/pgagroal.h:101`, `src/libpgagroal/memory.c`, `src/libpgagroal/ev.c`, `src/libpgagroal/deque.c`. Release and Performance builds will not include the secret in trace output.

2. **Docs.** Add a short note — alongside the `log_level` description in `doc/CONFIGURATION.md` (and the manual equivalent at `doc/manual/en/04-configuration.md`) — explaining that `log_level = trace` may log sensitive data and is intended for development/troubleshooting, not production.

**References**

- PR #868 — the source change
- [fluca1978's merge comment specifying compile-time guard + docs](https://github.com/pgagroal/pgagroal/pull/868#issuecomment-4333418760)

PR #877, opened 2026-04-28
[#876] Guard backend_secret trace logging behind #ifdef DEBUG

fheikens:
Implements #876.

**Code change** (`src/libpgagroal/pool.c`)

Wraps the eight `Backend Secret: %d` `pgagroal_log_trace` lines in `connection_details()` with `#ifdef DEBUG` / `#endif`. The secret is now only emitted in debug builds — Release and Performance builds compile it out. The PID lines next to each are kept as-is (not sensitive on their own).

The `#ifdef DEBUG` pattern matches existing usage in the codebase (`logging.h`, `pgagroal.h`, `memory.c`, `ev.c`, `deque.c`).

**Doc change**

Added a short note alongside the `log_level` description in both `doc/CONFIGURATION.md` and `doc/manual/en/04-configuration.md`:

> Note that `TRACE` is intended for development troubleshooting and may include sensitive data; it is not recommended for production.

The manual entry was also updated to include `TRACE` in the level list (it was missing the entry that the main config doc already has).

**Reference**

Follow-up to #868. Direction set by [@fluca1978's merge comment](https://github.com/pgagroal/pgagroal/pull/868#issuecomment-4333418760).


fheikens:
Friendly ping on this one @fluca1978 — it's a small security-hardening change that guards `backend_secret` from ever reaching trace logs by putting it behind `#ifdef DEBUG`. CI is green across all jobs and it still merges cleanly against current master. Happy to adjust if you'd prefer a different guard mechanism.

fluca1978:
Sorry for the delay, I missed this, shame on me.
Thanks for you contribution @fheikens

Issue #848, opened 2026-04-13
Transaction pipeline does not enforce max_size — intended behavior or missing cap?


fheikens:
## Summary

When using `pipeline = transaction`, the configured `max_size` limit is not enforced. The number of backend connections can grow beyond `max_size` up to `max_connections`.

## Observed behavior

- The `max_size` check in `pool.c` is guarded by `!transaction_mode`
- The transaction pipeline calls `pgagroal_get_connection(..., transaction_mode=true, ...)`
- As a result, `max_size` is not applied in transaction mode
- Under load, the number of backend connections grows beyond the configured pool limit

Reference: `src/libpgagroal/pool.c`, specifically the conditional around:

```c
if (*slot == -1 && !transaction_mode)
```

## Expected behavior

In transaction pooling, a common expectation is:

- Many client connections are multiplexed onto a bounded number of backend connections
- `max_size` acts as a hard cap
- When the pool is exhausted, requests wait instead of opening new backend connections

## Current behavior

- `max_size` is effectively ignored in transaction mode
- Backend growth is instead bounded by `max_connections`
- This allows the pool to grow beyond the configured limit during bursts

## Questions

1. Is this behavior intentional?
2. If yes, should this be documented more explicitly?
3. Would it make sense to optionally enforce `max_size` in transaction mode (queue instead of grow)?

## Why this matters

- Risk of unexpectedly hitting PostgreSQL connection limits
- Harder capacity planning
- Behavior differs from typical transaction pooling expectations

fluca1978:
Isn't the code protected by <https://github.com/pgagroal/pgagroal/blob/master/src/libpgagroal/pool.c#L112>?
So we have a test that fails in keeping `max_connections`?

fheikens:
Posting a minimal reproducer and one clean run.

### Summary

With `pipeline = transaction` and a per-rule `max_size` set in `pgagroal_databases.conf`, the number of backend connections pgagroal holds open at PostgreSQL grows well beyond `max_size`, up to (not past) the global `max_connections`. The effective cap is `max_connections`, not `max_size`.

### Setup

- pgagroal: `elevarq/pgagroal:2.1.1` (binary reports `pgagroal 2.1.0`)
- PostgreSQL: `postgres:16` with PG-side `max_connections=500` (not the limiter)
- `pgagroal.conf`: `pipeline = transaction`, `max_connections = 500`
[...]

PR #878, opened 2026-04-28
[#875] Fix retry path missing freed slots under concurrent contention

### LINKED ISSUE #856: Core dump in pgagroal_pool_status() due to race condition in active_connections counter
**Bug Description**

I faced a coredump while working with transaction pipeline and upon further investigation i found the following:

There is an assert statement in pool.c in pgagroal_pool_status() that is checked in DEBUG mode

Image

in pool.c : pgagroal_get_connection()  config->active_connections is atomically incremented and then is checked verus max_connections, and if exceeded it goes to retry block that then decrements active_connections

Image

This leaves some time where config->active_connections exceeds max_connections which causes assert to fail if pool_status function was called before decrement was complete

**To Reproduce**

1. I have added a 10 seconds sleep period before going to retry & decrementing active_connections to be able to capture the period that it exceeds max_value 
   note: I did that because the window was hard to replicate without adding a sleeping period to amplify it

Image

2. I built a debug build

3. I used configuration with max_connections = 1 & transaction_pipeline

4.  start pgagraol

5.  Open two connections and send a simple select 1; (the first one will run while the second one is pending on a connection and active_connections is incremented to 2

6. I send a SIGTRAP signal to pgagroal to force pgagroal_pool_status to get called

result: coredump
 
```
2026-04-17 02:16:58 INFO  pool.c:114 EDITTED: Triggered race condition sleep. Pool is over max limit. will sleep for 10
agroal: zeyaddaowd@localhost:5433/test: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
2026-04-17 02:17:01 INFO  main.c:2690 pgagroal: core dump requested
agroal: zeyaddaowd@localhost:5433/test: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
agroal: main: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
2026-04-17 02:17:01 INFO  main.c:2690 pgagroal: core dump requested
agroal: main: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
agroal: zeyaddaowd@localhost:5433/test: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_poolagroal: authenticating: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Asserti_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
2026-04-17 02:17:01 INFO  main.c:2690 pgagroal: core dump requested
on `atomic_load(&config->active_connections) <= config->max_connections' failed.
agroal: zeyaddaowd@localhost:5433/test: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
2026-04-17 02:17:01 INFO  main.c:2690 pgagroal: core dump requested
agroal: authenticating: /home/zeyaddaowd/Desktop/GSOC/forked/pgagroal/src/libpgagroal/pool.c:1278: pgagroal_pool_status: Assertion `atomic_load(&config->active_connections) <= config->max_connections' failed.
Aborted                    (core dumped) ./pgagroal -c pgagroal.conf -a pgagroal_hba.conf -u pgagroal_users.conf -l pgagroal_databases.conf
```

**Configuration**

* pgagroal.conf 

  pipeline = transaction
  max_connections = 1
  
* pgagroal_databases.conf
test zeyaddaowd 1 1 1

### Question / Discussion

Is this temporary overflow expected behavior (optimistic increment + rollback)?

If so, should the assertion account for this, or be removed?
or should the increment itself be bounded and safe?

### LINKED ISSUE #875: Retry path in `pgagroal_get_connection` fails to acquire freed slots under concurrent contention
**Describe the bug**

`pgagroal_get_connection()` returns `pool is full` to clients in the retry loop even when backend slots are free at the time of the retry's pool-scan. Clients consume the entire `blocking_timeout` window without ever acquiring an available slot.

This is distinct from #813 (which is about the polling cadence / latency of the 500ms retry delay). The bug here is correctness: blocked clients **fail to detect freed slots** as they become available.

**Impact**

Under concurrent saturation, clients in the retry loop can time out with `pool is full` despite the pool genuinely having free slots for many seconds. The defect is in the `retry2 → start → scan` cycle in `pgagroal_get_connection()` (pool.c).

**Reproduction**

Originally documented in [my comment on #813 from 2026-04-01](https://github.com/pgagroal/pgagroal/issues/813#issuecomment-4173218338). Brief recap:

- `max_connections = 6`, `blocking_timeout = 10`, `pipeline = session`, `ev_backend = epoll`
- `pgbench -S -c 8 -j 8 -t 100` (multi-threaded — single-threaded does **not** reproduce, because sequential connection setup avoids the concurrent contention)
- Observed (consistent across 4 CI builds):
  - T+0s: 5 clients authenticate, 3 enter the retry loop
  - T+1s: connected clients finish; pool drops to 1/6 — slots **are** available
  - T+10s: all 3 blocked clients time out with `pool is full`

The blocked clients re-scan the pool every 500ms but never acquire a freed slot.

**Suggested next step**

Investigate the race in the `retry2 → start → scan` cycle. Pattern: a blocked client wakes, calls `atomic_fetch_add(&active_connections, ...)`, scans for a `STATE_FREE` slot matching the rule, and either grabs one or returns to retry. Under concurrent contention, the scan can consistently miss slots that another worker just freed — likely because the slot is transitioning state, or because the `active_connections` precheck sees a stale value.

CI groundwork already exists (#815 added the saturation test config), so a regression test for this issue is straightforward to add.

**References**

- #813 (parent issue) — the latency-mechanism design discussion remains there
- @fluca1978 approved splitting this on #813 (2026-04-28): *"I think the (2) should be a dedicated issue and PR, and I believe also it should be completed before #813."*

fheikens:
Implements #875 (and closes #856 as a side effect — same root cause).

## The bug

`pgagroal_get_connection()` incremented `active_connections` at the entry of every attempt (`pool.c:103`) and decremented it again on the retry path. Two windows resulted:

- **Counter inflation.** Between `atomic_fetch_add` and the retry-path `atomic_fetch_sub`, the counter could exceed `max_connections`. #856 caught this with the assertion at `pool.c:1278` (in debug builds with a small max).
- **Release-ordering race.** On `pgagroal_return_connection`, the slot transitioned to `STATE_FREE` (`pool.c:515`) before the counter `fetch_sub` (`pool.c:516`). Retry clients waking in that window saw the counter still at the in-use value and bounced even though a slot was already free.

Together these caused the failure documented on #875: with `max_connections = 6`, `pgbench -c 8 -j 8 -t 100`, three clients entered the retry loop and never acquired freed slots — they all timed out at `blocking_timeout` despite 2–5 free slots being available for 9 of the 10-second window.

## The fix

Increment the counter only after a slot is actually acquired:

- The reuse loop (`pool.c:140`) does the increment (and per-rule increment, and `has_lock = true`) once `*slot = i` confirms a match.
- The new-slot loop (`pool.c:170`) does the same after the `STATE_NOTINIT → STATE_INIT` CAS succeeds.
- The entry check becomes a non-mutating `atomic_load` fast-path bail.
- The retry-path decrement is now gated on `has_lock` for both the global and per-rule counters, so it only fires when an increment actually happened.

The slot array is now the sole source of truth for the cap. The counter is a metric, observable but never load-bearing.

## What's still in scope but not addressed

- The reuse-loop CAS-then-restore pattern (briefly takes a non-matching slot, then restores) is unchanged. Less impactful with the inflation fixed; can be a separate ticket if it shows up under load.
- Polling latency (#813) is unchanged — still a 500ms `nanosleep` between retries. That's the separate design discussion fluca, jesperpedersen and I had on #813.

## Regression test

Deferred to a follow-up — see #879.

A regression test was attempted in this PR but had to be dropped: the obvious `pgbench -c N -j N` shape deadlocks during pgbench's setup phase under the performance and session pipelines, because pgbench waits for all N connections to be set up before running any transactions. The earlier saturation-test attempt in PR #832 was abandoned for the same reason. A proper test needs `pgbench -C` or a custom psql harness; that's its own design and is now tracked separately in #879.

## Closes / refs

- Closes #875
- Closes #856 (same root cause — assertion no longer reachable since the counter never speculatively exceeds the cap)
- References #813 (the latency-mechanism design discussion remains there)

### DIFF
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -63,6 +63,7 @@ static bool is_alias_of_limit(char* database, int limit_index);
 static int get_connection_count_for_limit_rule(int rule_index, char* username);
 static char* resolve_database_name(char* database, int best_rule);
 static void check_graceful_shutdown_trigger(void);
+static bool increase_connections(int best_rule);
 
 int
 pgagroal_get_connection(char* username, char* database, bool reuse, bool transaction_mode, int* slot, SSL** ssl)
@@ -100,15 +101,12 @@ start:
    do_init = false;
    has_lock = false;
 
-   connections = atomic_fetch_add(&config->active_connections, 1);
-
-   if (best_rule >= 0)
-   {
-      // increment the active connections for the current limit rule
-      atomic_fetch_add(&config->limits[best_rule].active_connections, 1);
-   }
-
-   has_lock = true;
+   /* Fast-path bail when the pool is likely saturated. The counter is
+    * read non-mutating; the slot array (below) is the source of truth.
+    * Counter inflation cannot occur because we do not increment until
+    * a slot has actually been acquired.
+    */
+   connections = atomic_load(&config->active_connections);
    if (connections >= config->max_connections)
    {
       goto retry;
@@ -139,7 +137,16 @@ start:
 
             if (can_reuse)
             {
-               *slot = i;
+               if (increase_connections(best_rule))
+               {
+                  *slot = i;
+                  has_lock = true;
+               }
+               else
+               {
+                  atomic_store(&config->states[i], STATE_FREE);
+                  goto retry;
+               }
             }
             else
             {
@@ -167,8 +174,17 @@ start:
 
          if (atomic_compare_exchange_strong(&config->states[i], &not_init, STATE_INIT))
          {
-            *slot = i;
-            do_init = true;
+            if (increase_connections(best_rule))
+            {
+               *slot = i;
+               do_init = true;
+               has_lock = true;
+            }
+            else
+            {
+               atomic_store(&config->states[i], STATE_NOTINIT);
+               goto retry;
+            }
          }
       }
    }
@@ -314,12 +330,12 @@ start:
    else
    {
 retry:
-      if (best_rule >= 0)
-      {
-         atomic_fetch_sub(&config->limits[best_rule].active_connections, 1);
-      }
       if (has_lock)
       {
+         if (best_rule >= 0)
+         {
+            atomic_fetch_sub(&config->limits[best_rule].active_connections, 1);
+         }
          atomic_fetch_sub(&config->active_connections, 1);
       }
 retry2:
@@ -1381,6 +1397,27 @@ is_alias_of_limit(char* database, int limit_index)
    }
 
    return false;
+}
+
+static bool
+increase_connections(int best_rule)
+{
+   struct main_configuration* config = (struct main_configuration*)shmem;
+   int connections;
+
+   connections = atomic_fetch_add(&config->active_connections, 1);
+   if (connections >= config->max_connections)
+   {
+      atomic_fetch_sub(&config->active_connections, 1);
+      return false;
+   }
+
+   if (best_rule >= 0)
+   {
+      atomic_fetch_add(&config->limits[best_rule].active_connections, 1);
+   }
+
+   return true;
 }
 
 static int
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -1408,6 +1408,18 @@ increase_connections(int best_rule)
    connections = atomic_fetch_add(&config->active_connections, 1);
    if (connections >= config->max_connections)
    {
+      /* Cap race: another thread acquired between the atomic_load fast-path
+       * bail in pgagroal_get_connection() and this fetch_add. Roll the
+       * global counter back so it does not drift upward as a metric. The
+       * post-decrement value is not claimed to be below max_connections
+       * (another thread may increment in between); the actual cap is
+       * enforced by the slot array (config->states[] has exactly
+       * max_connections entries), so a missed rollback at worst leaves
+       * the metric off by N rather than producing more backends than the
+       * cap.
+       */
+      pgagroal_log_debug("increase_connections: cap race (active=%d, max=%d); rolling back counter",
+                         connections + 1, config->max_connections);
       atomic_fetch_sub(&config->active_connections, 1);
       return false;
    }
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -63,6 +63,7 @@ static bool is_alias_of_limit(char* database, int limit_index);
 static int get_connection_count_for_limit_rule(int rule_index, char* username);
 static char* resolve_database_name(char* database, int best_rule);
 static void check_graceful_shutdown_trigger(void);
+static bool increase_connections(int best_rule);
 
 int
 pgagroal_get_connection(char* username, char* database, bool reuse, bool transaction_mode, int* slot, SSL** ssl)
@@ -100,15 +101,12 @@ start:
    do_init = false;
    has_lock = false;
 
-   connections = atomic_fetch_add(&config->active_connections, 1);
-
-   if (best_rule >= 0)
-   {
-      // increment the active connections for the current limit rule
-      atomic_fetch_add(&config->limits[best_rule].active_connections, 1);
-   }
-
-   has_lock = true;
+   /* Fast-path bail when the pool is likely saturated. The counter is
+    * read non-mutating; the slot array (below) is the source of truth.
+    * Counter inflation cannot occur because we do not increment until
+    * a slot has actually been acquired.
+    */
+   connections = atomic_load(&config->active_connections);
    if (connections >= config->max_connections)
    {
       goto retry;
@@ -139,7 +137,16 @@ start:
 
             if (can_reuse)
             {
-               *slot = i;
+               if (increase_connections(best_rule))
+               {
+                  *slot = i;
+                  has_lock = true;
+               }
+               else
+               {
+                  atomic_store(&config->states[i], STATE_FREE);
+                  goto retry;
+               }
             }
             else
             {
@@ -176,8 +183,26 @@ start:
 
          if (atomic_compare_exchange_strong(&config->states[i], &not_init, STATE_INIT))
          {
-            *slot = i;
-            do_init = true;
+            if (increase_connections(best_rule))
+            {
+               *slot = i;
+               do_init = true;
+               has_lock = true;
+            }
+            else
+            {
+               atomic_store(&config->states[i], STATE_NOTINIT);
+               /* The global max_connections gate lost the race after we had
+                * already reserved a backend against the per-rule max_size cap
+                * (#848). Release that reservation here; the retry: label only
+                * unwinds active_connections, so without this the reservation
+                * would leak and throttle the rule below max_size permanently. */
+               if (best_rule >= 0)
+               {
+                  atomic_fetch_sub(&config->limits[best_rule].backend_connections, 1);
+               }
+               goto retry;
+            }
          }
       }
 
@@ -341,12 +366,12 @@ start:
    else
    {
 retry:
-      if (best_rule >= 0)
-      {
-         atomic_fetch_sub(&config->limits[best_rule].active_connections, 1);
-      }
       if (has_lock)
       {
+         if (best_rule >= 0)
+         {
+            atomic_fetch_sub(&config->limits[best_rule].active_connections, 1);
+         }
          atomic_fetch_sub(&config->active_connections, 1);
       }
 retry2:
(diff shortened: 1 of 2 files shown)


src/libpgagroal/pool.c
```diff
@@ -140,6 +137,12 @@ pgagroal_get_connection(char* username, char* database, bool reuse, bool transac
             if (can_reuse)
             {
                *slot = i;
+               atomic_fetch_add(&config->active_connections, 1);
```
fluca1978: I'm wondering if we need an utility function like `bolol increase_connections( int best_rule )` that:
- checks again if the number of contains is over the limit
- `atomic_fetch_add` on `active_connections`
- if `best_rule >= 0`  then `atomic_fetch_add` on `limits`
- return true if the lock has been acquired

In this way, we can have the same behavior everywhere.
WDYT?

src/libpgagroal/pool.c
```diff
@@ -140,6 +137,12 @@ pgagroal_get_connection(char* username, char* database, bool reuse, bool transac
...
+               if (best_rule >= 0)
+               {
+                  atomic_fetch_add(&config->limits[best_rule].active_connections, 1);
```
fluca1978: should we also test if the max connections for a limit has been reached before granting the slot?

fluca1978 (commented):
See my inline comments.
Also rebase and squash where possible.

fheikens:
Thanks for the review. Pushed a rebased + squashed update addressing both inline comments.

**On the `increase_connections()` helper (line 140 comment).** Extracted as suggested. The helper does the global `atomic_fetch_add` + post-hoc cap re-check (rolling back the global counter on overshoot), then the per-rule `atomic_fetch_add`, and returns true on success. Both call sites — reuse-loop success and new-slot CAS success — now go through it; on `false`, the caller releases the slot CAS (back to `STATE_FREE` or `STATE_NOTINIT`) and `goto retry`.

**On the per-rule cap on the reuse path (line 143 comment).** I don't think the reuse path needs an additional per-rule cap check, and I'd like to leave it out unless you disagree. A reused slot is already counted toward its rule by `get_connection_count_for_limit_rule()`, which is what gates slot creation on the new-slot path. Reuse only bumps the rule's `active_connections` metric, not the slot count the cap is enforced against — so the rule's slot budget is already respected by virtue of the slot existing in the first place.

Branch is now a single commit rebased on current `master`. CI re-running.

fluca1978:
See inline comments.
We need a test for this to get in.

fheikens:
Thanks for the follow-up. Three responses:

**On the silent rollback in `increase_connections()`.** Fair point. I'll add a `pgagroal_log_debug` on the rollback path so it's observable, and add a comment in the function explaining the rollback is metric-only (see point 2). If you'd rather have an explicit out-param so the caller can branch on it, happy to do that instead — just say which.

**On "`sub` does not mean `connections < max_connections`".** Correct, and I should have made this explicit. The rollback isn't claiming the post-state is below cap — another thread could have incremented in between. Its only purpose is to keep the counter from drifting up as a metric. The actual cap is enforced by the slot array: `config->states[]` has exactly `max_connections` entries, every allocation loop iterates `0..max_connections-1`, so a missed rollback at worst leaves the metric off by N — never more backends than `max_connections`. I'll spell that out in a comment alongside the `fetch_sub`.

**On the test.** I'd like to defend the split into #879, with evidence from PR #832. The obvious shape — `pgbench -c N -j N` against a small `max_connections` — deadlocks during pgbench's setup phase under both `session` and `performance` pipelines, because pgbench waits for *all* N connections to be established before running any transaction. The first N–1 clients hold pool slots idle while the remaining block on slots that will never free. PR #832 was abandoned for exactly this. Two patterns would actually exercise the retry path:

1. **`pgbench -C`** (connect-per-transaction) — saturation manifests at the connection rate, no setup deadlock.
2. **Custom psql harness** — N parallel `BEGIN; SELECT pg_sleep(seconds); COMMIT;` clients (same shape as the #848 reproducer). Predictable, deterministic.

Either needs `pgagroal_tsclient_*` extended plus a dedicated `test/conf/` entry — non-trivial test infrastructure on top of the fix, which is why I tracked it in #879.

If you'd prefer it in this PR instead of a follow-up, which pattern: `pgbench -C` or custom psql? If the split is acceptable, I'll push the rollback log + comment now and we land #878 with #879 tracking the test.

fluca1978:
> **On the test.** I'd like to defend the split into #879, with evidence from PR #832. The obvious shape — `pgbench -c N -j N` against a small `max_connections` — deadlocks during pgbench's setup phase under both `session` and `performance` pipelines, because pgbench waits for _all_ N connections to be established before running any transaction. The first N–1 clients hold pool slots idle while the remaining block on slots that will never free. PR #832 was abandoned for exactly this. Two patterns would actually exercise the retry path:
> 

Ok, but then I would like to have #879 merged first.


> 
> Either needs `pgagroal_tsclient_*` extended plus a dedicated `test/conf/` entry — non-trivial test infrastructure on top of the fix, which is why I tracked it in #879.
> 
> If you'd prefer it in this PR instead of a follow-up, which pattern: `pgbench -C` or custom psql? If the split is acceptable, I'll push the rollback log + comment now and we land #878 with #879 tracking the test.

I don't have an opinion right now, so implement in #879 whatever is simpler and suffices to test the problem. Then it is always possible to open another issue/patch to improve the testing.

fheikens:
Pushed `1260043` addressing the rollback observation feedback:

- `pgagroal_log_debug` on the cap-race rollback in `increase_connections()` so the contention is visible instead of silent.
- Comment spelling out what the rollback is (metric hygiene) and what it is not (a claim that the post-decrement value is below `max_connections`). Notes that the slot array (`config->states[]`, sized exactly `max_connections`) is the authoritative cap, so a missed rollback at worst leaves the metric off by N rather than producing more backends than the cap.

The regression test for #875 will land on top of this PR once #879 (https://github.com/pgagroal/pgagroal/pull/887) merges — the test depends on the `pgagroal_tsclient_execute_concurrent_holds()` helper introduced there. After #879 lands I'll rebase, add `test_pgagroal_pool_saturation_retry` (10 concurrent clients = max_connections + 2 against `test/conf/17`, asserts all succeed within `blocking_timeout`), and re-request review.

fheikens:
Gentle nudge here @fluca1978. Since your last review I pushed `1260043`, which makes the cap-race rollback in `increase_connections()` observable via `pgagroal_log_debug` (addressing the silent-rollback feedback) and explained the rationale for splitting the regression test into #879/#887 so master never goes red. The two earlier inline threads are now outdated by that push. CI is green and it merges cleanly against current master — let me know if anything else is needed to land it.

fluca1978:
> Gentle nudge here @fluca1978. Since your last review I pushed `1260043`, which makes the cap-race rollback in `increase_connections()` observable via `pgagroal_log_debug` (addressing the silent-rollback feedback) and explained the rationale for splitting the regression test into #879/#887 so master never goes red. The two earlier inline threads are now outdated by that push. CI is green and it merges cleanly against current master — let me know if anything else is needed to land it.

Sorry for the delay, I'm a little busy, give me a few days to review cleanly.

fheikens:
Rebased onto current master, which now carries the concurrent-holds harness (#887) and the per-rule max_size cap (#848). Two notes on the result:

- The new-slot path now sits on top of #848's `backend_connections` reservation. When the post-acquisition global `max_connections` gate loses a race it does `goto retry`, but the `retry:` label only unwinds `active_connections` — so the per-rule reservation has to be released explicitly on that path, otherwise it leaks and throttles the rule below `max_size`. Done in `74fd00f`.
- Added the promised regression test `test_pgagroal_pool_saturation_retry` (`e7366a0`). It drives `max_connections + 2` concurrent holds; the two clients past the cap must acquire slots freed by the first batch within `blocking_timeout`. It derives the load from the live config and skips profiles whose `blocking_timeout` is too small to isolate the regression, so it is meaningful across all run-configs profiles and exercises `conf/17`.

Force-pushed (rebase rewrote the two earlier commits). CI re-running.

jesperpedersen:
@fheikens Remember `Squash` / `Force push`

fheikens:
Squashed to a single commit and force-pushed. Thanks both.

fluca1978:
merged, thanks @fheikens

Issue #879, opened 2026-04-28
Regression test for #875 (retry-path slot acquisition)


fheikens:
#875 fixed the bug in `pgagroal_get_connection`'s retry path. A regression test was attempted in the same PR (#878) but had to be dropped: the obvious shape (`pgbench -c N -j N`) deadlocks during pgbench's setup phase under the performance and session pipelines, since pgbench waits for all N connections to be set up before running any transactions, so the first N-1 clients hold pool slots idly while the remaining clients wait on slots that will not free. The earlier saturation-test attempt in PR #832 was abandoned for the same reason.

Need a test pattern that actually exercises retry-path acquisition under saturation. Two candidates:

1. **pgbench `-C`** (connect per transaction). Each transaction opens a new connection. Saturation manifests at the connection rate, not as a setup deadlock.
2. **Custom psql harness** — N parallel `BEGIN; SELECT pg_sleep(seconds); COMMIT;` clients (the same shape used in the #848 Bug A reproducer). Predictable hold time per client, deterministic saturation, doesn't depend on pgbench's setup semantics.

Either pattern needs to extend `pgagroal_tsclient_*` to support the new flag/path, plus a dedicated `test/conf/` entry with a small `max_connections` (or per-rule `max_size`) so saturation actually manifests.

No urgency — the fix has landed; this is closing a loose end.

PR #887, opened 2026-05-06
[#879] Add concurrent-holds test harness

fheikens:
Implements the test infrastructure for #879.

## What's in this PR

- **`pgagroal_tsclient_execute_concurrent_holds(user, database, client_count, hold_seconds)`** in `test/libpgagroaltest/tsclient.c`. Spawns `client_count` pthreads, each running `psql -c "BEGIN; SELECT pg_sleep(hold_seconds); COMMIT;"` through pgagroal, and returns 0 only if every session returned success. Uses pthreads + `system(psql)` to match the existing tsclient style — no new build dependencies.
- **`test/conf/17/`** — `max_connections=6`, `blocking_timeout=10`, performance pipeline. Gives `run-configs` mode a deterministic small-cap environment for saturation scenarios.
- **`test_pgagroal_pool_concurrent_smoke`** — two concurrent holds, 1s `pg_sleep` each, well under any config's `max_connections`. Validates the helper end-to-end without depending on retry-path behaviour.

## Why the split with #878

A regression test for #875 by definition fails on master pre-fix. Landing it alone would leave master red.

Per [@fluca1978's comment on PR #878](https://github.com/pgagroal/pgagroal/pull/878#issuecomment-4386278575) ("I would like to have #879 merged first"), this PR lands first and contains only what's safe to ship without the fix: the harness, the config, and the smoke test that exercises the harness. The actual retry-path regression test (`test_pgagroal_pool_saturation_retry`, N+2 clients exceeding the cap) is added in #878 alongside the fix, so master is never red.

## Refs

- References #875 (the bug being tested)
- References #878 (the companion fix + retry-path test)
- Issue #879 stays open until #878 lands the saturation test

### DIFF
--- a/test/conf/17/pgagroal.conf
+++ b/test/conf/17/pgagroal.conf
@@ -1,0 +1,22 @@
+[pgagroal]
+host = localhost
+port = 2345
+
+log_type = file
+log_level = debug5
+log_path = test.log
+
+# Short blocking_timeout makes the retry-path regression for #875
+# fail quickly when present: clients exceeding max_connections must
+# acquire freed slots before blocking_timeout fires.
+max_connections = 8
+blocking_timeout = 10
+idle_timeout = 600
+validation = off
+unix_socket_dir = /tmp/
+pipeline = performance
+
+
+[primary]
+host = localhost
+port = 5432
--- a/test/conf/17/pgagroal_hba.conf
+++ b/test/conf/17/pgagroal_hba.conf
@@ -1,0 +1,4 @@
+#
+# TYPE  DATABASE USER  ADDRESS  METHOD
+#
+host    all all all trust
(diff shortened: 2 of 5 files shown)


fheikens:
@fluca1978 ready for review when you have a moment. This is the test-infrastructure half of #879 (helper API + `test/conf/17` + smoke test); the actual retry-path regression test for #875 is added together with the fix in #878 so master never goes red. CI green across all jobs.

fheikens:
Nudging this for review @fluca1978 — this is the test-infrastructure half of #879 (helper API + `test/conf/17` + smoke test), and the actual retry-path regression test for #875 rides along with the fix in #878. CI is green across all jobs and it merges cleanly against current master.

fluca1978:
@fheikens can you squash please? I would do, but I prefer you to meld the commit messages.

fheikens:
Squashed to a single commit with the commit messages melded, as requested.

fluca1978:
merged, thanks @fheikens

PR #889, opened 2026-05-10
[#843] Handle server identifier validation on conf reload

### LINKED ISSUE #843: Run server identifier validation on conf reload, not just at startup
Right now, `startup_validation` only runs when pgagroal starts. If you add a new server and run `pgagroal-cli conf reload`, the check doesn’t run again - which means you could accidentally configure two servers pointing to the same PostgreSQL cluster without any warning.

We probably don’t want to fail or stop pgagroal on reload like we do at startup, but we should still catch the issue and report it.

A simple approach could be:

* On reload if servers are added or modified, re-run the identifier check across all configured servers.
* If `startup_validation` itself changes (e.g. from `off` to `on`), run the check immediately.
* If a duplicate is found at runtime, log a warning or error instead of shutting down, since there may already be active connections.
* When using `conf set` - to update server fields like `host` or `port`, trigger the check again.

Abdelrhmansersawy:
resolved #843 

## Changes

1. Added new field attribute `valid` to `struct server`, and define new macros `FOREACH_SERVER`, `FOREACH_VALID_SERVER`, `FOREACH_INVALID_SERVER` .
2. Implemented new utilities `pgagroal_server_is_primary(srv_idx)` , and `has_active_connections(srv_idx)`
3. Add a new mechanism func `determine_loser` to detect invalid server between two ones based on rules 
   1. If one server has active connections, keep it
   2. If one server is PRIMARY, keep it
   3. Otherwise keep the lower index (first in config order)
4. Extend `pgagroal_check_server_identifiers` to have parameter `bool strict` to just warn in runtime server validation and mark invalid server (no stop for pgagroal)

### DIFF
--- a/src/include/pgagroal.h
+++ b/src/include/pgagroal.h
@@ -388,6 +388,7 @@ struct server
    int minor_version;            /**< The minor version of the server */
    char system_identifier[64];   /**< The system identifier of the server */
    bool tls;                     /**< Use TLS if possible */
+   bool valid;                   /**< Is the server valid */
    char tls_cert_file[MAX_PATH]; /**< TLS certificate path */
    char tls_key_file[MAX_PATH];  /**< TLS key path */
    char tls_ca_file[MAX_PATH];   /**< TLS CA certificate path */
@@ -397,6 +398,14 @@ struct server
    atomic_schar auth_type;       /**< The authentication type used for health check */
    int lineno;                   /**< The line number within the configuration file */
 } __attribute__((aligned(64)));
+
+#define FOREACH_SERVER for (int i = 0; i < config->number_of_servers; i++)
+#define FOREACH_VALID_SERVER                           \
+   for (int i = 0; i < config->number_of_servers; i++) \
+      if (config->servers[i].valid)
+#define FOREACH_INVALID_SERVER                         \
+   for (int i = 0; i < config->number_of_servers; i++) \
+      if (!config->servers[i].valid)
 
 /** @struct connection
  * Defines a connection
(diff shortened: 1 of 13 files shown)


Abdelrhmansersawy:
Once we finalize the runtime server validation structure, I’ll start working on adding more test cases and improving the docs.

Abdelrhmansersawy:
I'm thinking about should we not call `system_identifers` not-changed servers or it's safe to call at each one ?

src/include/pgagroal.h
```diff
@@ -398,6 +399,14 @@ struct server
    int lineno;                   /**< The line number within the configuration file */
 } __attribute__((aligned(64)));
 
+#define FOREACH_SERVER for (int i = 0; i < config->number_of_servers; i++)
```
fluca1978, line 402: what if we introduce variables in these macros?

```
#define FOREACH_SERVER(i, current) \
    for (int i = 0; i < num_servers && ((current) = servers[i], 1); i++)
```

this way the usage would be something like:

```
FOREACH_SERVER(i, current ) {
   if ( current.valid ) ....
}
```

fluca1978 (commented):
LGTM, @jesperpedersen what do you think about the macros? Do we need arguments or do they suffice as they are?

fluca1978:
> I'm thinking about should we not call `system_identifers` not-changed servers or it's safe to call at each one ?

Not clear, but I would leave the call for everyone server.

jesperpedersen:
@fluca1978 I think the macros are ok

fluca1978:
merged, thanks @Abdelrhmansersawy

Issue #844, opened 2026-04-08
PAUSE / RESUME for Connection Pools


Abdelrhmansersawy:
Right now, pgagroal can’t temporarily stop sending new connections without dropping clients. It would be great to have `PAUSE` and `RESUME` commands, like PgBouncer and PgCat, so admins can hold traffic at the pooler during maintenance or failover.

When paused:

* New connections are accepted but **wait at the pooler**
* Existing queries continue normally
* `RESUME` lets waiting clients continue

This helps with maintenance, failovers, overloads, and config reloads - all without causing client errors.

Currently, `blocking_timeout` only rejects clients after a set limit, whereas PAUSE would **hold them indefinitely** until the admin resumes - or temporarily if a `pause_timeout` is configured.

```bash
pgagroal-cli pause
pgagroal-cli pause <server>

pgagroal-cli resume
pgagroal-cli resume <server>
```

`pgagroal-cli status` should show which servers are currently paused.

fluca1978:
This is very interesting, but I think that once in pause mode, existing connections from `pgagroal` to PostgreSQL should be flushed, returning to the pool as soon as possible.

With regard to the command line, I think we should provide `pgagroal-cli <pause|resume> [server|all]`.
It could be interesting to keep also a per-server `last_paused` and `last_resumed` timestamps to show on the command output, so that admins could also gather information about how much time a system has been paused.

And last but not least, exporting data to prometheus too.

AbduAllahhany:
If this is still available, I’d like to take it.

Abdelrhmansersawy:
@AbduAllahhany 
Sorry, I'm currently working on it.
You can navigate issue tabs.

Thanks !

jesperpedersen:
@AbduAllahhany @Abdelrhmansersawy You can work together - break each others code with test cases

Abdelrhmansersawy:
@jesperpedersen 
No problem !
@AbduAllahhany 
What currently am I looking to see at how pgbouncer handle this
https://github.com/pgbouncer/pgbouncer/blob/master/src/admin.c#L1270

We need first to keep client connection on pooler, https://github.com/pgagroal/pgagroal/issues/844#issuecomment-4212444669.

Feel free to share ideas

Abdelrhmansersawy:
@AbduAllahhany 
Once we have full picture about solution, we can break them on smaller PR.

After holiday, I would share my ideas.

jesperpedersen:
[...]

PR #890, opened 2026-05-11
[#844] Add pause/resume server feature

AbduAllahhany:
resolved [#844]
# Add pause/resume for PostgreSQL servers in the pooler
  ## changes
  Adds **per-server (or all-servers) pause and resume** at the pooler so traffic and new assignments to one or more PostgreSQL backends can be stopped without shutting down pgagroal, 
  then restored with **prefill**. Pause state is visible in **status/JSON**, and successful pause/resume operations are counted in **Prometheus**.
  ### Behavior
  - **`pgagroal-cli pause [server]`** and **`pgagroal-cli resume [server]`** — optional server name; `*` or omission for **all** servers.
  - **Pause:** waits for in-use slots to clear, **flushes** backend connections for affected servers, and **blocks** new slot assignment and connects until resume.
  - **Resume:** clears pause state, runs **prefill**, and restores normal pooling.
  - **Status / JSON:** `AllPaused`, per-server `Paused`, `LastPaused`, `LastResumed`.
  - **Management errors:** unknown server and fork failure (and aligned resume errors as defined in `management.h`).
  - **Prometheus:** `pgagroal_server_pause_total` and `pgagroal_server_resume_total` per server.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1851,6 +1851,261 @@ accept_mgt_cb(struct io_watcher* watcher)
 
       pgagroal_management_response_ok(NULL, client_fd, start_time, end_time, compression, encryption, payload);
    }
+   else if (id == MANAGEMENT_PAUSE)
+   {
+      struct json* req = NULL;
+      struct json* res = NULL;
+      struct json* servers = NULL;
+      char* server = NULL;
+      time_t now;
+      int flush_server = -1;
+
+      pgagroal_log_debug("pgagroal: Management pause: ");
+      pgagroal_pool_status();
+
+      start_time = time(NULL);
+
+      req = (struct json*)pgagroal_json_get(payload, MANAGEMENT_CATEGORY_REQUEST);
+      server = (char*)pgagroal_json_get(req, MANAGEMENT_ARGUMENT_SERVER);
+
+      pgagroal_management_create_response(payload, -1, &res);
+      pgagroal_json_create(&servers);
+
+      if (!strcmp("*", server))
+      {
+         struct json* js = NULL;
+
+         now = time(NULL);
+         config->all_paused = true;
+         for (int i = 0; i < config->number_of_servers; i++)
+         {
+            if (strlen(config->servers[i].name) > 0)
+            {
+               config->servers[i].paused = true;
+               config->servers[i].last_paused = now;
+               bool all_done;
+               do
+               {
+                  all_done = true;
+                  for (int j = 0; j < config->max_connections; j++)
+                  {
+                     if (config->connections[j].server == i &&
+                         atomic_load(&config->states[j]) == STATE_IN_USE)
+                     {
+                        all_done = false;
+                        break;
+                     }
+                  }
+                  if (!all_done)
+                  {
+                     SLEEP(500000000L);
+                  }
+               }
+               while (!all_done);
+
+               /* Flush all existing backend connections gracefully */
+               pid = fork();
+               if (pid == -1)
+               {
+                  config->all_paused = false;
+                  pgagroal_management_response_error(NULL, client_fd, NULL, MANAGEMENT_ERROR_PAUSE_NOFORK, compression, encryption, payload);
+                  pgagroal_log_error("Pause: No fork (%d)", MANAGEMENT_ERROR_PAUSE_NOFORK);
+                  pgagroal_json_destroy(servers);
+                  pgagroal_json_destroy(res);
+                  goto error;
+               }
+               else if (pid == 0)
+               {
+                  pgagroal_flush_server(i);
+               }
+               else
+               {
+                  pgagroal_prometheus_server_pause(i);
+               }
+            }
+            pgagroal_json_create(&js);
+
+            pgagroal_json_put(js, MANAGEMENT_ARGUMENT_SERVER, (uintptr_t)server, ValueString);
+            pgagroal_json_put(js, MANAGEMENT_ARGUMENT_RESUMED, (uintptr_t)false, ValueBool);
+
+            pgagroal_json_append(servers, (uintptr_t)js, ValueJSON);
+         }
+      }
+      else
+      {
+         bool found = false;
+
+         for (int i = 0; !found && i < config->number_of_servers; i++)
+         {
+            struct json* js = NULL;
+
+            if (strlen(config->servers[i].name) == 0)
+            {
+               continue;
+            }
+
+            if (!strcmp(config->servers[i].name, server))
+            {
+               now = time(NULL);
+               config->servers[i].paused = true;
+               config->servers[i].last_paused = now;
+               flush_server = i;
+
+               bool all_done = false;
+               while (!all_done)
+               {
+                  all_done = true;
+                  for (int j = 0; j < config->max_connections; j++)
+                  {
+                     if (config->connections[j].server == i &&
+                         atomic_load(&config->states[j]) == STATE_IN_USE)
+                     {
+                        all_done = false;
+                        break;
+                     }
+                  }
+                  if (!all_done)
+                  {
+                     SLEEP(500000000L);
+                  }
+               }
+
+               pgagroal_json_create(&js);
+
+               pgagroal_json_put(js, MANAGEMENT_ARGUMENT_SERVER, (uintptr_t)server, ValueString);
+               pgagroal_json_put(js, MANAGEMENT_ARGUMENT_RESUMED, (uintptr_t)false, ValueBool);
+
+               pgagroal_json_append(servers, (uintptr_t)js, ValueJSON);
+               found = true;
+            }
+         }
+
+         if (!found)
+         {
+            pgagroal_management_response_error(NULL, client_fd, NULL, MANAGEMENT_ERROR_PAUSE_UNKNOWN_SERVER, compression, encryption, payload);
+            pgagroal_log_error("Pause: Unknown server %s", server);
+            pgagroal_json_destroy(servers);
+            pgagroal_json_destroy(res);
+            goto error;
+         }
+
+         /* Flush this server's existing backend connections gracefully */
+         pid = fork();
+         if (pid == -1)
+         {
+            config->servers[flush_server].paused = false;
+            pgagroal_management_response_error(NULL, client_fd, NULL, MANAGEMENT_ERROR_PAUSE_NOFORK, compression, encryption, payload);
+            pgagroal_log_error("Pause: No fork (%d)", MANAGEMENT_ERROR_PAUSE_NOFORK);
+            pgagroal_json_destroy(servers);
+            pgagroal_json_destroy(res);
+            goto error;
+         }
+         else if (pid == 0)
+         {
+            pgagroal_flush_server((int_fast8_t)flush_server);
+         }
+         else
+         {
+            pgagroal_prometheus_server_pause((int)flush_server);
+         }
+      }
+
+      pgagroal_json_put(res, MANAGEMENT_ARGUMENT_SERVERS, (uintptr_t)servers, ValueJSON);
+
+      end_time = time(NULL);
+
+      pgagroal_management_response_ok(NULL, client_fd, start_time, end_time, compression, encryption, payload);
+   }
+   else if (id == MANAGEMENT_RESUME)
+   {
+      struct json* req = NULL;
+      struct json* res = NULL;
+      struct json* servers = NULL;
+      char* server = NULL;
+      time_t now;
+
+      pgagroal_log_debug("pgagroal: Management resume: ");
+      pgagroal_pool_status();
+
+      start_time = time(NULL);
+
+      req = (struct json*)pgagroal_json_get(payload, MANAGEMENT_CATEGORY_REQUEST);
+      server = (char*)pgagroal_json_get(req, MANAGEMENT_ARGUMENT_SERVER);
+
+      pgagroal_management_create_response(payload, -1, &res);
+      pgagroal_json_create(&servers);
+
+      if (!strcmp("*", server))
+      {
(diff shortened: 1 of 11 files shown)


jesperpedersen:
@AbduAllahhany Please see DEVELOPERS.md. You need to add yourself to the authors files, and squash your commits

jesperpedersen:
@Abdelrhmansersawy Please, go through this as well

Abdelrhmansersawy:
@AbduAllahhany 
Remember
https://github.com/pgagroal/pgagroal/blob/master/AUTHORS

The Commit message should start with `[#issue_id] ...`

Abdelrhmansersawy:
@AbduAllahhany 
Great work, I'd give you a reivew tommorw at morning.

src/main.c
```diff
@@ -1851,6 +1851,261 @@ accept_mgt_cb(struct io_watcher* watcher)
...
+
+      start_time = time(NULL);
+
+      req = (struct json*)pgagroal_json_get(payload, MANAGEMENT_CATEGORY_REQUEST);
+      server = (char*)pgagroal_json_get(req, MANAGEMENT_ARGUMENT_SERVER);
+
+      pgagroal_management_create_response(payload, -1, &res);
+      pgagroal_json_create(&servers);
+
+      if (!strcmp("*", server))
```
Abdelrhmansersawy: I think it's better to be `all` intead of `*`, so it would look like:
`pgagroal-cli <pause|resume> [server|all]`

src/main.c
```diff
@@ -1851,6 +1851,261 @@ accept_mgt_cb(struct io_watcher* watcher)
...
```
Abdelrhmansersawy: same

src/main.c
```diff
@@ -1851,6 +1851,261 @@ accept_mgt_cb(struct io_watcher* watcher)
...
+                        all_done = false;
+                        break;
+                     }
+                  }
+                  if (!all_done)
+                  {
+                     SLEEP(500000000L);
+                  }
+               }
+               while (!all_done);
```
Abdelrhmansersawy: I’m a bit concerned that these iterations could take a long time, especially with long-running transactions.

I think we could introduce something like `suspend_timeout` - the maximum time the server waits for an active client to finish before forcing a pause.

We could reuse `suspend_timeout` for other commands to prevent hanging on long-running queries:
1. `pgagroal-cli shutdown gracefully`
2. `pgagroal-cli reload`
3. Later, we could extend this to failover / HA switchover

If we go with this approach, we should implement it in a separate PR.

@jesperpedersen @fluca1978 thoughts?

src/main.c
```diff
@@ -1851,6 +1851,261 @@ accept_mgt_cb(struct io_watcher* watcher)
...
+
+            if (!strcmp(config->servers[i].name, server))
+            {
+               now = time(NULL);
+               config->servers[i].paused = true;
+               config->servers[i].last_paused = now;
+               flush_server = i;
+
+               bool all_done = false;
+               while (!all_done)
```
Abdelrhmansersawy: Same

Abdelrhmansersawy (changes requested):
Let’s start with these and tackle it incrementally.

src/main.c
fluca1978: > I think it's better to be `all` intead of `*`, so it would look like: `pgagroal-cli <pause|resume> [server|all]`

I like the `all` suggestion, makes also less noise in shell (`*` could be interpeloted if not escaped).

src/main.c
fluca1978: > I’m a bit concerned that these iterations could take a long time, especially with long-running transactions.
> 
> I think we could introduce something like `suspend_timeout` - the maximum time the server waits for an active client to finish before forcing a pause.
> 
> We could reuse `suspend_timeout` for other commands to prevent hanging on long-running queries:
> 
>     1. `pgagroal-cli shutdown gracefully`
> 
>     2. `pgagroal-cli reload`
> 
>     3. Later, we could extend this to failover / HA switchover
> 
> 
> If we go with this approach, we should implement it in a separate PR.
> 
> @jesperpedersen @fluca1978 thoughts?

I like the idea, I would call it `flush_timeout` to indicate that is the time to wait for a flush to happen. Sounds to me like this is what we need in all the mentioned commands too.
I also agree that we need this as a separate issue/PR related to this one.

Abdelrhmansersawy: Which option do you think is better?

1. Implement it now in a separate PR, then continue this one.

or

2. Finish and merge this PR first, then add it later as a common enhancement for all commands.

fluca1978:
The basic idea seems good to me, I think we should split this into two parts, as suggested by @Abdelrhmansersawy : we need a way to "flush" existing connections in an organized way and within a timeout (hence `flush_timeout` or alike). Then on top of that we can build the functionality of the pause/resume commands already in here.

However, I've a doubt in the semantic of this: `pause` essentially wants to interrupt the connections and `resume` does another prefill. What makes this different from `enable` and `disable` commands at this point?

src/main.c
Abdelrhmansersawy: > > I think it's better to be `all` intead of `*`, so it would look like: `pgagroal-cli <pause|resume> [server|all]`
> 
> I like the `all` suggestion, makes also less noise in shell (`*` could be interpeloted if not escaped).

We should keep commands consistent by using `all` instead of `*`.
For example, `pgagroal-cli flush` still uses `*`.
I think we should refactor all commands to use all. WDYT?

Abdelrhmansersawy:
> However, I've a doubt in the semantic of this: `pause` essentially wants to interrupt the connections and `resume` does another prefill. What makes this different from `enable` and `disable` commands at this point?

I think `disable / enable` operates at the **database level** - you disable a specific database (or all databases) from being accessed by clients.

While `pause / resume` operates at the **backend server level** - you pause traffic going to a specific PostgreSQL instance (or all instances).

Do you think we can merge both into something like:

```
pgagroal-cli enable|disable server|database <name>|all
```

Or should we keep them as two separate commands?

However, I have some doubt about handling `database | backend` levels. It feels like we are starting to have two separate commands, each dealing with a different level - so maybe they should stay separate.

On the other hand, users might not like having multiple commands. Fewer commands with more options could be a better direction. What do you think?

Abdelrhmansersawy:
As we would user `flush_timeout` here I think we need to give DBAs more control (we don't need to force this behavior to him) - I think we can get some idea from `pgagroal-cli flush [gracefully|timeout|all]`

we can have also ` pgagroal-cli pause [gracefully|timeout|all] <server>` 

*   **`gracefully`**: Wait indefinitely for active clients to finish.
*   **`all`**: Instantly kill all active and idle connections.
*   **`timeout`**: Wait for clients to finish, but forcefully kill them if they exceed the configured `flush_timeout`.

AbduAllahhany:
And when paused the the new connection will have slot in the pool but in disable it will reject the connection in pgagroal_authenticate

Abdelrhmansersawy:
> And when paused the the new connection will have slot in the pool but in disable it will reject the connection in pgagroal_authenticate

Agreed. I think DBAs use `pause | resume` for migrations or server emergencies, so they usually handle the situation very fast. Because of this, it makes sense to flush the existing connections and let the new ones wait in the pool. If a connection is stuck in the pool for too long, we control that with `blocking_timeout` to eventually return an error to the user.

AbduAllahhany:
> Agreed. I think DBAs use `pause | resume` for migrations or server emergencies, so they usually handle the situation very fast. Because of this, it makes sense to flush the existing connections and let the new ones wait in the pool. If a connection is stuck in the pool for too long, we control that with `blocking_timeout` to eventually return an error to the user.

I was a bit concerned when long running queries is flushed and the frontend side does not rerun it again after resumed so I decided to wait indefinitely but we can add ` pgagroal-cli pause [gracefully|idle|all] <server>`  as you mentioned ,It will be better

src/main.c
fluca1978: > We should keep commands consistent by using `all` instead of `*`. For example, `pgagroal-cli flush` still uses `*`. I think we should refactor all commands to use all. WDYT?


yeah, you were faster than me looking at commands.
I would start in here with `all`, and open an issue to remove the `*` (updating documentation, tests, ...).

fluca1978:
> > However, I've a doubt in the semantic of this: `pause` essentially wants to interrupt the connections and `resume` does another prefill. What makes this different from `enable` and `disable` commands at this point?
> 
> I think `disable / enable` operates at the **database level** - you disable a specific database (or all databases) from being accessed by clients.
> 

So far, we have `disable all` that sounds like `pause` with connection rejection (while `pause` hold them).
Assume we need to do maintenance on a server, are we going to use `pause` or `disable all`? I would use `disable all` to prevent new incoming connections, while `pause` does sense to me for a short-time stuff, like "wait a little longer and we'll be back". In this scenario, I think `pause` should not flush connections and then refill, simply traffic should not be routed during the paused window. 

It could be I'm totally looking at this from the wrong perspective.
Thoughts?

jesperpedersen:
This is one of the extremely hard ones.

And, long running transactions is precisely the reason.

So, having a policy on the `PAUSE` is mandatory.

The main reason for `PAUSE`/`RESUME` is that the cluster being updated. So, all sessions will go away from the server side. It doesn't have to be the case - suspend all client application such that we can run analysis directly on the cluster.

So, `PAUSE`/`RESUME` need to be able to cancel sessions, and reestablish credentials for the same client socket descriptor.

The main problem is that there is typically a connection pool - like IronJacamar - above pgagroal that will keep sessions alive, so we will need to look at the state of each of the connections - with the view of policy passed

jesperpedersen:
A long running transaction is long running for a reason - it was "designed" that way.

So, there are metrics around this, we need to be able to report when the pool is fully in `PAUSE` mode. Remember, this could take days...

Is that acceptable for a paused pool that doesn't accept new connections ? And, all idle sessions were removed ?

jesperpedersen:
@AbduAllahhany This is will require a lot documentation with pros/cons for each policy

jesperpedersen:
`PAUSE` is meant to keep the client socket descriptor.

You can not nuke the process handling it

jesperpedersen:
The key for this feature - for some of the policies - is

CancelRequest @ https://www.postgresql.org/docs/devel/protocol-message-formats.html

Abdelrhmansersawy:
@AbduAllahhany Are you ok with this ?  we can improve together.

AbduAllahhany:
> @AbduAllahhany Are you ok with this ? we can improve together.

I currently working on it I added at worker.c two signal one for resume and other for pause that stop and start io_watcher.
there is two mode 
- one gracefully that waitng all queries to finish 
- all send cancel request but keeping client fd . 

the issue here that all session variable is gone I think we should store it before pause or smth else,
Now, I am  testing it for all cases. Sorry I was late due to my final exams.

Abdelrhmansersawy:
> > @AbduAllahhany Are you ok with this ? we can improve together.
> 
> I currently working on it I added at worker.c two signal one for resume and other for pause that stop and start io_watcher. there is two mode
> 
> * one gracefully that waitng all queries to finish
> * all send cancel request but keeping client fd .
> 
> the issue here that all session variable is gone I think we should store it before pause or smth else, Now, I am testing it for all cases. Sorry I was late due to my final exams.

Good luck with your exams !

I think we need to finish this issue  #891 before going with current one

fluca1978:
> I think we need to finish this issue #891 before going with current one

I think it too, so it is better to coordinate to evolve in that direction.
@AbduAllahhany it would be also nice to sketch the approach you have in mind, so we can help here if needed.

AbduAllahhany:
@fluca1978 @Abdelrhmansersawy  @jesperpedersen 
I registered in `worker.c` two watchers `pause_watcher` and `resume_watcher`  with `pause_callback` and `resume_callback` that's break the event loop for
- pasue cammand pauses the `client_io.io`  
-  resume command resume the  `client_io.io`  and `reauth` if the backend fd is not valid any more. 

So for that I added two function in `ev.c`  `pgagroal_io_pause` and `pgagroal_io_resume` that tells the kernel to stop monitoring the fd without closing fd or watcher destruction. 
and in `security.c` I added reauth fuction so we can reauth the slot 

so genrally pause can be like this ` pgagroal-cli pause [gracefully|all] <server>`
- all : send cancel request to current quries and pause the watcher  immediately (even if there is a pooler on top of pgagroal like `pgxpool`
- gracefully : wait the current queries  to finish and add in `pgagroal-cli status` the current pause status if all connection is paused or not.

There is something I didn't get it why we introduce `time_out` as gracefully meant to wait.

If anything I got it wrong please tell me.

Abdelrhmansersawy:
> * all : send cancel request to current quries and pause the watcher  immediately (even if there is a pooler on top of pgagroal like `pgxpool`
> * gracefully : wait the current queries  to finish and add in `pgagroal-cli status` the current pause status if all connection is paused or not.

I think we should have another option like `time_out`, as I discussed here https://github.com/pgagroal/pgagroal/pull/890#issuecomment-4438927724

Think of long-running transaction

AbduAllahhany:
> > * all : send cancel request to current quries and pause the watcher  immediately (even if there is a pooler on top of pgagroal like `pgxpool`
> > * gracefully : wait the current queries  to finish and add in `pgagroal-cli status` the current pause status if all connection is paused or not.
> 
> I think we should have another option like `time_out`, as I discussed here [#890 (comment)](https://github.com/pgagroal/pgagroal/pull/890#issuecomment-4438927724)
> 
> Think of long-running transaction

I will work on this first

Abdelrhmansersawy:
@AbduAllahhany No need to close this PR, we just wait until finish #891. Then back continue work here

PR #893, opened 2026-05-14
[#888] Add High Availability chapter to the manual

### LINKED ISSUE #888: Introduce High Availability (HA) Chapter
Introduce `21-ha.md` chapter - and a `HA.md`

Abdelrhmansersawy:
resolved #888 

## Changes

1. Added `doc/HA.md` as a standalone HA reference covering all four topology strategies with minimal config snippets.

2. Added `doc/manual/en/21-ha.md` -- a new **High Availability** chapter in the manual covering:
   - Two replication topologies: **Linear (Cascading)** and **Star**
   - Four configuration strategies: Linear Backup, Linear Partner, Star Single Level, Star Two Level
   - Each strategy documented with introduction, pros/cons, PostgreSQL configuration, and HAProxy configuration
   - Unified health check setup using a single `pgagroal_health` role for both pgagroal and HAProxy probes
   - Best practices section with scenario-based topology recommendations

3. Added 8 topology diagrams to `doc/images/` covering each configuration strategy at both the PostgreSQL replication level and the HAProxy+pgagroal routing level.

4. Updated `pgagroal.spec` to include `HA.md` and the new diagram images in the RPM package.


doc/manual/en/21-ha.md
```diff
@@ -0,0 +1,415 @@
...
+| Read-write | `10.0.0.21:5432` | `10.0.0.11:2345`                           |
+| Read-only  | `10.0.0.22:5433` | `10.0.0.12:2345`, `10.0.0.13:2345` |
+
+## Health Check
+
+[HAProxy][haproxy] probes each [**pgagroal**][pgagroal] instance to decide whether to route traffic to it.
+[**pgagroal**][pgagroal] probes the [PostgreSQL][postgresql] backend directly.
+[HAProxy][haproxy] does not contact [PostgreSQL][postgresql] directly.
+
+A single role -- `pgagroal_health` -- is used for both layers, which simplifies setup and credential management.
```
Abdelrhmansersawy: I was thinking: why don’t we use a single user role to handle both (HAProxy , pgagroal) health checks instead of having separate ones? I think it would be better.

What do you think?

doc/manual/en/21-ha.md
```diff
@@ -0,0 +1,415 @@
...
+    stats refresh 5s
+```
+
+**HAProxy read-only configuration** (`haproxy-ro.cfg`):
+
+[HAProxy][haproxy] supports several load balancing algorithms for the read backend:
+
+| Algorithm    | Behavior                                              | Use when                                  |
+|--------------|-------------------------------------------------------|-------------------------------------------|
+| `leastconn`  | Routes to the replica with the fewest active connections | Connection lifetimes vary -- **recommended** for pgagroal |
```
Abdelrhmansersawy: What do you see that best load balance algorithms we should recommend to use now ?

I'm not sure if the least active connections is the better, so WDYT ?

doc/manual/en/21-ha.md
jesperpedersen: Ideally, the front-end layer should use the Prometheus metric for health. However, we can go with a single role for now

doc/manual/en/21-ha.md
jesperpedersen: We can have multiple policies

* Least connections
* Round-robin
* Ratio

doc/HA.md
```diff
@@ -0,0 +1,118 @@
+# pgagroal high availability
+
+[pgagroal][pgagroal] achieves high availability by sitting in front of [PostgreSQL][postgresql]
+streaming replication and coordinating with [HAProxy][haproxy] to route client traffic.
+
+Two replication topologies are supported:
+
+## Linear (Cascading)
+
+Replicas replicate in a chain:
```
fluca1978, line 10: we should use the term `standby` for a replica. The terms used in the documentation are `primary` and `standby`, unluckily this requires to change the pictures, so I'm asking if it is worth (now).
@jesperpedersen toughts?

doc/HA.md
```diff
@@ -0,0 +1,118 @@
...
+Primary -> Replica 1 -> Replica 2
+```
+
+### Backup
+
+Replica 2 acts as a spare. HAProxy routes reads to Replica 1 under normal conditions
+and promotes Replica 2 only if Replica 1 fails.
+
+```
+backend be_pgagroal_replicas
```
fluca1978, line 22: this is an excerpt of haproxy configuration, right?

doc/HA.md
```diff
@@ -0,0 +1,118 @@
...
+       Replica 1   Replica 2
+       /     \     /      \
+ Replica1-1 Replica1-2  Replica2-1  Replica2-2
+```
+
+Second-level replicas use `track` so HAProxy removes them from rotation
+when their upstream replica fails.
+
+```
+backend be_pgagroal_replicas
```
fluca1978, line 81: we should indicate this snippets to what they belong

doc/HA.md
```diff
@@ -0,0 +1,118 @@
...
+    server pgagroal-replica2   10.0.0.13:2345 check inter 3s rise 2 fall 3
+    server pgagroal-replica1-1 10.0.0.14:2345 track be_pgagroal_replicas/pgagroal-replica1
+    server pgagroal-replica1-2 10.0.0.15:2345 track be_pgagroal_replicas/pgagroal-replica1
+    server pgagroal-replica2-1 10.0.0.16:2345 track be_pgagroal_replicas/pgagroal-replica2
+    server pgagroal-replica2-2 10.0.0.17:2345 track be_pgagroal_replicas/pgagroal-replica2
+```
+
+## Health Check
+
+A single role is used for both pgagroal and HAProxy health probes:
```
fluca1978, line 94: `A single PostgreSQL role is used [...], health probes happen in an ad-hoc database created as shown below` or something like this.

doc/HA.md
```diff
@@ -0,0 +1,118 @@
...
+
+```sql
+CREATE ROLE pgagroal_health WITH LOGIN PASSWORD 'your_secure_password' CONNECTION LIMIT 1;
+CREATE DATABASE pgagroal_health WITH OWNER pgagroal_health;
+```
+
+In `pgagroal.conf`:
```
fluca1978, line 101: `ensure you have health check turned on and configured properly`

fluca1978 (changes requested):
Great, but I think we should indicate clearly that a few blocks are in the HAProxy space, not in pgagroal.

doc/manual/en/21-ha.md
fluca1978: Having a single role is simpler, and can be used for the current aim. In the long term, however, I would prefer keep things separated.

doc/manual/en/21-ha.md
fluca1978: I think the simpler to understand could be round-robin, then least-connections and then others. I think we should mention that the choice of the right alghoritm depends on the use case.

doc/manual/en/21-ha.md
```diff
@@ -0,0 +1,415 @@
...
+| Replica 1 | `10.0.0.12` | 5432 |
+| Replica 2 | `10.0.0.13` | 5432 |
+
+**pgagroal instances**
+
+| Instance | Listen           | Backend              |
+|----------|------------------|----------------------|
+| Write    | `10.0.0.11:2345` | Primary `10.0.0.11:5432`  |
+| Read 1   | `10.0.0.12:2345` | Replica 1 `10.0.0.12:5432` |
+| Read 2   | `10.0.0.13:2345` | Replica 2 `10.0.0.13:5432` |
```
fluca1978: table formatting?

doc/HA.md
jesperpedersen: Personally, I don't like `standby` as it implies that this is the machine that will be promoted. So, I'm ok with `replica` as it is neutral in context

doc/manual/en/21-ha.md
```diff
@@ -0,0 +1,415 @@
...
+```
+Primary -> Replica 1 -> Replica 2
+```
+
+Replica 2 does not receive data from the primary directly; it receives it through Replica 1.
+This is known as cascading replication.
+
+**Configuring a replica to replicate from another replica**
+
+On Replica 2, set `primary_conninfo` in `postgresql.conf` (PostgreSQL 12+) to point at Replica 1 instead of the primary:
```
fluca1978: is it worth mentioning that the setup does not include replication slots (that I think should be used in a production environment)?

doc/manual/en/21-ha.md
```diff
@@ -0,0 +1,415 @@
...
+
+![Star topology -- Two Level (HAProxy + pgagroal)](images/ha-star-two-level-haproxy.png){width=100%}
+
+---
+
+## Best Practices
+
+### Choosing a Topology
+
+* **Simple setup, independent replicas** -> Star Single Level.
```
fluca1978: I don't like the approach `->`, I think for the manual we need a more verbose writing style.
So something like:
`If you need a **simple setup with independent standbys**, then choose the *Start Single Level* topology`. or something like that.

fluca1978 (changes requested):
I would use a more verbose approach for the manual, but the overall is very good.

doc/manual/en/21-ha.md
Abdelrhmansersawy: I think it’s better to include this. Keeping users safe is important, even if it’s mainly needed in production environments.

fluca1978:
please rebase

Abdelrhmansersawy:
> please rebase

I think I have already rebased, Let me know if I'm wrong.
```
❯ git fetch upstream;
❯ git rebase upstream/master;
Current branch HA-docs is up to date.
❯ git push origin HA-docs --force;
Everything up-to-date
```

fluca1978:
My fault doing diffs.

PR #894, opened 2026-05-14
[#814] Add TLS certificate expiry monitor worker

### LINKED ISSUE #814: TLS Certificate Expiry Monitor Worker
pgagroal parses TLS certificates and exposes metrics such as pgagroal_tls_certificate_expiration_seconds to Prometheus. However, these metrics are only updated during the initial startup of the service.

For long-running instances, this is problematic if certificates are renewed on disk as the Prometheus metrics will continue to report the old expiration date until pgagroal is restarted.

trxvorr:
closes: #814 

adds a background worker that periodically refreshes tls certificate metrics in shared memory, so prometheus always shows the current expiry info without needing a restart.

it handles the main cert, metrics cert, and all server certs. i also added a spinlock to ensure shared memory updates are thread safe against concurrent scrapes. 

defaults to a 1h refresh interval


jesperpedersen:
@trxvorr I think we need to split this into 2.

1) A CLI command - we can likely use `conf reload` and/or `conf set`
2) The monitor job looks for expired certificates and does `ERROR`s

fluca1978:
> @trxvorr I think we need to split this into 2.
> 
>     1. A CLI command - we can likely use `conf reload` and/or `conf set`
> 

Agree, I think `conf` is the right place to hook this into.


>     2. The monitor job looks for expired certificates and does `ERROR`s

Increasing the related metrics too, I guess.

So far it looks good to me, and I agree with changes requested and proposed by Jesper.

fheikens:
Triage note: this has drifted into conflict with master since it was last green. A rebase on the latest master would clear it and make the TLS cert-expiry worker (#814) reviewable again. Thanks!

jesperpedersen:
LGTM

PR #903, opened 2026-05-22
[#848] Enforce per-rule max_size as a hard cap in every pipeline mode

fheikens:
Implements the per-rule max_size hard cap discussed in #848.

Behaviour confirmed on the issue by @fluca1978 and @jesperpedersen: when the
rule cap is reached, a further request waits on the existing exhaustion/retry
path until blocking_timeout and then returns the standard "pool is full"
error. No new config, no new error type.

Approach: replace the non-atomic count-then-create check in
pgagroal_get_connection with an atomic reservation against a new per-rule
counter (struct limit.backend_connections), mirroring the global
max_connections gate. Lock-free, no new mutex. Full rationale and the rollback
sites are in the commit message.

Regression test: test/conf/18 (transaction pipeline, max_size=5) drives more
concurrent holds than the cap and asserts the peak pgagroal-owned backend count
on the primary never exceeds max_size. Fails on master, passes with this fix.

Note: stacked on #887 (concurrent-holds harness), which it reuses. Until #887
merges this PR shows its commit too; will rebase once #887 lands.

### DIFF
--- a/src/include/pgagroal.h
+++ b/src/include/pgagroal.h
@@ -457,6 +457,7 @@ struct limit
    char aliases[MAX_ALIASES][MAX_DATABASE_LENGTH]; /**< The aliases for the database */
    int aliases_count;                              /**< The number of aliases */
    atomic_ushort active_connections;               /**< The active number of connections */
+   atomic_ushort backend_connections;              /**< Live backend connections for the rule (hard-capped at max_size) */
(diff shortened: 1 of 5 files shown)


jesperpedersen:
@fheikens Tell orangu to squash

fluca1978:
merged, thanks @fheikens

Issue #904, opened 2026-05-22
Integration test for per-rule max_size hard cap (#848 follow-up)


fheikens:
Follow-up to #848 / PR #903. The production fix for the per-rule max_size hard cap landed in #903 (validated by direct instrumented reproduction in debug and release: under heavy contention the per-rule live-backend count caps exactly at max_size). The automated integration test was split out to here because its measurement approach proved unreliable in CI.

## What is needed
A regression test that fails on the pre-#903 code (over-creation past max_size) and passes with the fix, runnable under test/check.sh run-configs.

## Why the first attempt was deferred
The attempted test drove N concurrent clients through a small-max_size rule and sampled the peak via `pg_stat_activity` (count of `application_name = pgagroal` backends for the user). Findings:

- The fix is correct: an instrumented build shows the reservation gate caps the live-backend count exactly at max_size, race-free, even under heavy contention (e.g. 147 rollbacks for 6 live slots).
- The `pg_stat_activity` peak sampling over-counts during kill/recreate churn: when pgagroal kills a cached backend and immediately creates a replacement under the freed cap slot, the closing PG-side connection lingers briefly in `pg_stat_activity` alongside the new one. CI's PostgreSQL tears these down slower than a local postgres image, so the instantaneous count transiently reads above max_size (observed 14 vs cap 6) even though live backends never exceed the cap.
- The harness needed `pgagroal_tsclient_init` to load the LIMIT configuration (done in #903) so a limit-aware test can see the rules in shmem.

## Suggested direction
Assert on pgagroal's own authoritative live-backend count for the rule (the new `backend_connections` counter, exposed via the metrics/management interface) rather than a racy `pg_stat_activity` peak. That is exactly what the cap governs and is not subject to PG-side teardown lag.

Depends on the concurrent-holds harness in #887/#879.

fheikens:
Filed #908 with the options for making this testable. The cap test cannot be hosted in run-configs under the Debug 8-connection ceiling (the existing mydb + utf8db load tests already consume the full budget, and a single config file cannot differ per build type). Keeping this open pending a direction in #908; the scaffolding lives in #907.

PR #906, opened 2026-05-23
[#905] Expose per-rule backend_connections via pgagroal_limit

### LINKED ISSUE #905: Expose per-rule backend_connections via the pgagroal_limit metric
The per-rule hard cap from #848/#903 is enforced against `backend_connections` (live backends per limit rule, capped at `max_size`). That counter is the authoritative measure of how close a rule is to its cap, but it currently lives only in shared memory and is not observable.

`pgagroal_limit` already reports `min`/`initial`/`max`/`active` per rule. `active` is the `active_connections` metric, which counts in-use connections (including reuse) rather than the live backend count the cap actually governs, so it is not a faithful read of the cap state.

Proposal: add a `type="backend"` series to `pgagroal_limit` reporting `limits[i].backend_connections`.

- Operators get direct visibility into how close each rule is to `max_size`.
- It gives the #904 regression test an authoritative signal to assert on, instead of a racy `pg_stat_activity` peak.

Scope: `limit_information()` in `prometheus.c`, the `pgagroal_limit` help section on the metrics HTML page, and metrics docs.

fheikens:
Closes #905.

The per-rule hard cap from #848 is enforced against `backend_connections` (live backends per rule, capped at `max_size`), but that counter lived only in shared memory and was not observable.

This adds a `type="backend"` series to the `pgagroal_limit` metric reporting `limits[i].backend_connections`, alongside the existing `min`/`initial`/`max`/`active` series, and lists it on the metrics HTML page. `active` reports `active_connections`, which counts in-use connections including reuse and so is not a faithful read of the cap state; `backend` is.

This also gives the #904 regression test an authoritative signal to assert the cap against, instead of a racy `pg_stat_activity` peak.

### DIFF
--- a/src/libpgagroal/prometheus.c
+++ b/src/libpgagroal/prometheus.c
@@ -1600,6 +1600,7 @@ home_page(SSL* client_ssl, int client_fd)
    data = pgagroal_append(data, "            <li>initial</li>\n");
    data = pgagroal_append(data, "            <li>max</li>\n");
    data = pgagroal_append(data, "            <li>active</li>\n");
+   data = pgagroal_append(data, "            <li>backend</li>\n");
    data = pgagroal_append(data, "          </ul>\n");
    data = pgagroal_append(data, "        </td>\n");
    data = pgagroal_append(data, "      </tr>\n");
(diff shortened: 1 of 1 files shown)


fluca1978:
merged, thanks @fheikens 

This also emphatizes that we should improve prometheus documentation.

PR #907, opened 2026-05-23
[#904] Add per-rule max_size hard-cap regression test

### LINKED ISSUE #904: Integration test for per-rule max_size hard cap (#848 follow-up)

fheikens:
## Why

The per-rule `max_size` hard cap (#848 / #903) was validated by instrumented reproduction but had no automated regression coverage. An earlier attempt sampled `pg_stat_activity`, which over-counts during PG-side teardown lag and was unreliable in CI (#904).

## What

`test_pgagroal_pool_max_size_cap` oversubscribes a rule with `2 * max_size` concurrent holds and, while they run, samples pgagroal's own per-rule live backend counter — the `pgagroal_limit{type="backend"}` series (#905) — and asserts the peak never exceeds `max_size`. It fails on the pre-#903 count-then-create TOCTOU (over-creation past `max_size`) and passes with the hard cap. Sampling pgagroal's authoritative counter sidesteps the `pg_stat_activity` teardown-lag over-count.

- `pgagroal_tsclient_limit_backend_peak()` drives the holds and samples the metrics endpoint.
- The cap config (`test/conf/release-only/18`) enables metrics and ships a no-prefill rule, so every acquisition is a fresh backend creation — prefilled connections would be reused rather than created and would mask the cap.
- Parameters are derived from the live config; configurations without metrics or without a matching rule are skipped.

## Dependency

Stacked on #914 (#908): the cap config needs more connections than the Debug ceiling allows, so it lives in the build-type room and runs only against a non-Debug build. Review #914 first; this PR will rebase onto master once it merges.

## Testing

Built clean; the test links and registers. Fail-before/pass-after is exercised by the Release `run-configs` CI job, where the release-only config runs.

closes #904

### DIFF
--- a/test/conf/release-only/18/pgagroal.conf
+++ b/test/conf/release-only/18/pgagroal.conf
@@ -1,0 +1,33 @@
+[pgagroal]
+host = localhost
+port = 2345
+
+log_type = file
+log_level = debug5
+log_path = test.log
+
+metrics = 2346
+
+# max_connections is set above 2 * mydb max_size so the per-rule max_size cap
+# (not the global max_connections gate) is the binding constraint when
+# test_pgagroal_pool_max_size_cap oversubscribes the rule with 2 * max_size
+# concurrent holds.
+max_connections = 24
+blocking_timeout = 10
+idle_timeout = 600
+validation = off
+unix_socket_dir = /tmp/
+pipeline = performance
+
+# Pin the epoll event backend for this test. The max_size cap this test
+# verifies is enforced correctly on every backend, but the io_uring backend
+# has a separate liveness stall under this exact oversubscription shape
+# (tracked in #923) that would hang a waiter indefinitely. epoll is clean, so
+# the cap regression runs deterministically here. Remove this line once #923
+# is fixed to also exercise the default backend.
+ev_backend = epoll
+
+
+[primary]
+host = localhost
+port = 5432
--- a/test/conf/release-only/18/pgagroal_databases.conf
+++ b/test/conf/release-only/18/pgagroal_databases.conf
@@ -1,0 +1,15 @@
+#
+# DATABASE=ALIAS1,ALIAS2 USER MAX_SIZE INITIAL_SIZE MIN_SIZE
+#
+# max_size gives headroom above what the other run-configs testcases need:
+# the connection/alias load tests run pgbench -c 6 (~7 backends including
+# pgbench's setup connection), so max_size = 10 keeps them clear of the cap.
+# No prefill (initial/min = 0) so every acquisition is a fresh backend
+# creation -- prefilled connections would be reused rather than created and
+# would mask the per-rule max_size hard cap (#848) the cap testcase exercises
+# by oversubscribing with 2 * max_size holds. The utf8 rule mirrors the
+# default so utf8 testcases keep working when this configuration overrides the
+# default databases.conf in run-configs.
+#
+mydb=pgalias1,pgalias2 myuser 10 0 0
+utf8db utf8user 2 2 1
--- a/test/conf/release-only/18/pgagroal_hba.conf
+++ b/test/conf/release-only/18/pgagroal_hba.conf
@@ -1,0 +1,1 @@
+host    all all all trust
--- a/test/include/tsclient.h
+++ b/test/include/tsclient.h
@@ -102,6 +102,23 @@ pgagroal_tsclient_init_pgbench(char* user, char* database, int scale);
 int
 pgagroal_tsclient_execute_concurrent_holds(char* user, char* database, int client_count, int hold_seconds);
 
+/**
+ * Drive client_count concurrent holds against a per-rule max_size cap and
+ * sample pgagroal's own live-backend counter for the rule while they run.
+ * Reads the pgagroal_limit type="backend" series from the metrics endpoint
+ * (issue #905) and returns the peak observed during the hold window. Used to
+ * assert that the per-rule max_size hard cap (issue #848) is never exceeded
+ * under concurrent contention.
+ * @param user name of the user
+ * @param database name of the database
+ * @param client_count number of concurrent psql sessions to spawn
+ * @param hold_seconds duration of the pg_sleep() hold per session
+ * @return the peak per-rule live backend count, or -1 if the metrics endpoint
+ * is disabled or could not be scraped
+ */
+int
+pgagroal_tsclient_limit_backend_peak(char* user, char* database, int client_count, int hold_seconds);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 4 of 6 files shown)


fheikens:
Marking this draft. The automated regression test is fighting the run-configs harness rather than the fix.

Findings:
- The test build caps `max_connections` at 8 (compile-time) and requires `sum(max_size) <= max_connections`. Within 8 total connections there is no room to both oversubscribe a rule for the cap test and leave room for the existing `pgbench -c 6` load tests that run against the same configuration on the same rule. Moving the cap test to `utf8db` only relocates the collision (it is load-tested too).
- Oversubscription churn also surfaced what looks like a `backend_connections` accounting discrepancy in the merged #848 path: after the cap test a subsequent client blocks for the full `blocking_timeout` and then gets "pool is full" even though a slot is free (the counter appears to sit at the cap while a slot is actually idle/unused). That deserves separate investigation rather than being papered over by this test.

The cap fix in #903 was validated by instrumented reproduction, and #906 now exposes the per-rule live backend count via `pgagroal_limit{type="backend"}`, so the cap is observable. Leaving #904 open and revisiting the automated test with a harness-friendly design once the accounting question above is resolved.

fheikens:
Options for unblocking this are in #908. Holding as a draft until a direction is chosen there.

fheikens:
Pushed ccd36bd to address the Release `run-configs` hang.

Root cause: the test reached `pthread_join` after the metric window closed, but at least one psql client never returned — a connection handler stalled under `2 * max_size` oversubscription on the performance pipeline (config-18 daemon log shows the cap holding and the main daemon idle for the full ~25 min until CI cancelled). With no bound on the client invocations, one stuck client pinned the whole run.

The fix wraps each psql invocation in the concurrent-hold helpers with `timeout(1)`, bounded by `blocking_timeout + 2*hold + margin` — well above the ~7s legitimate runtime, so passing runs are unaffected and a stuck client is killed instead of hanging CI.

The underlying connection stall is tracked separately in #923.

Note: this branch is still based on the now-merged #908 commit and needs a rebase onto `master` (dropping that commit) before merge, as flagged in the description.

fheikens:
Status update. The harness blocker that put this in draft — #908 (build-type configuration room for `run-configs`) — is now merged to `master` (`e061c13`), so the `sum(max_size) <= max_connections` collision against the existing load tests on the same rule is resolved.

The one remaining reason to keep this in draft is that the test legitimately exercises the stall tracked in #923. The watchdog from `ccd36bd` keeps a stuck client from hanging CI, but until #923 is root-caused the test stays gated on it rather than landing as a known-red regression. I'll rebase onto `master` (dropping the now-merged #908 commit this branch was based on) when #923 lands, so it's mergeable in one step. Holding as draft until then.

fluca1978:
See CI failures

fluca1978:
Please rebase

fheikens:
Rebased onto current master (`f204693`) as requested — the two commits now sit cleanly on top with no conflicts.

One thing to flag on the CI state: this test deliberately drives the per-rule `max_size` oversubscription path, which is exactly the connection-handler stall tracked in #923. That issue is still open and unresolved, so the FreeBSD `run-configs` failure reflects the underlying stall rather than a problem with the test itself. The `timeout(1)` watchdog from the second commit keeps a stuck client from hanging CI indefinitely, but the run still goes red while #923 stands.

My preference is to hold this until #923 is root-caused, then land it as a passing regression in one step. Happy to keep it open (or move it back to draft) in the meantime — whichever you'd prefer.

fheikens:
Pushed `3ec5bd7` to unblock this without waiting on the #923 fix.

The test's job is to verify the per-rule `max_size` hard cap (#848 follow-up), and that invariant holds on **every** event backend — the cap is never exceeded, including under io_uring. What only fails under io_uring is a separate *liveness* stall (a waiter hanging past `blocking_timeout`), which I've traced to the io_uring backend's main-loop accept path and am tracking in **#923** (epoll is clean on the identical recipe).

So I've pinned `ev_backend = epoll` in `test/conf/release-only/18`, with a comment pointing at #923 and a note to remove the pin once that's fixed. This lets the cap regression run deterministically now instead of gating on an unrelated event-backend bug. Only the Linux `run-configs` job runs this configuration (macOS/FreeBSD build only), where epoll is fully supported and clean.

Rebased onto current master earlier, so this is a clean three-commit stack. Once CI is green I think it's ready to merge; the #923 liveness fix will lift the pin as a follow-up.

fluca1978:
can we squash the commits?

fheikens:
Squashed to a single commit (ae0c3a5); the message carries the watchdog and epoll-pin rationale from the individual commits.

fheikens:
The red CI here is not this branch: master itself fails to build (#961 — #941/#871 semantic conflict; fix in #962). The squash didn't change the tree (git diff against the pre-squash head is empty). Once #962 lands, a re-run should come back green.

fluca1978:
rerunning the jobs still fails on FreeBSD, since #962 is merged, could it be the need for a rebase?

fheikens:
Rebased onto current master (368a762). The FreeBSD failure was the 14.3 package-repository mismatch fixed in #964, which the branch now includes — CI should come back green.

fluca1978:
Merged, thankls @fheikens

PR #913, opened 2026-05-25
[#912] Fix double-free when startup packet omits database parameter

### LINKED ISSUE #912: Double-free in pgagroal_extract_username_database when startup packet omits database
`pgagroal_extract_username_database aliases` the database pointer to username when the client startup packet has no database parameter:

Callers such as `use_unpooled_connection` in `security.c` then `free()` both pointers independently, causing a double-free that kills the worker process.

The most common trigger is HAProxy's option pgsql-check` health probe, which sends a startup packet with only a user field and no database.

Abdelrhmansersawy:
pgagroal_extract_username_database was aliasing *database to *username when no database was provided. Callers free both pointers independently, causing a double-free that silently killed the worker process.

Common trigger: HAProxy pgsql-check probes, which never include a database field, crashed a worker on every health check interval.

Fix by using strdup instead of aliasing.

Close #912

### DIFF
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -148,7 +148,7 @@ pgagroal_extract_username_database(struct message* msg, char** username, char** 
 
    if (*database == NULL)
    {
-      *database = *username;
+      *database = strdup(*username);
    }
 
    pgagroal_log_trace("Username: %s", *username);


fluca1978:
merged, thanks @Abdelrhmansersawy

jesperpedersen:
@Abdelrhmansersawy Remember to cross-port

PR #914, opened 2026-05-25
[#908] Add build-type configuration room to run-configs

### LINKED ISSUE #908: Automated regression coverage for the per-rule max_size hard cap is blocked by the Debug connection ceiling
Follow-up to #904. The per-rule `max_size` hard cap (#848 / #903) is validated by instrumented reproduction and is now observable via `pgagroal_limit{type="backend"}` (#906), but an automated regression test has proven impossible to add as a `run-configs` integration test. This issue records why and proposes options.

## Why it is blocked

The integration path in CI is `check.sh run-configs-ci-nonbuild`, which runs every testcase against every `test/conf/*` directory. A cap regression test has to oversubscribe a per-rule `max_size` (drive more concurrent clients than the rule allows) and assert the live backend count never exceeds it. That requires slot headroom, and there is none in the Debug build:

- `MAX_NUMBER_OF_CONNECTIONS` is `8` in Debug and `10000` otherwise (`src/include/pgagroal.h`). CI runs both Debug and Release against the same config files.
- In Debug, `max_connections` is clamped to 8 and pgagroal refuses to start if `sum(max_size) > max_connections` (FATAL `LIMIT: Too many connections defined`).
- The existing load testcases already consume the entire budget: `mydb` (`pgbench -c 6`, needs `max_size` 6) + `utf8db` (`pgbench -c 2`, needs `max_size` 2) = 8.

So there is no room for a dedicated oversubscribable rule in Debug, and a single `conf/NN` file cannot differ per build type. Moving the cap rule onto `mydb` or `utf8db` collides with their load tests; prefilling to satisfy the load tests masks the cap (reused connections are never created, so over-creation never triggers); adding a third rule exceeds the Debug `sum(max_size) <= 8` limit.

## Options

1. **Raise the Debug `MAX_NUMBER_OF_CONNECTIONS`** (e.g. 8 -> 16 or 32). Smallest change; leaves room for a dedicated no-prefill cap rule in a purpose-built `conf/NN`, and the cap testcase can skip when the ceiling is still small. Risk: changes Debug-build pool sizing and any debug-only assumptions that rely on the small ceiling.
2. **Per-config test selection in the harness.** Let a testcase declare which `test/conf/*` it applies to (or let a config opt testcases in/out), so the cap test runs only against a config built for it. Larger harness change, but generally useful.
3. **Focused unit test of the reservation logic.** Exercise the `backend_connections` reserve/rollback path (the #848 serialization point) directly under simulated contention, without real backends or the pool ceiling. Avoids `run-configs` entirely; tests the mechanism rather than end-to-end behaviour.

## Current state

- Fix: #903 (validated by instrumented reproduction).
- Observability: #906 adds `pgagroal_limit{type="backend"}`.
- Scaffolding: #907 (draft) carries the `tsclient` metrics-sampling helper and the cap testcase; it can be revived once a direction here is chosen.

Recommendation: option 1 if the Debug ceiling can be raised without disrupting other debug-only behaviour; otherwise option 3.

fheikens:
## Why

`run-configs` runs every `test/conf/*` directory against every build. A configuration that oversubscribes a per-rule `max_size` cap needs more connections than the Debug ceiling allows — `MAX_NUMBER_OF_CONNECTIONS` is `8` under `-DDEBUG` — so it cannot be added without breaking the Debug build's `sum(max_size) <= max_connections` start-up check. This blocked the regression test in #908.

## What

Adds a build-type configuration room:

- Shared configurations under `test/conf/*` still run against every build.
- Configurations under `test/conf/release-only/` run only against a non-Debug build.

The build type is read from `build/CMakeCache.txt` (`CMAKE_BUILD_TYPE=Debug` maps to `-DDEBUG` in `src/CMakeLists.txt`). An undetectable build type is treated as Debug, so a missing cache never triggers a start-up FATAL.

The per-config run logic is factored into a helper so both the shared loop and the release-only loop share one code path.

## Testing

- `bash -n` and `shellcheck` clean on the changed file.
- Verified the selection logic against both a Release cache (release-only configs included) and a Debug cache (release-only configs skipped), and that `release-only` is never treated as a config in the shared loop.

This is the harness change only. The cap regression test that uses the room is in #904 (stacked on this PR).

closes #908

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -528,72 +528,114 @@ execute_testcases() {
    set -e
 }
 
+# Determine the build type of the compiled binary. The per-rule cap config in
+# test/conf/release-only needs more than the Debug connection ceiling
+# (MAX_NUMBER_OF_CONNECTIONS = 8 under -DDEBUG), so it only runs against a
+# non-Debug build. CMAKE_BUILD_TYPE=Debug maps to -DDEBUG in src/CMakeLists.txt,
+# so the build cache is the authoritative signal.
+detect_build_type() {
+   local cache="$PROJECT_DIRECTORY/build/CMakeCache.txt"
+   if [ -f "$cache" ]; then
+      grep -E '^CMAKE_BUILD_TYPE:' "$cache" | head -n1 | cut -d= -f2
+   fi
+}
+
+# Run the full testsuite against a single configuration directory.
+run_single_config_dir() {
+   local entry config_name
+   entry=$(realpath "$1")
+   config_name=$(basename "$entry")
+
+   if [[ ! -d "$entry" || ! -f "$entry/pgagroal.conf" || ! -f "$entry/pgagroal_hba.conf" ]]; then
+      echo "Warning: Configuration directory '$entry' is missing required files"
+      echo "  Required: pgagroal.conf and pgagroal_hba.conf"
+      return
+   fi
+
+   echo ""
+   echo "=========================================="
+   echo "Testing configuration: $config_name"
+   echo "=========================================="
+
+   # Backup current configuration
+   cp "$CONFIGURATION_DIRECTORY/pgagroal.conf" "$CONFIGURATION_DIRECTORY/pgagroal.conf.backup"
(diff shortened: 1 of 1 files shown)


fluca1978:
merged, thanks @fheikens

PR #916, opened 2026-05-25
[#915] Configure cmake for both compiler choices

### LINKED ISSUE #915: check.sh: misplaced `fi` makes cmake run only in the gcc branch
## Summary

In `test/check.sh` the build step's `fi` is positioned after the `cmake` invocation, so `cmake` is configured only on the gcc path. The default (clang) path never runs `cmake`.

## Location

`test/check.sh` (build step):

```sh
if [[ $MODE != gcc ]]; then
  export CC=$(which clang)
  echo "Using Clang compiler with LLVM coverage: $CC"
else
  export CC=$(which gcc)
  echo "Using GCC compiler without coverage: $CC"
cmake -DCMAKE_C_COMPILER=$CC \
     -DCMAKE_BUILD_TYPE=Debug \
     ..
fi
make -j$(nproc)
```

The `if/else` is only meant to select the compiler (`CC`); the `cmake` call should run unconditionally afterwards. As written, `cmake` lives inside the `else`, so:

- In the default clang mode, `cmake` is never invoked. `make` then relies on a pre-existing `CMakeCache.txt`; on a fresh `build/` it fails, and on an existing one it silently reuses whatever build type the cache already has.
- `CC` is exported but only applied to the cmake configure on the gcc path.

## Expected

`cmake` runs for both compiler choices. Move the `fi` to before the `cmake` call:

```sh
if [[ $MODE != gcc ]]; then
  export CC=$(which clang)
  echo "Using Clang compiler with LLVM coverage: $CC"
else
  export CC=$(which gcc)
  echo "Using GCC compiler without coverage: $CC"
fi
cmake -DCMAKE_C_COMPILER=$CC \
     -DCMAKE_BUILD_TYPE=Debug \
     ..
make -j$(nproc)
```

## Notes

Found while working on the build-type config room (#908). Not addressed there to keep that change scoped.

fheikens:
## Why

In `test/check.sh` the build step's `fi` sat after the `cmake` call, so `cmake` was configured only on the gcc path. In the default clang mode `cmake` was never invoked and `make` relied on a pre-existing `CMakeCache.txt` — failing on a fresh `build/` and silently reusing the cached build type otherwise.

## What

The `if/else` only selects the compiler (`CC`), so move the `fi` before the `cmake` call. `cmake` now runs unconditionally after the compiler is chosen.

## Testing

`bash -n` clean. The build step now configures with the selected compiler regardless of mode.

closes #915

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -1055,10 +1055,10 @@ do_setup() {
    else
      export CC=$(which gcc)
      echo "Using GCC compiler without coverage: $CC"
+   fi
    cmake -DCMAKE_C_COMPILER=$CC \
         -DCMAKE_BUILD_TYPE=Debug \
         ..
-   fi
    make -j$(nproc)
    cd ..
   fi


Userfrom1995:
@fheikens Thank you for your contribution.

PR #919, opened 2026-05-25
[#918] Add a performance-regression check to CI

### LINKED ISSUE #918: Add a performance-regression check to CI
## Summary

Performance is one of pgagroal's core claims, but nothing currently guards it from regression. This adds a per-PR performance check.

## Design

CI runs on `ubuntu-latest` — shared runners with large run-to-run variance — so an absolute committed baseline would be flaky (a perf gate that cries wolf gets ignored). Instead:

- **A/B on the same runner.** Build both the base (merge-base / master) and the PR, run the same pgbench workloads through pgagroal against each, back-to-back on the same machine. Runner speed cancels out; regression is measured PR-vs-base, not against a static number — and there's no baseline to maintain or let go stale.
- **Median of N runs** (default 3) per side, with a documented noise band.
- **Advisory first.** Post the PR-vs-base delta (TPS, p99) as a PR comment, non-blocking. Promote to a blocking status once it's proven stable over real PRs.
- **Time-boxed.** Each benchmark phase has a hard cap so the job can never hang a runner.

## Workloads

read-only, read-write, and connect-heavy pgbench through pgagroal, reusing the container / pgbench setup already in `test/check.sh`.

## Acceptance criteria

- A no-op PR reports a near-zero delta, within the noise band.
- A real regression (e.g. a 20% TPS drop vs base) is flagged in the comment.
- The job is bounded in time and never silently passes on a setup/build failure.
- Defaults: TPS regression > 5% or p99 > 10% flagged; tunable.

Part of protecting the performance moat; complements the fuzzing/hardening work.

fheikens:
## Why

Performance is a core pgagroal claim, but nothing guards it from regression today.

## What

An **advisory** per-PR performance check. CI runs on `ubuntu-latest`, whose run-to-run variance makes an absolute baseline flaky, so this measures the **PR relative to its base on the same runner** (A/B): it builds and benchmarks both back-to-back and reports the PR-vs-base delta — no static baseline to maintain.

- `test/perf/run_bench.sh` — read-only / read-write / connect-heavy pgbench through pgagroal → TPS + p99 JSON; each workload is `timeout`-capped so the job can't hang.
- `test/perf/bench_ref.sh` — build a source tree, start pgagroal against the local PostgreSQL, run the workloads, stop it.
- `test/perf/compare.py` — diff base vs PR; flags TPS drop > 5% or p99 rise > 10%.
- `.github/workflows/perf.yml` — orchestrates the A/B run and posts the delta as a comment.

## Rollout

Advisory for now — it comments, it does not fail the build. Promote to blocking (`--fail-on-regression`) once it's proven stable over real PRs.

## Note

`pull_request` workflows run from the base branch's copy, so this won't run on its own PR — it first executes on the next PR after merge.

closes #918

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -1,0 +1,120 @@
+name: Performance
+
+# On-demand only: trigger from the Actions tab (Run workflow) with the number
+# of the pull request to benchmark. The PR head is measured A/B against the
+# current master on the same runner.
+on:
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: "Pull request number to benchmark (head vs current master)"
+        required: true
+        type: string
+
+permissions:
+  contents: read   # results go to the job summary + artifacts; no PR write needed
+
+jobs:
+  perf:
+    runs-on: ubuntu-latest
+    # Advisory only: a shared GitHub runner is too noisy to gate merges on, so
+    # a red run is informational, not a failure to fix. The A/B comparison is
+    # published to the job summary and the raw results are uploaded as an
+    # artifact for humans to inspect.
+    continue-on-error: true
+    container:
+      image: rockylinux/rockylinux:10
+      options: --security-opt seccomp=unconfined
+    timeout-minutes: 30
+    env:
+      PGA_PORT: "6432"
+      PGHOST: localhost
+      PGPORT: "5432"
+      PGUSER: bench
+      PGDATABASE: benchdb
+      PGPASSWORD: benchpass
+      PGAGROAL_RUN_AS: pg
+      PERF_DURATION: "20"
+      PERF_CLIENTS: "16"
+    steps:
+      - uses: actions/checkout@v6   # the PR head
+        with:
+          ref: refs/pull/${{ inputs.pr }}/head
+
+      - name: Check out base for A/B
+        uses: actions/checkout@v6
+        with:
+          ref: master
+          path: base-src
+
+      - name: Install dependencies
+        run: |
+          dnf -y install dnf-plugins-core
+          dnf config-manager --set-enabled crb
+          dnf -y install epel-release
+          dnf -y install \
+            git gcc clang cmake make \
+            openssl openssl-devel \
+            systemd systemd-devel zlib zlib-devel \
+            libzstd libzstd-devel lz4 lz4-devel \
+            libatomic liburing liburing-devel \
+            python3 python3-docutils bzip2 bzip2-devel \
+            net-tools procps-ng util-linux sudo which
+
+      - name: Install PostgreSQL
+        run: |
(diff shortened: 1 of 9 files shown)


fluca1978:
@Userfrom1995 thoughts?

jesperpedersen:
I'm ok with having this as an optional thing on CI - but it will likely trigger every time.

jesperpedersen:
@fheikens You need license headers

fheikens:
Added the BSD copyright header to the six test/perf scripts (c1d8a13), matching test/check.sh. perf.yml and README.md are left as-is since workflows and markdown carry no header in the repo.

fheikens:
Made the job advisory in d4022ac: `continue-on-error: true` on the perf job, so it never blocks a PR. The A/B comparison still publishes to the job summary and uploads the raw results as an artifact, but a red run is informational. That keeps it as the optional CI signal you asked for without it gating merges on runner noise.

fheikens:
Both review points are addressed and CI is green across all platforms, including the new `perf` job: license headers added to the six `test/perf` scripts (`c1d8a13`), and the job is advisory via `continue-on-error: true` (`d4022ac`) — a red run only publishes the A/B comparison to the job summary and uploads the raw results as an artifact, it never gates a PR. That keeps it as the optional signal you asked for without it blocking anything. Whenever you have a moment for another look — no rush.

Userfrom1995:
> @Userfrom1995 thoughts?

I like the idea, but I don't think we need to run it on every PR. @fheikens, could you make it so it can be triggered manually instead? Alternatively, we could support a PR comment slash command like `/perf_ci`, where a collaborator or reviewer can post that comment to trigger the performance CI test.

I also think we could expand this into a more comprehensive performance regression test suite in the future by adding more test cases and functionality. That can obviously be done in future PRs.

fheikens:
Done in 64b4fb1 — the `pull_request` trigger is gone; the workflow is now `workflow_dispatch` only. A collaborator triggers it from Actions → Performance → Run workflow with the PR number as input, and the job checks out `refs/pull/<N>/head` and measures it A/B against current `master`. The README documents the procedure.

I left the `/perf_ci` comment command out of this PR deliberately: an `issue_comment` workflow only ever runs from the copy of the file on the default branch, so it can't be exercised or reviewed from this PR, and it needs author-association gating to stop arbitrary commenters from triggering builds of fork code. It's a small self-contained follow-up once this file is on master — happy to file an issue for it.

Agreed on growing this into a broader suite over time; the run_bench.sh/compare.py split was designed so new workloads are just additional entries.

fluca1978:
@fheikens please re-edit commit and squash, so we can merge and see this in action!

fheikens:
Squashed into a single commit (4d9a266), CI is green across all platforms. Ready to merge.

fluca1978:
merged, thanks @fheikens

PR #921, opened 2026-05-26
[#891] add flush timeout to shutdown and flush commands

### LINKED ISSUE #891: Introduce flush_timeout opt-config
**Is your feature request related to a problem? Please describe.**
 Introduce `flush_timeout` - the maximum time the server waits for an active client to finish before forcing a pause.

We could reuse `flush_timeout` for all commands to prevent hanging on long-running queries, e.g. ( `pgagroal-cli shutdown gracefully`, `pgagroal-cli reload`)

AbduAllahhany:
Resolves #891 

## Summary

Adds a new `flush_timeout` configuration option and two new CLI commands:

- `flush --timeout [Duration] [database]`
- `shutdown --timeout [Duration]`

These option performs graceful operations bounded by a timeout.

If the timeout expires:

- bounded flush escalates remaining graceful connections to `FLUSH_ALL`
- bounded shutdown escalates to immediate shutdown

`Duration` provided on the CLI overrides the configured `flush_timeout`.
If omitted, pgagroal uses the configured value.
If `flush_timeout = 0`, the operation remains unbounded.

Introduce oneshot timer in `ev.c`  adding repeat_ms to `struct periodic_watcher`


doc/manual/en/04-configuration.md
```diff
@@ -46,6 +46,7 @@ The available keys and their accepted values are reported in the table below.
...
 | rotate_frontend_password_length | 8 | Int | No | The length of the randomized frontend password |
 | max_connection_age | 0 | String | No | The maximum amount of time that a connection will live. If this value is specified without units, it is taken as seconds. It supports the following units as suffixes: 'S' for seconds (default), 'M' for minutes, 'H' for hours, 'D' for days, and 'W' for weeks. (disable = 0) |
+| flush_timeout | 0 | String | No | Default upper bound applied to `pgagroal-cli flush timeout` and `pgagroal-cli shutdown timeout` when the CLI does not pass an explicit number of seconds. If this value is specified without units, it is taken as seconds. It supports the following units as suffixes: 'S' for seconds (default), 'M' for minutes, 'H' for hours, 'D' for days, and 'W' for weeks. When the timer expires, `flush timeout` escalates to `flush all` for the targeted database, and `shutdown timeout` forces an immediate shutdown. (disable = 0; the bounded commands then run unbounded unless an explicit value is passed on the CLI) |
```
fluca1978: specify first the unit (seconds) and then when it is used

doc/manual/en/14-cli-tools.md
```diff
@@ -46,23 +46,33 @@ The `flush` command performs a connection flushing.
 It accepts a *mode* to operate the actual flushing:
 - `gracefully` (the default if not specified), flush connections when possible;
 - `idle` to flush only connections in state *idle*;
-- `all` to flush all the connections (**use with caution!**).
+- `all` to flush all the connections (**use with caution!**);
+- `timeout` performs a graceful flush bounded by a deadline; once the deadline expires, any connections still marked for flush are forcibly terminated (equivalent to `flush all` for the targeted database).
```
fluca1978: shouldn't be a `gracefully --timeout` then? Can we apply to all the other flushing methods then?

src/include/configuration.h
```diff
@@ -143,6 +144,7 @@ extern "C" {
 #define CONFIGURATION_DEFAULT_LOG_PATH        "/tmp/pgagroal.log"
 #define CONFIGURATION_DEFAULT_MAX_CONNECTIONS "100"
 #define CONFIGURATION_DEFAULT_IDLE_TIMEOUT    "600"
+#define CONFIGURATION_DEFAULT_FLUSH_TIMEOUT   "0"
```
fluca1978: default is disabled? I think we should set to `60` or so.

src/include/management.h
```diff
@@ -235,7 +237,7 @@ pgagroal_management_create_outcome_failure(struct json* json, int32_t error, str
  * @return 0 upon success, otherwise 1
  */
 int
-pgagroal_management_request_flush(SSL* ssl, int socket, int32_t mode, char* database, uint8_t compression, uint8_t encryption, int32_t output_format);
+pgagroal_management_request_flush(SSL* ssl, int socket, int32_t mode, char* database, int32_t timeout, uint8_t compression, uint8_t encryption, int32_t output_format);
```
fluca1978: should be `uint_8` instead of `int32`? Is `PGAGROAL_TIME_SEC` returning a `int32`?

fluca1978 (changes requested):
Is it worth changing `timeout` from a command to an argument as `flush all --timeout`?

AbduAllahhany:
@fluca1978 
I change it to be `gracefully --timeout` and the commands will be 
- flush gracefully [Duration] [database]
- shutdown gracefully [Duration]

There is two oneshot timers one for shutdown and one for flush db that scan all dbs and check if it should be flushed if its time expires 

I changed `ev.c` so that I introduced oneshot timer for the opeartion  and add guard to  eparate a cancel request CQE from the original request’s terminal CQE by tagging the cancel SQE with NULL user_data, so the handler routes it through the null-watcher path.
This prevents double-handling and lets the real cancelled request finish cleanly with res == -ECANCELED, which is treated as a normal cancellation, so we can memset the watcher safely.
Should we do this in sepereate pr ? 

For `timeout_ms` it is int so the max will be 24 days is that acceptable or we should change `static int (*periodic_init)(struct periodic_watcher*, int, int);` in `ev.c` .

Is there any changes I should do ?

fluca1978:
Looks good, please see CI errors.
I need a few days to better review this.

jesperpedersen:
@fluca1978 Yes, the direction is good

fluca1978:
merged, thanks @AbduAllahhany

PR #922, opened 2026-05-26
[#813] Replace fixed 500ms pool-acquisition retry with bounded exponential back-off

### LINKED ISSUE #813: Revisit hardcoded 500ms retry delay in pool acquisition under blocking_timeout

fheikens:
When `blocking_timeout` is set and the pool is saturated, `pgagroal_get_connection()` retried with a fixed 500ms sleep (`pool.c` `retry2`). Under brief saturation a connection freed shortly after a sleep began was not detected until the full 500ms elapsed — a median unnecessary wait of ~250ms.

## Change

Replace the fixed sleep with bounded exponential back-off: 1ms initial, doubling each retry, capped by a new `connection_retry_delay` setting (default 250ms), still bounded by `blocking_timeout`. The doubling/clamp is factored into a pure `pgagroal_pool_next_retry_delay()` helper.

A simulation posted on the issue showed the 1ms→250ms back-off gives ~38x lower overshoot than the fixed 500ms under brief saturation, for only ~2.7x more retries under sustained saturation.

## Config

`connection_retry_delay` (milliseconds, default 250) — the back-off cap. Range-validated to 1–999ms (the `SLEEP` macro is sub-second only; `nanosleep` rejects `tv_nsec >= 1e9`); out-of-range values clamp with a warning. Reloadable. Documented in `CONFIGURATION.md`, the manual, and the man page.

## Tests

New MCTF unit tests in `test/testcases/test_pool_backoff.c` (pure, no PostgreSQL harness):
- `test_backoff_sequence` — sequence is 1,2,4,…,250,250 ms
- `test_backoff_cap_clamped` — cap clamps to [1,999]ms
- `test_connection_retry_delay_validation` — valid / clamp-low / clamp-high / reject

## Scope

Per the discussion on #813, two directions are deliberately deferred to follow-ups: cross-process callback/wakeup signalling (needs a shmem futex/semaphore between the returning worker and the waiter) and per-pipeline default delays.

close #813


src/include/pgagroal.h
```diff
@@ -715,6 +718,7 @@ struct main_configuration
    bool allow_unknown_users;         /**< Allow unknown users */
 
    pgagroal_time_t blocking_timeout;                 /**< The duration of blocking timeout (Default seconds) */
+   int connection_retry_delay;                       /**< Back-off cap (milliseconds) on the blocking acquisition retry path */
```
fluca1978: uint?

src/libpgagroal/pool.c
```diff
@@ -452,6 +457,42 @@ pgagroal_get_connection(char* username, char* database, bool reuse, bool transac
...
+
+   cap_ns = (long)cap_ms * 1000000L;
+
+   if (current_ns <= 0)
+   {
+      next = 1000000L; /* 1ms seed */
+   }
+   else
+   {
+      next = current_ns * 2;
```
fluca1978, line 486: this is not "exponential" per se, only if I start with `1` or a `2`-based value. So we should document that the capability is doubled at each round.

doc/man/pgagroal.conf.5.rst
```diff
@@ -96,6 +96,11 @@ blocking_timeout
   'H' for hours, 'D' for days, and 'W' for weeks.
   (disable = 0) Default is 30
 
+connection_retry_delay
+  When ``blocking_timeout`` is set, the cap (in milliseconds) for the exponential back-off between
```
fluca1978: it is not really exponential, see comment below. We should document better, maybe with an example in here.
In any case, it is doubled, right?

fluca1978 (changes requested):
Great, minor changes required.

src/libpgagroal/pool.c
fheikens: Done — reworded the inline comment to describe the delay as doubling each retry (1ms, 2ms, 4ms, ...) and dropped "exponential".

doc/man/pgagroal.conf.5.rst
fheikens: Done. The man page now describes the doubling and gives the worked sequence under the default 250ms cap: 1, 2, 4, 8, 16, 32, 64, 128, 250, 250, ... ms. Yes — it doubles each round from a 1ms seed up to the cap, then holds at the cap. Applied the same wording to CONFIGURATION.md and the manual so the three stay in sync.

src/include/pgagroal.h
fheikens: The value is validated and clamped to [1, 999] ms, so it is never negative at runtime. I kept the field as `int` to match every other field in `main_configuration` and the `to_int()` (de)serialization helper, which is `int`-typed — making just this one `unsigned int` would be inconsistent and would need a cast on the serialization path. Happy to switch it if you would prefer unsigned here; just say the word.

fheikens: Done — switched `connection_retry_delay` to `unsigned int` (commit e136be3). Builds clean under `-Werror`; the value is validated to [1, 999] so behaviour is unchanged.

fheikens:
All three review points are addressed:

- `connection_retry_delay` is now `unsigned int` (e136be3).
- The back-off is documented as *doubling each retry* (1, 2, 4, 8, … ms up to the cap, then held), with the worked sequence under the default 250 ms cap — wording kept in sync across the inline comment, `pgagroal.conf.5.rst`, `CONFIGURATION.md`, and the manual. Dropped the "exponential" framing.

CI is green across all platforms. Re-requesting review.

fluca1978:
@fheikens please squash, ready to merge.

fluca1978:
> @fheikens please squash, ready to merge.

@fheikens already merged, I got time to sqaush. Re-edited commit message, commit 2bb767b545904bafdbec656484bdfebd7af8d25d

PR #927, opened 2026-06-09
[#926] Size port buffer for 5-digit ports in pgagroal_connect

### LINKED ISSUE #926: pgagroal_connect: sport[5] too small — stack buffer overflow for 5-digit backend ports
### Summary

`pgagroal_connect()` formats the backend port into a 5-byte stack buffer that is one byte too small for 5-digit ports, so any backend (`[primary]`/server) listening on a port `>= 10000` overflows the buffer.

`src/libpgagroal/network.c`:

```c
char sport[5];
...
memset(&sport, 0, sizeof(sport));
sprintf(&sport[0], "%d", port);   // "54318" => 5 chars + NUL = 6 bytes into a 5-byte buffer
```

The maximum TCP port (65535) is 5 digits and needs 6 bytes including the NUL terminator. The sibling resolver in the same file already sizes this correctly:

```c
sport = calloc(1, 6);
sprintf(sport, "%d", port);
```

so this is an inconsistency rather than a design choice — line 253 simply predates (or missed) that sizing.

### Impact

- Any pgagroal instance whose server port is 5 digits (e.g. `54318`, `15432`, `25432`) writes one byte past `sport` on every `pgagroal_connect()`.
- On builds compiled with `_FORTIFY_SOURCE` (the default Debug build), `__sprintf_chk` detects the overflow and raises an abort. The event-loop signal handler intercepts that signal and runs the graceful-shutdown callback, so instead of crashing cleanly the worker re-enters the faulting `sprintf` and **spins at 100% CPU**; no backend connection is ever established.
- With the standard PostgreSQL port `5432` (4 digits) the buffer fits exactly, which is why CI and most deployments never hit it.

### Why CI does not catch it

The test suite runs PostgreSQL on port `5432` (4 digits, 5 bytes including NUL), so the buffer is never overflowed. Non-default high ports are common in real deployments but absent from CI.

### Reproduction

1. Run a PostgreSQL on a 5-digit port (e.g. `54318`).
2. Point `[primary]` at it and connect a single client through pgagroal.
3. On a Debug (`_FORTIFY_SOURCE`) build the worker spins at 100% CPU in `pgagroal_connect`; the client never receives a response.

### Fix

Size the buffer to hold the largest port plus the NUL terminator, matching the sibling function:

```c
char sport[6];
```

Found while reproducing #923 locally (the spin masked the actual pool behaviour until the buffer was sized correctly).

fheikens:
### What

`pgagroal_connect()` sized the backend port string as `char sport[5]`, one byte too small for a 5-digit port. `sprintf(sport, "%d", port)` writes six bytes (`"54318" + NUL`) for any port `>= 10000`, overflowing the stack buffer. Sized it to `char sport[6]`, matching the sibling resolver in the same file that already uses `calloc(1, 6)`.

Fixes #926.

### Why it matters

Any pgagroal pointing at a backend on a 5-digit port overflows on every connect. On `_FORTIFY_SOURCE` builds the abort is intercepted by the event-loop signal handler and the worker spins at 100% CPU instead of crashing, so no connection is established. The default port 5432 is four digits and fits, which is why CI (Postgres on 5432) never exercised it.

### Validation

- Before: a Debug build pointed at PostgreSQL 18 on port `54318` spins at 100% CPU in `pgagroal_connect` (confirmed by stack sample: `pgagroal_connect` -> `__sprintf_chk` -> `__chk_fail_overflow`); no client connection completes.
- After: the same client connects cleanly and the backend is established.
- `clang-format` clean; Debug build clean.

Note: `test/check.sh` runs its Postgres on a 4-digit port (`7432`), so the local suite does not exercise the 5-digit path; the fix was validated directly against a backend on a 5-digit port as above. CI runs the full suite on this PR.

### DIFF
--- a/src/libpgagroal/network.c
+++ b/src/libpgagroal/network.c
@@ -250,11 +250,11 @@ pgagroal_connect(const char* hostname, int port, int* fd, bool keep_alive, bool 
    int yes = 1;
    socklen_t optlen = sizeof(int);
    int rv;
-   char sport[5];
+   char sport[6];
    int error = 0;
 
    memset(&sport, 0, sizeof(sport));
-   sprintf(&sport[0], "%d", port);
+   pgagroal_snprintf(&sport[0], sizeof(sport), "%d", port);
 
    /* Connect to server */
    memset(&hints, 0, sizeof hints);


jesperpedersen:
We should use pgagroal_snprintf

fheikens:
Done in 9ebf253 — switched the port format to the bounded `pgagroal_snprintf(&sport[0], sizeof(sport), "%d", port)`, matching the other formatted writes in this file. The buffer stays at six bytes so the maximum port (65535) plus its NUL fits.

jesperpedersen:
@fheikens Did you check if `pgagroal_append_int` is enough ?

fheikens:
Yes, I checked it. `pgagroal_append_int` would work, but it heap-allocates via `pgagroal_append`, so `sport` becomes a `char*` that needs a NULL check (allocation failure) before `getaddrinfo` and a `free()` on every exit path of `pgagroal_connect`. Internally it also performs the same bounded `pgagroal_snprintf` into a stack buffer before copying to the heap, so on this path it is the current fix plus an allocation per connect.

Since the port is protocol-bounded (65535 plus NUL fits in six bytes), the allocation-free stack write seemed the better fit for the connection-establishment path. Happy to switch to the helper if you prefer it for consistency.

jesperpedersen:
@fluca1978 LGTM

jesperpedersen:
@fheikens This will need a cross-port

fluca1978:
merged, thanks @fheikens

PR #932, opened 2026-06-10
[#834] Read complete SCRAM messages before parsing

### LINKED ISSUE #834: Segmentation fault

fheikens:
## Problem

`pgagroal-cli` segfaults when a SCRAM message arrives split across network segments (#834). `read_message()` returns whatever a single `read()` yields, so the parser can see a partial `AuthenticationSASLContinue`; `sasl_continue->length - 9` then underflows (`size_t`) and the huge value reaches `get_scram_attribute()`, which crashes in `calloc()`/`memcpy()`. The same single-read assumption exists in every SCRAM exchange, and a message missing an expected attribute additionally leads to `strlen(NULL)`.

## Fix

Per the direction in the issue — keep loading until the message is complete, and bound-check before the offset arithmetic:

- New `pgagroal_read_complete_message()` in `message.c`: reads until the byte count declared in the message header (kind + int32 length) has arrived; a declared length below 4 bytes or beyond the parse buffer is rejected as an error.
- The SCRAM exchanges in `pgagroal_remote_management_scram_sha256()` (the reported path), `server_scram256()`, and the auth_query SCRAM flow now use it for the SASL continue/final reads.
- Every SCRAM parse site guards the message length before the `- 9` / `- 11` / `- 26` / `- 5` arithmetic and checks the extracted attributes for NULL, so a short or malformed message fails authentication cleanly instead of crashing. This includes the daemon-side parsing of client-supplied SCRAM messages.
- While there: `server_scram256()` did not check the read status before using the message, and the (currently unreferenced) `pgagroal_scram_client_authenticate()` ran its server-signature verification only when `pgagroal_extract_message()` failed — with a NULL `sasl_final`. Both aligned with the conventions used at the other sites.

## Tests

`test/testcases/test_message_complete.c` (MCTF, module `message_complete`) covers: a single-segment message, the #834 case of a message split across three segments with delays, a peer closing mid-message, and an invalid declared length. All four pass; the split-segment case fails without the new read behaviour.

`art` and `utf8` module failures observed locally are pre-existing on master (verified against a clean checkout) and unrelated.

close #834

### DIFF
--- a/src/include/message.h
+++ b/src/include/message.h
@@ -66,6 +66,21 @@ int
 pgagroal_read_block_message(SSL* ssl, int socket, struct message** msg);
 
 /**
+ * Read a complete typed message in blocking mode. The message must carry the
+ * standard header (1 byte kind + 4 byte length); the read continues until the
+ * number of bytes declared in the header has arrived, so a message split
+ * across multiple network segments is never returned partially. A message
+ * whose declared length is malformed (less than 4 bytes or larger than the
+ * parse buffer) is an error.
+ * @param ssl The SSL struct
+ * @param socket The socket descriptor
+ * @param msg The resulting message
+ * @return One of MESSAGE_STATUS_ZERO, MESSAGE_STATUS_OK or MESSAGE_STATUS_ERROR
+ */
+int
+pgagroal_read_complete_message(SSL* ssl, int socket, struct message** msg);
+
+/**
  * Read a message with a timeout
  * @param ssl The SSL struct
  * @param socket The socket descriptor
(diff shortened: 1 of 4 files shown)


fluca1978:
merged, thanks @fheikens

jesperpedersen:
@fheikens Did you figure out if you needed to cross-port ?

PR #933, opened 2026-06-16
Harden SCRAM iteration count

trxvorr:


src/libpgagroal/security.c
```diff
@@ -3592,6 +3602,60 @@ generate_nounce(char** nounce)
...
+   }
+
+   if (value > SCRAM_MAX_ITERATIONS)
+   {
+      pgagroal_log_error("SCRAM-SHA-256: iteration count %ld exceeds the maximum of %d",
+                         value, SCRAM_MAX_ITERATIONS);
+      return 1;
+   }
+
+   *iterations = (int)value;
```
fluca1978, line 3660: should check `value` boundaries to avoid integer overflow

fluca1978 (changes requested):
Looks good to me, but I'm not sure why we are using a long trimmed to an int, at least check for boundaries.

Also, please avoid empty commit messages.

jesperpedersen:
@fluca1978 LGTM
@trxvorr Please, cross-port

fluca1978:
merged, thanks @trxvorr

PR #934, opened 2026-06-17
 [#844] Add pause and resume feature

### LINKED ISSUE #844: PAUSE / RESUME for Connection Pools

AbduAllahhany:
# [#844] Add Pause and Resume Feature

## Overview
closes #844 
Follwing what we disscussed in [#890]. 
Implementing graceful and forced connection pool pausing and resuming for  `pgagroal`. It introduces the ability to suspend client traffic temporarily without dropping them, and seamlessly re-authenticates dead backend connections upon resumption if the database restarted while paused.

`pgagroal-cli pause [gracefully|all] [<server>|*] --timeout `
`pgagroal-cli resume [<server>|*]`

## Implementation Details

* **Worker Signal Handling (`worker.c`)**:
  * Rewrote the worker's execution flow by wrapping `pgagroal_event_loop_run()` in a `while (true)` state machine.
  * Mapped `SIGUSR1` to `pause_callback` (`WORKER_PAUSED`) and `SIGUSR2` to `resume_callback` (`WORKER_RESUME`).
  * When paused, it explicitly calls `pgagroal_io_pause(&client_io.io)` to halt reading from the client, holding them in suspense.

* **Re-authentication on Resume (`worker.c`)**:
  * When resuming, the worker actively checks if the backend is still alive using `pgagroal_socket_isvalid()`.
  * If the database backend died (e.g., from a DB restart during the pause), the worker drops the dead socket and calls `pgagroal_reauth_slot()` to transparently re-authenticate the client to a new backend connection, re-binding the event loop cleanly.

* **Cancel Requests (`pool.c`)**:
  * In `PAUSE_MODE_ALL`, the pool actively connects to the backend using a temporary socket (UNIX or TCP depending on config) and sends a PostgreSQL `CancelRequest`.
  * Immediately after sending the cancel message, it issues `kill(pid, SIGUSR1)` to force the worker into the paused state.
  * In `PAUSE_MODE_GRACEFULLY`, it simply lets active transactions finish naturally with timeout.

### DIFF
--- a/src/libpgagroal/ev.c
+++ b/src/libpgagroal/ev.c
@@ -71,6 +71,7 @@ static int (*loop_destroy)(void);
 
 static int (*io_start)(struct io_watcher*);
 static int (*io_stop)(struct io_watcher*);
+static int (*io_pause)(struct io_watcher*);
 
 static void signal_handler(int signum, siginfo_t* info, void* p);
 
@@ -92,6 +93,7 @@ static int ev_io_uring_setup_buffers(void);
 
 static int ev_io_uring_io_start(struct io_watcher*);
 static int ev_io_uring_io_stop(struct io_watcher*);
+static int ev_io_uring_io_pause(struct io_watcher*);
 
 static int ev_io_uring_periodic_init(struct periodic_watcher*, int64_t, int64_t);
 static int ev_io_uring_periodic_start(struct periodic_watcher*);
@@ -106,6 +108,7 @@ static int ev_epoll_handler(void*);
 
 static int ev_epoll_io_start(struct io_watcher*);
 static int ev_epoll_io_stop(struct io_watcher*);
+static int ev_epoll_io_pause(struct io_watcher*);
 static int ev_epoll_io_handler(struct io_watcher*);
 
 static int ev_epoll_periodic_init(struct periodic_watcher*, int64_t, int64_t);
@@ -123,6 +126,7 @@ static int ev_kqueue_handler(struct kevent*);
 
 static int ev_kqueue_io_start(struct io_watcher*);
 static int ev_kqueue_io_stop(struct io_watcher*);
+static int ev_kqueue_io_pause(struct io_watcher*);
 static int ev_kqueue_io_handler(struct kevent*);
 
 static int ev_kqueue_periodic_init(struct periodic_watcher*, int64_t, int64_t);
@@ -223,6 +227,7 @@ setup_ops(void)
       loop_start = ev_io_uring_loop;
       io_start = ev_io_uring_io_start;
       io_stop = ev_io_uring_io_stop;
+      io_pause = ev_io_uring_io_pause;
       periodic_init = ev_io_uring_periodic_init;
       periodic_start = ev_io_uring_periodic_start;
       periodic_stop = ev_io_uring_periodic_stop;
@@ -243,6 +248,7 @@ setup_ops(void)
       loop_start = ev_epoll_loop;
       io_start = ev_epoll_io_start;
       io_stop = ev_epoll_io_stop;
+      io_pause = ev_epoll_io_pause;
       periodic_init = ev_epoll_periodic_init;
       periodic_start = ev_epoll_periodic_start;
       periodic_stop = ev_epoll_periodic_stop;
@@ -257,6 +263,7 @@ setup_ops(void)
       loop_start = ev_kqueue_loop;
       io_start = ev_kqueue_io_start;
       io_stop = ev_kqueue_io_stop;
+      io_pause = ev_kqueue_io_pause;
       periodic_init = ev_kqueue_periodic_init;
       periodic_start = ev_kqueue_periodic_start;
       periodic_stop = ev_kqueue_periodic_stop;
@@ -435,6 +442,11 @@ pgagroal_event_loop_is_running(void)
 {
    return atomic_load(&loop->running);
 }
+bool
+pgagroal_io_paused(struct io_watcher* watcher)
+{
+   return watcher->paused;
+}
 
 int
 pgagroal_event_accept_init(struct io_watcher* watcher, int listen_fd, io_cb cb)
@@ -539,6 +551,74 @@ pgagroal_io_stop(struct io_watcher* watcher)
    return PGAGROAL_EVENT_RC_OK;
 }
 
+int
+pgagroal_io_pause(struct io_watcher* watcher)
+{
+   int found_idx = -1;
+
+   if (event_loop_called_from_child("pgagroal_io_pause"))
+      return PGAGROAL_EVENT_RC_OK;
+
+   assert(loop != NULL && watcher != NULL);
+
+   if (watcher->paused)
+   {
+      pgagroal_log_warn("pgagroal_io_pause: watcher already paused (fd rcv=%d)",
+                        watcher->fds.worker.rcv_fd);
+      return PGAGROAL_EVENT_RC_ERROR;
+   }
+
+   for (int i = 0; i < loop->events_nr; i++)
+   {
+      if (watcher == (struct io_watcher*)loop->events[i])
+      {
+         found_idx = i;
+         break;
+      }
+   }
+
+   if (found_idx < 0)
+   {
+      pgagroal_log_warn("pgagroal_io_pause: watcher not found (fd rcv=%d)",
+                        watcher->fds.worker.rcv_fd);
+      return PGAGROAL_EVENT_RC_ERROR;
+   }
+
+   int rc = io_pause(watcher);
+   if (rc != PGAGROAL_EVENT_RC_OK)
+   {
+      pgagroal_log_error("pgagroal_io_pause: io_pause failed %d", rc);
+      return rc;
+   }
+
+   watcher->paused = true;
+
+   for (int j = found_idx; j < loop->events_nr - 1; j++)
+      loop->events[j] = loop->events[j + 1];
+   loop->events_nr--;
+   loop->events[loop->events_nr] = NULL;
+
+   return PGAGROAL_EVENT_RC_OK;
+}
+int
+pgagroal_io_resume(struct io_watcher* watcher)
+{
+   if (event_loop_called_from_child("pgagroal_io_resume"))
+      return PGAGROAL_EVENT_RC_OK;
+
+   assert(loop != NULL && watcher != NULL);
+
+   if (!pgagroal_io_paused(watcher))
+   {
+      pgagroal_log_warn("pgagroal_io_resume: watcher not paused (fd rcv=%d)",
+                        watcher->fds.worker.rcv_fd);
+      return PGAGROAL_EVENT_RC_ERROR;
+   }
+
+   watcher->paused = false;
+
+   return pgagroal_io_start(watcher);
+}
 int
 pgagroal_periodic_init(struct periodic_watcher* watcher, periodic_cb cb,
                        int64_t msec, int64_t repeat_ms)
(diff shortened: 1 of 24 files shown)


fluca1978:
I placed a few comments inline, I need a few days to test this.

fluca1978:
Tested on epool and io_uring, works fine for me.
@AbduAllahhany with regard to `pgagroal-cli` output, should we place a timestamp for the last paused and last resumed? At least it would be easier to read for humans.
@jesperpedersen looks good to me, thoughts?

jesperpedersen:
Yes, we should have a timestamp for pause and resume and have those being part of the management protocol, and thereby displayed.

The follow-up to this PR is to work at pool level such that pause/resume a specific pool instead of a complete server

jesperpedersen:
@Userfrom1995 Can you sign-off on the ev.h changes ?

Userfrom1995:
> @Userfrom1995 Can you sign-off on the ev.h changes ?

Yes, I'll review the ev changes.

Userfrom1995:
While testing this, I observed two issues:

1. After pausing and resuming a transaction pipeline, the first query after the resume hangs. The remaining queries in that pipeline seem to execute correctly, and the other two pipeline modes appear to work fine.

2. After running a query on a transaction pipeline, the CLI becomes unresponsive. However, this does not appear to be related to this change, as I can reproduce the same behavior on `master`. This deserves a separate issue.

fluca1978:
Using commit 44f555bd37a3eaf32bc7965d4c5efdb11a8c6045
- `epoll` or `io_uring` + `transaction` if I pause while queries are running I got:
```
% pgagroal-cli pause rachel  
-> WARN  management.c:1428 pgagroal-cli: read_byte: (nil) 3 Connection reset by peer
```
- `epoll` + `session` works but pgagroal shutdown while paused leaved processes running
- same with `io_uring`
```
% pgagroal-cli shutdown                
Header:
  ClientVersion: 2.2.0
  Command: shutdown-gracefully
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260624120631
Outcome:
  Status: true
  Time: 00:00:00
Request:

# note that
-> WARN  main.c:2957 pgagroal: shutdown timeout expired - forcing immediate shutdown
-> INFO  main.c:1429 pgagroal: shutdown


% ps -auxw | grep pgagroal
luca       25848  4.5  1.1  14360  8652 pts/0    S+   12:05   0:04 pgagroal: main
luca       25871  2.0  0.3  14624  3064 pts/0    S    12:06   0:01 pgagroal: pgbench@127.0.0.1:5432/pgbench
luca       25872  2.6  0.3  14624  3064 pts/0    S    12:06   0:01 pgagroal: pgbench@127.0.0.1:5432/pgbench
luca       25873  2.3  0.3  14624  3064 pts/0    S    12:06   0:01 pgagroal: pgbench@127.0.0.1:5432/pgbench
luca       25874  2.2  0.3  14624  3064 pts/0    S    12:06   0:01 pgagroal: pgbench@127.0.0.1:5432/pgbench
luca       25875  2.3  0.3  14624  3064 pts/0    S    12:06   0:01 pgagroal: pgbench@127.0.0.1:5432/pgbench
```

so I suspect that while in pause state, the graceful shutdown is not working.

Last, still there is no timestamp in the paused -resumed information of the command line:

```
% pgagroal-cli pause rachel --timeout 5
Header:
  ClientVersion: 2.2.0
  Command: pause
  Compression: none
  Encryption: none
  Output: text
  Timestamp: 20260624120408
Outcome:
  Status: true
  Time: 00:00:00
Request:
  Mode: 0
  Server: rachel
  Timeout: 5
Response:
  ServerVersion: 2.2.0
  Servers:
    rachel:
      LastPaused: 1782295448   # should be 20260624....
      LastResumed: 0
```

AbduAllahhany:
@fluca1978  @Userfrom1995   

- check [#947] should fix all transaction pipeline issues.
- fixed stale processes and timestamp.

fluca1978:
LGTM, I've tested again and it works as expected.
@Userfrom1995 thoughts?

Userfrom1995:
> LGTM, I've tested again and it works as expected.
> @Userfrom1995 thoughts?

I haven't tested it yet. I'll test it again and get back to you.

src/libpgagroal/ev.c
```diff
@@ -1187,6 +1285,10 @@ ev_io_uring_handler(struct io_uring_cqe* cqe)
          break;
       case PGAGROAL_EVENT_TYPE_WORKER:
          io = (struct io_watcher*)watcher;
+
+         if (pgagroal_io_paused(io))
+            return PGAGROAL_EVENT_RC_OK;
+
```
Userfrom1995: We're losing data here. With `io_uring`, the kernel has already consumed the data from the socket by the time the CQE fires  early-returning discards it. Move the paused check down to the rearm guard instead:

```c
if (pgagroal_event_loop_is_running() && !pgagroal_io_paused(io))
{
   ev_io_uring_io_start(io);
}
```

src/libpgagroal/pool.c
```diff
@@ -1214,6 +1206,345 @@ pgagroal_request_flush(SSL* ssl __attribute__((unused)), int client_fd, uint8_t
...
+         pgagroal_log_error("pgagroal_pause: Unknown server %s", server);
+         goto done;
+      }
+   }
+
+   now = time(NULL);
+   if (config->pause_started == 0)
+   {
+      config->pause_started = now;
+   }
```
Userfrom1995: `config->pause_started = 0` unconditionally resets on every resume. If two servers are paused and you resume one, the counter drops to zero while the other is still paused. Was this intended? Should we only reset when no server remains paused?

src/libpgagroal/pool.c
```diff
@@ -1214,6 +1206,345 @@ pgagroal_request_flush(SSL* ssl __attribute__((unused)), int client_fd, uint8_t
...
+               pgagroal_write_terminate(NULL, config->connections[i].fd);
+            }
+            pgagroal_kill_connection(i, NULL);
+            break;
+
+         case STATE_IN_USE:
+            if (mode == PAUSE_MODE_ALL)
+            {
+               int ret;
+               int socket;
```
Userfrom1995: `int socket = -1;`  ?

src/libpgagroal/security.c
```diff
@@ -643,6 +639,167 @@ pgagroal_prefill_auth(char* username, char* password, char* database, int* slot,
...
+   int status = -1;
+   int ret = -1;
+   char* username = NULL;
+   char* database = NULL;
+   char* password = NULL;
+   int server = -1;
+
+   config = (struct main_configuration*)shmem;
+
+   *server_ssl = NULL;
```
Userfrom1995, line 670: `*server_ssl = NULL` without freeing the old context. 
Free it before you null it, maybe via `pgagroal_tls_from_ssl` + `pgagroal_tls_free`.
I think this function is the right place to own it since it takes `SSL**`.

src/libpgagroal/pool.c
AbduAllahhany: I removed pause_started and update the metrics which is collected

AbduAllahhany:
Here is the metrics the server collect
Screenshot from 2026-07-18 05-34-52

src/libpgagroal/worker.c
```diff
@@ -181,8 +194,73 @@ pgagroal_worker(int client_fd, char* address, char** argv)
...
+      {
+         pgagroal_event_loop_run();
+         if (exit_code == WORKER_PAUSED)
+         {
+            if (!paused)
+            {
+               pgagroal_log_debug("pgagroal_worker: slot %d paused", slot);
+               paused = true;
+               pgagroal_io_pause(&client_io.io);
+            }
```
Userfrom1995, line 206: You never pause `server_io` here. Shouldn't it be paused like the  `client_io `? 
Also once you pause it, `pgagroal_io_start` in the reauth path should become `pgagroal_io_resume`  and you need to resume `server_io` in the backend-alive path too?

AbduAllahhany: I didnot pause server_io so the server would send cancel message to the client when there is a long running transaction

AbduAllahhany: In the reauth path the server fd may change during the pause so we should use pgagroal_io_start

Userfrom1995: Agree, make sense.

src/libpgagroal/worker.c
```diff
@@ -181,8 +194,73 @@ pgagroal_worker(int client_fd, char* address, char** argv)
...
+
+            if (pgagroal_socket_isvalid(config->connections[slot].fd))
+            {
+               pgagroal_log_debug("pgagroal_worker: slot %d backend alive", slot);
+            }
+            else
+            {
+               pgagroal_log_debug("pgagroal_worker: slot %d backend gone, reauthenticating", slot);
+               pgagroal_disconnect(config->connections[slot].fd);
+
```
Userfrom1995, line 235: Shouldn't we call `pgagroal_io_stop(&server_io.io)` before `pgagroal_disconnect` + `pgagroal_event_worker_init` + `pgagroal_io_start`? 
Otherwise io_start appends a duplicate entry into loop->events?

Userfrom1995:
> While testing this, I observed two issues:
> 
>     1. After pausing and resuming a transaction pipeline, the first query after the resume hangs. The remaining queries in that pipeline seem to execute correctly, and the other two pipeline modes appear to work fine.
> 

I re-tested this, and the issue still occurs.

After pausing and resuming  transaction pipeline, the first query after the resume hangs. Subsequent queries in the same pipeline execute correctly, and the other two pipeline modes continue to work as expected.

AbduAllahhany:
@Userfrom1995 thanks for the review and testing

Fixed the first query after pause + resume by calling pgagroal_prefill_if_can(true, true) at the end of pgagroal_resume().

In transaction mode, pause kills the `FREE` backends and resume never refills. So the first client after resume opens its own backend, and when it's returned the socket moves to main `connections[].fd` now holds main's fd, which the worker doesn't have. It fails the check in `pgagroal_get_connection()` and retries forever [here](https://github.com/pgagroal/pgagroal/blob/master/src/libpgagroal/pool.c#L323) .

Prefilling on resume gives main a pooled backend before any new worker forks, so the worker has the descriptor and uses it normally same as `pgagroal_flush_server()` already does

I've noticed that if a worker was already running when pause killed a `FREE` slot, it still cannot use the new backend afterward (same fd problem; prefill only helps workers started later) ,`flush all` has the same hang for those older workers. I can open a separate issue for that if you want.

AbduAllahhany:
@fluca1978  @Userfrom1995  Any updates on this should I change anything else ?

Userfrom1995:
> @fluca1978 @Userfrom1995 Any updates on this should I change anything else ?

Sorry for the delay, I know it’s been a while since you pushed. I’ll test it again soon and let you know. I’ll make sure to test it this week.

fluca1978:
> @fluca1978 @Userfrom1995 Any updates on this should I change anything else ?

sorry, I got this out of my list, I'm going to test it again in the next days.

fluca1978:
Tested on epool and io_uring, seems to work as expected.
A couple of inline comments, nothing really important, just spaces.
Doing some more test, but seems there is the need for a rebase.

fluca1978:
@AbduAllahhany please consider `pgagroal_compare_string` instead of `strcmp` pretty much everywhere (except for command dispatching, where we used `strcmp`, we will need to adapt this too in a future PR).

fluca1978:
@Userfrom1995 this is generally fine for me, thoughts?

PR #937, opened 2026-06-22
Integrate TLS wrapper into connection path

trxvorr:
Drive client and backend TLS through the socket-decoupled wrapper, replacing the SSL_set_fd-bound contexts in the auth and pooling paths.

### DIFF
--- a/src/libpgagroal/security.c
+++ b/src/libpgagroal/security.c
@@ -37,6 +37,7 @@
 #include <prometheus.h>
 #include <security.h>
 #include <server.h>
+#include <tls.h>
 #include <tracker.h>
 #include <utils.h>
 #include <utf8.h>
@@ -134,7 +135,6 @@ static int server_signature(char* password, char* salt, int salt_length, int ite
                             unsigned char** result, size_t* result_length);
 
 static bool is_tls_user(char* username, char* database);
-static int create_ssl_client(SSL_CTX* ctx, char* key, char* cert, char* root, int socket, SSL** ssl);
 static int establish_client_tls_connection(int server, int fd, SSL** ssl);
 static int create_client_tls_connection(int fd, SSL** ssl, char* tls_key_file, char* tls_cert_file, char* tls_ca_file);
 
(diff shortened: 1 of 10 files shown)


src/libpgagroal/security.c
```diff
@@ -4461,21 +4470,20 @@ create_ssl_client(SSL_CTX* ctx, char* key, char* cert, char* root, int socket, S
 }
 
 int
-pgagroal_create_ssl_server(SSL_CTX* ctx, char* key_file, char* cert_file, char* ca_file, int socket, SSL** ssl)
+configure_server_ctx(SSL_CTX* ctx, char* key_file, char* cert_file, char* ca_file)
```
jesperpedersen: static ?

src/libpgagroal/security.c
```diff
@@ -4461,21 +4470,20 @@ create_ssl_client(SSL_CTX* ctx, char* key, char* cert, char* root, int socket, S
...
 {
-   SSL* s = NULL;
    STACK_OF(X509_NAME)* root_cert_list = NULL;
 
    if (strlen(cert_file) == 0)
    {
       pgagroal_log_error("No TLS certificate defined");
-      goto error;
+      return 1;
```
jesperpedersen: We use `goto error;`

src/libpgagroal/tls.c
```diff
@@ -144,6 +150,26 @@ pgagroal_tls_free(struct tls* tls)
    free(tls);
 }
 
+void
+pgagroal_tls_set_fd(struct tls* tls, int fd)
+{
+   if (tls != NULL)
+   {
+      tls->fd = fd;
+   }
```
jesperpedersen: And, if `NULL` ?

jesperpedersen (changes requested):

jesperpedersen:
@trxvorr Shouldn't `tls.h` be a layer above `security.h` ? So, most of the `security.c` changes should be in `tls.c` instead

trxvorr:
the pre-existing socket-bound creators pgagroal_create_ssl_server / create_ssl_client still live in security.c, but they now source their config from the tls layer. Did you want those relocated into tls.c as well?

jesperpedersen:
@trxvorr Yes, move them up

fluca1978:
Looks good to me, apart from the code movement requested.

jesperpedersen:
Merged.

PR #944, opened 2026-07-01
[#942] Support backend passwords longer than 256 characters

### LINKED ISSUE #942: Support backend passwords longer than 256 characters (cloud IAM DB-auth tokens)
## Problem

pgagroal caps stored passwords at `MAX_PASSWORD_CHARS = 256` (UTF-8 chars) in a `char password[MAX_PASSWORD_LENGTH]` buffer where `MAX_PASSWORD_LENGTH = 1024` (`src/include/pgagroal.h`). The users-file loader (`pgagroal_read_users_configuration` in `src/libpgagroal/configuration.c`) and `pgagroal-admin` both enforce this and **reject** an over-length entry.

This makes pgagroal unable to front a PostgreSQL instance that uses **cloud IAM database authentication**, where the "password" is a short-lived token far longer than 256 chars:

- **AWS RDS/Aurora IAM auth** tokens are presigned SigV4 URLs, ~1.8 KB (we observed 1874 chars → `Password too long (1874 characters). Maximum allowed: 256 characters.`).
- **Azure AD** and **GCP IAM** access tokens (JWTs) are similar, ~1–4 KB.

## Why it matters

Passwordless / short-lived IAM DB auth is an increasingly common security best practice on managed Postgres (no static DB passwords). A pooler that can't hold these tokens can't sit in front of an IAM-authenticated RDS/Aurora/Azure/GCP database. (The current limit appears to originate from #138, and the UTF-8 char-count from #51 — apologies if there's context we've missed.)

## Candidate approaches (seeking your preference, not presuming)

1. Raise `MAX_PASSWORD_LENGTH` / `MAX_PASSWORD_CHARS` to accommodate ~4 KB credentials.
2. Make the limit compile-time or runtime configurable.
3. A dedicated long-credential path for backend auth, if raising the global buffer is undesirable for memory/DoS reasons.

## Questions before we propose a PR

- Is raising this limit acceptable in principle, and is there a deliberate reason for 256 we should design around?
- Which approach would you accept?
- Are there **other** buffers/paths that also bound effective password length (backend auth-message construction, SCRAM, startup packet) that a fix would need to cover?

Context: we build a managed distribution on top of pgagroal and hit this fronting RDS with IAM auth. Happy to prepare the PR once you confirm the direction you'd accept.

fheikens:
Raises the 256-character backend password limit so pgagroal can front a Postgres backend that uses cloud IAM database authentication, where the "password" is a short-lived token (~1.8 KB for AWS RDS/Aurora IAM, 1-4 KB for Azure AD / GCP).

Follows the direction in #942: raise the limit and clean up the duplicated constant.

## Changes
- **Consolidate + raise the caps** (`pgagroal.h`): `MAX_PASSWORD_LENGTH` 1024 -> 8192, `MAX_PASSWORD_CHARS` 256 -> 2048. Removed the duplicate `MAX_PASSWORD_CHARS` define in `configuration.c` so the header is the single source of truth.
- **Credential-file line buffer**: the users/frontend/admins/superuser/vault readers store `username:base64(AES-256-GCM(password))`. A long token expands under encryption + base64 past the 512-byte `LINE_LENGTH` used for ordinary config lines, so `fgets` would truncate the entry. Those five readers now use a dedicated `MAX_USER_LINE_LENGTH` (16384); `LINE_LENGTH` (and the section-name array that shares it) is unchanged.
- **Backend auth message**: `server_password()` copies the outgoing `PasswordMessage` into the per-connection `security_messages` slot, bounded by `SECURITY_BUFFER_SIZE`. Raised it 1024 -> 16384 so a max-length credential's message fits, and added a bounds check before the copy (mirroring the existing check for the received message) so an over-size message errors instead of overrunning the buffer.

## Memory note
`security_messages` is `NUMBER_OF_SECURITY_MESSAGES x SECURITY_BUFFER_SIZE` per connection, so this raises the per-connection footprint by ~75 KB (5 x 15 KB). If you'd prefer a smaller default I'm happy to make `SECURITY_BUFFER_SIZE` configurable, or to set `MAX_PASSWORD_LENGTH` to 4096 (still covers every known IAM token) with `SECURITY_BUFFER_SIZE` at 8192 - whichever you'd like.

## Tests
Adds `test/testcases/test_password_length.c`:
- accept: an IAM-token-sized (~1.8 KB) password round-trips through the encrypted users file and is stored intact;
- reject: a password beyond the character cap is still refused;
- invariant: `SECURITY_BUFFER_SIZE` holds a PasswordMessage carrying a max-length credential.

close #942

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -471,9 +471,9 @@ master_key(char* password, bool generate_pwd, int pwd_length, int32_t output_for
                      password = NULL;
                      continue;
                   }
-                  if (char_count > MAX_PASSWORD_CHARS)
+                  if (strlen(password) >= MAX_PASSWORD_LENGTH)
                   {
-                     printf("Master key too long (%zu characters). Maximum allowed: %d characters.\n", char_count, MAX_PASSWORD_CHARS);
+                     printf("Master key too long (%zu bytes). Maximum allowed: %d bytes.\n", strlen(password), MAX_PASSWORD_LENGTH - 1);
                      free(password);
                      password = NULL;
                      continue;
@@ -745,9 +745,9 @@ password:
       password = NULL;
       goto password;
    }
-   if (char_count > MAX_PASSWORD_CHARS)
+   if (strlen(password) >= MAX_PASSWORD_LENGTH)
    {
-      warnx("Password too long (%zu characters). Maximum allowed: %d characters.", char_count, MAX_PASSWORD_CHARS);
+      warnx("Password too long (%zu bytes). Maximum allowed: %d bytes.", strlen(password), MAX_PASSWORD_LENGTH - 1);
       if (do_free)
       {
          free(password);
@@ -796,9 +796,9 @@ password:
          password = NULL;
          goto password;
       }
-      if (verify_char_count > MAX_PASSWORD_CHARS)
+      if (strlen(verify) >= MAX_PASSWORD_LENGTH)
       {
-         warnx("Verification password too long (%zu characters). Maximum allowed: %d characters.", verify_char_count, MAX_PASSWORD_CHARS);
+         warnx("Verification password too long (%zu bytes). Maximum allowed: %d bytes.", strlen(verify), MAX_PASSWORD_LENGTH - 1);
          free(verify);
          verify = NULL;
          if (do_free)
@@ -1060,9 +1060,9 @@ password:
             password = NULL;
             goto password;
          }
-         if (char_count > MAX_PASSWORD_CHARS)
+         if (strlen(password) >= MAX_PASSWORD_LENGTH)
          {
-            warnx("Password too long (%zu characters). Maximum allowed: %d characters.", char_count, MAX_PASSWORD_CHARS);
+            warnx("Password too long (%zu bytes). Maximum allowed: %d bytes.", strlen(password), MAX_PASSWORD_LENGTH - 1);
             if (do_free)
             {
                free(password);
@@ -1108,9 +1108,9 @@ password:
                password = NULL;
                goto password;
             }
-            if (verify_char_count > MAX_PASSWORD_CHARS)
+            if (strlen(verify) >= MAX_PASSWORD_LENGTH)
             {
-               warnx("Verification password too long (%zu characters). Maximum allowed: %d characters.", verify_char_count, MAX_PASSWORD_CHARS);
+               warnx("Verification password too long (%zu bytes). Maximum allowed: %d bytes.", strlen(verify), MAX_PASSWORD_LENGTH - 1);
                free(verify);
                verify = NULL;
                if (do_free)
--- a/src/cli.c
+++ b/src/cli.c
@@ -773,9 +773,9 @@ username:
          warnx("pgagroal-cli: Invalid UTF-8 encoding in password");
          goto done;
       }
-      if (char_count > MAX_PASSWORD_CHARS)
+      if (strlen(password) >= MAX_PASSWORD_LENGTH)
       {
-         warnx("pgagroal-cli: Password too long (max %d characters)", MAX_PASSWORD_CHARS);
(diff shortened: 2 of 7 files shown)


jesperpedersen:
@trxvorr We need a review of this - especially the

```
src/include/pgagroal.h:#define MAX_PASSWORD_LENGTH                      1024
src/include/pgagroal.h:#define MAX_PASSWORD_CHARS                       256
src/libpgagroal/configuration.c:#define MAX_PASSWORD_CHARS 256 /* Maximum UTF-8 characters in password */
```

part... _LENGTH should be enough...

fluca1978:
Is this impacting also the vault in any way?

fheikens:
Thanks both.

@jesperpedersen agreed - `MAX_PASSWORD_LENGTH` is enough. I've dropped `MAX_PASSWORD_CHARS` entirely (latest push): the byte-length limit plus the existing UTF-8 validity checks now cover it, so there's a single limit and the duplicate define is gone. Updated every enforcement site - the five credential loaders, `sasl_prep()`, and the pgagroal-admin / pgagroal-cli / pgagroal-vault password-entry paths (their length messages now report bytes). The minimum-length check is unchanged.

@fluca1978 yes, the vault is covered consistently: `pgagroal_vault_read_users_configuration` uses the same larger line buffer, so it can load a long user credential too, and its `struct user` password buffer grows with `MAX_PASSWORD_LENGTH` (one user, negligible). The vault has no connection pool, so it doesn't touch `security_messages` / `SECURITY_BUFFER_SIZE` - the per-connection memory note doesn't apply to it.

trxvorr:
reviewed the length part and agree MAX_PASSWORD_LENGTH is enough on its own, the char cap was just a redundant second limit and dropping it is fine. UTF-8 validation is still handled separately in sasl_prep so nothing's lost there.

only thing id double check is SECURITY_BUFFER_SIZE at 16384 a max password message is only ~8KB (right?), so 8192 would do and saves some memory per connection.

LGTM otherwise

jesperpedersen:
@trxvorr There are headers too, so we need to calculate the maximum size. However, we should try and have our structs defined in a way where they are cache line aligned

trxvorr:
@jesperpedersen ah you're right. so 16384 makes sense both to fit the header and to stay 64-byte aligned. my bad, scratch that suggestion

fheikens:
@jesperpedersen both points addressed:

- **Constant cleanup** (0768e4c): dropped `MAX_PASSWORD_CHARS` entirely — `MAX_PASSWORD_LENGTH` is now the single password-length limit with one definition in `pgagroal.h`, raised to 8192 bytes to fit cloud IAM DB-auth tokens (AWS RDS/Aurora, Azure AD, GCP). Updated the call sites in `admin.c`/`cli.c`/`configuration.c`/`security.c`/`vault.c` and the `test_password_length` test accordingly.
- **Cross-port**: filed pgexporter/pgexporter#561 and pgmoneta/pgmoneta#1221 for the same longer-token support + constant consolidation, so the three projects stay aligned.

fluca1978:
LGTM, @jesperpedersen can we merge or do you have any other request?

jesperpedersen:
@fluca1978 LGTM
@fheikens Check if documentation need updating as well
@fheikens This need to be cross-ported

fheikens:
@jesperpedersen Checked the docs — no update needed for the length change. No user-facing doc references the backend password length limit (the only "password length" docs are the admin `-l/--length` random-password generator and `rotate_frontend_password_length`, both unrelated), and the change is internal constants only. The cross-port I'll handle separately.

fluca1978:
@fheikens please fix commit message and we can merge

fheikens:
@fluca1978 Squashed into a single commit (1ff69b3) with a message reflecting the final state — MAX_PASSWORD_CHARS is removed rather than raised, MAX_PASSWORD_LENGTH is the single byte limit. Ready to merge.

fluca1978:
merged, thanks @fheikens

Issue #824, opened 2026-03-30
pgagroal enters high CPU loop when Unix socket is missing in multi-instance setup


ankushT369:
**Description**

I am running multiple pgagroal instances on the same machine using the same unix_socket_dir (/tmp). Both instances end up using the same Unix sockets (.s.pgagroal and .s.pgagroal.tu).

When one instance is stopped, the socket files get removed. The second instance is still running but can no longer access the expected socket.

After some time (around ~14 minutes of runtime), the remaining instance starts showing the issue:

repeated attempts to connect to the Unix socket
continuous log spam (especially with debug level enabled)
noticeable CPU usage increase

**Expected reason**

It seems that when one instance is stopped, the shared Unix socket files are removed, leaving the other instance with invalid internal references. Since both instances use the same socket path, there is no isolation, and the remaining instance cannot recover properly.

PR #947, opened 2026-07-01
[#946]Fix pgagroal-cli failure in transaction mode when clients are c…

### LINKED ISSUE #946: Transaction pipeline worker binds MAIN_UDS and breaks pgagroal-cli when clients are connected
In pipeline = transaction mode, pgagroal-cli commands such as flush all fail with Connection reset by peer when at least one client is connected. The same commands work when no clients are connected.

  CLI fails with:

`  WARN pgagroal-cli: read_byte: (nil) 3 Connection reset by peer`

AbduAllahhany:
Summary

  Revert the transaction worker socket change from commit 9dd22c7 (#824 (https://github.com/pgagroal/pgagroal/issues/824)) that switched workers from .s.pgagroal.<pid> to MAIN_UDS. That caused workers to steal the CLI management socket `(.s.pgagrshooal.<port>)`, breaking pgagroal-cli with Connection reset by peer when clients are connected in transaction mode.

  Restore per-worker UDS in pipeline_transaction.c; keep the #824 port suffix in network.c.

closes #946

### DIFF
--- a/src/libpgagroal/pipeline_transaction.c
+++ b/src/libpgagroal/pipeline_transaction.c
@@ -120,7 +120,7 @@ transaction_start(struct event_loop* loop, struct worker_io* w)
    deallocate = false;
 
    memset(&p, 0, sizeof(p));
-   pgagroal_snprintf(&p[0], sizeof(p), MAIN_UDS);
+   pgagroal_snprintf(&p[0], sizeof(p), ".s.pgagroal.%d", (int)getpid());
 
    if (pgagroal_bind_unix_socket(config->unix_socket_dir, &p[0], &unix_socket))
    {
@@ -587,7 +587,7 @@ shutdown_mgt(struct event_loop* loop __attribute__((unused)))
    config = (struct main_configuration*)shmem;
 
    memset(&p, 0, sizeof(p));
-   pgagroal_snprintf(&p[0], sizeof(p), MAIN_UDS);
+   pgagroal_snprintf(&p[0], sizeof(p), ".s.pgagroal.%d", (int)getpid());
 
    pgagroal_io_stop(&io_mgt);
    pgagroal_disconnect(unix_socket);
--- a/src/libpgagroal/pipeline_transaction.c
+++ b/src/libpgagroal/pipeline_transaction.c
@@ -120,6 +120,7 @@ transaction_start(struct event_loop* loop, struct worker_io* w)
    deallocate = false;
 
    memset(&p, 0, sizeof(p));
+   /* Use a per-worker socket, not MAIN_UDS: binding MAIN_UDS here would collide with the main management socket and reset pgagroal-cli once a client is attached */
    pgagroal_snprintf(&p[0], sizeof(p), ".s.pgagroal.%d", (int)getpid());
 
    if (pgagroal_bind_unix_socket(config->unix_socket_dir, &p[0], &unix_socket))
@@ -587,6 +588,7 @@ shutdown_mgt(struct event_loop* loop __attribute__((unused)))
    config = (struct main_configuration*)shmem;
 
    memset(&p, 0, sizeof(p));
+   /* Must match the per-worker socket bound in transaction_start, not MAIN_UDS, so cleanup removes the right socket */
    pgagroal_snprintf(&p[0], sizeof(p), ".s.pgagroal.%d", (int)getpid());
 
    pgagroal_io_stop(&io_mgt);
--- a/test/check.sh
+++ b/test/check.sh
@@ -432,12 +432,16 @@ export_pgagroal_test_variables() {
 
   echo "export PGPASSWORD=$PG_USER_PASSWORD"
   export PGPASSWORD=$PG_USER_PASSWORD
+
+  echo "export PGAGROAL_TEST_BIN=$EXECUTABLE_DIRECTORY"
+  export PGAGROAL_TEST_BIN=$EXECUTABLE_DIRECTORY
 }
 
 unset_pgagroal_test_variables() {
   unset PGAGROAL_TEST_BASE_DIR
   unset PGAGROAL_TEST_CONF
   unset PGPASSWORD
+  unset PGAGROAL_TEST_BIN
   unset LLVM_PROFILE_FILE
   unset CC
 }
--- a/test/include/tsclient.h
+++ b/test/include/tsclient.h
@@ -119,6 +119,31 @@ pgagroal_tsclient_execute_concurrent_holds(char* user, char* database, int clien
 int
 pgagroal_tsclient_limit_backend_peak(char* user, char* database, int client_count, int hold_seconds);
 
+/**
+* Run a pgagroal-cli management command against the running test instance,
+ * using the testsuite configuration so the CLI targets the same management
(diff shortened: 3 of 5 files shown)


fluca1978:
@ankushT369 can you give a feedback on this?

ankushT369:
> @ankushT369 can you give a feedback on this?

yes !!! I checking this out!!

ankushT369:
@fluca1978 heres the thing the actual problem was !! 

The actual issue was this: earlier `./pgagroal` was using a single socket file `.s.pgagroal`. Because of that, if I ran two instances, both were sharing the same socket, which caused conflicts and basically led to an infinite loop problem.

To fix this, I introduced isolation by making the socket name include the port, like `.s.pgagroal.PORT` (for example: `.s.pgagroal.2345`, where `2345` is the port number). I also updated the `pipeline_transaction` code to keep everything consistent.

Previously, `pipeline_transaction` was using `.s.pgagroal.pid`, which is also a valid way to isolate instances. But somehow the client side wasn’t aware of this change, and that mismatch was causing the issue.

I would rather suggest to edit client side where problem is occurring if you want to maintain consistency or if you don't want to then going with this will work!!.
 
I exactly don't know whats the issue.

AbduAllahhany:
The issue was the transaction workers bind MAIN_UDS (.s.pgagroal.<port>), the same socket pgagroal-cli uses so when there is an active client the cli got `connection reset by peer` . so, we need separate sockets for CLI and workers

jesperpedersen:
The transactional pipeline should use its own Unix Domain Socket descriptors

ankushT369:
> The transactional pipeline should use its own Unix Domain Socket descriptors

the before was : `.s.pgagroal.pid` then I changed it to `.s.pgagroal.PORT` !! and thats where the client side got confused

jesperpedersen:
@ankushT369 Yes, but transaction need to interact with their PID UDS

ankushT369:
> The issue was the transaction workers bind MAIN_UDS (.s.pgagroal.), the same socket pgagroal-cli uses so when there is an active client the cli got `connection reset by peer` . so, we need separate sockets for CLI and workers

No it does not binds `.s.pgagroal.` the function 
```c
if (pgagroal_bind_unix_socket(config->unix_socket_dir, &p[0], &unix_socket))
```
converts the MAIN_UDS (`.s.pgagroal.`) to (`.s.pgagroal.PORT`)

ankushT369:
> @ankushT369 Yes, but transaction need to interact with their PID UDS

then the change would be fine !!

fheikens (commented):
Confirmed the root cause and reproduced the reasoning: #824 (\`9dd22c7\`) switched the transaction worker to \`MAIN_UDS\`, so it collides with the main management socket and resets \`pgagroal-cli\` once a client is attached. The revert to a per-worker \`.s.pgagroal.<pid>\` is the right minimal fix, and it is applied consistently in both \`transaction_start\` (bind) and \`shutdown_mgt\` (cleanup), while keeping #824's per-instance suffix in \`network.c\`. I also checked \`pipeline_session.c\` — it never binds \`MAIN_UDS\` (only \`.s.PGSQL.%d\`), so this is correctly transaction-only.

Two asks before merge: (1) could we add an MCTF regression test — a \`pgagroal-cli\` command (e.g. \`flush all\`) in transaction mode with a client connected, which should now succeed? That locks this in so #824-style changes cannot silently reintroduce it. (2) Minor: worth a one-line comment noting why the worker must not use \`MAIN_UDS\`, so the next person does not re-apply #824's change. LGTM otherwise.

fheikens:
Following up on my review's test ask — offering a starting point so this doesn't have to block on it. It needs a small `tsclient` `pgagroal-cli` helper (there isn't one yet) plus the suite running in **transaction** mode. Untested sketch, adapt as you see fit.

**1. A `pgagroal-cli` helper in `tsclient` — `test/libpgagroaltest/tsclient.c` (+ prototype in `test/include/tsclient.h`):**

```c
/* Run a pgagroal-cli management command against the test instance.
 * Returns 0 if pgagroal-cli exited successfully, otherwise 1. */
int
pgagroal_tsclient_execute_cli(char* args)
{
   char* cmd = NULL;
   char* conf = NULL;
   int ret;

   /* project_directory is set in pgagroal_tsclient_init(); the running instance
    * uses <project_directory>/pgagroal-testsuite/conf/pgagroal.conf */
   conf = pgagroal_append(conf, project_directory);
   conf = pgagroal_append(conf, PGAGROAL_CONFIGURATION_TRAIL);

   cmd = pgagroal_append(cmd, "pgagroal-cli -c ");
   cmd = pgagroal_append(cmd, conf);
   cmd = pgagroal_append_char(cmd, ' ');
   cmd = pgagroal_append(cmd, args);

   ret = system(cmd);

   free(cmd);
   free(conf);
   return (ret == 0) ? 0 : 1;
}
```

**2. Compose it with a live client** — the bug only shows while a transaction worker exists, so hold one client and run the CLI mid-hold. This reuses the exact background-hold machinery already in `pgagroal_tsclient_execute_concurrent_holds` (`BEGIN; SELECT pg_sleep(n); COMMIT;` on a thread); a thin `pgagroal_tsclient_cli_during_hold(user, database, args, hold_seconds)` that spawns one hold thread, `sleep(1)`, calls `pgagroal_tsclient_execute_cli(args)`, then joins and returns the CLI result.

**3. Testcase — `test/testcases/test_cli_transaction.c`:**

```c
#include <pgagroal.h>
#include <tsclient.h>
#include <mctf.h>

/*
 * Regression test for #946: in pipeline = transaction, a pgagroal-cli command
 * failed with "Connection reset by peer" whenever a client was connected,
 * because the transaction worker bound MAIN_UDS and stole the management
 * socket. This keeps one client connected and asserts the CLI succeeds.
 * Requires the suite to run with pipeline = transaction.
 */
MCTF_TEST(test_pgagroal_cli_transaction_client_connected)
{
   int rc = pgagroal_tsclient_cli_during_hold(user, database, "status", 3);
   MCTF_ASSERT(rc == 0, cleanup,
               "pgagroal-cli failed while a client was connected in transaction mode (#946)");
cleanup:
   MCTF_FINISH();
}
```

Two wiring points: register the testcase in `test/CMakeLists.txt` + `test/runner.c`, and make sure it runs under **`pipeline = transaction`** (if the suite is session-only today, that config path is the prerequisite). Happy to open this as a follow-up PR against your branch if that's easier — just say the word.

AbduAllahhany:
Ok, go ahead

fluca1978:
it is not clear to me if this requires other work or it's done

AbduAllahhany:
> it is not clear to me if this requires other work or it's done

I think it's done as @fheikens stated that "treat CI as the real validation"

fheikens (approved):
Confirming this is complete from my side. The per-worker UDS revert is the correct minimal fix, and the transaction-mode `pgagroal-cli` regression test (`22aac4b7`) that reproduces #946 is now in the branch and passing in CI — which is the real validation here. No further work needed; good to merge. Thanks @AbduAllahhany.

fluca1978:
merged, thanks @AbduAllahhany

PR #949, opened 2026-07-01
[#941] Configurable server_reset_query to prevent session-state leakage

### LINKED ISSUE #941: Transaction-mode connection reset (server_reset_query) to prevent session-state leakage
## Summary

In transaction pooling mode a backend connection can be handed to a different client between transactions. Without a reset, session state set by one client — GUCs (`SET`), temporary tables, prepared statements, advisory locks, `LISTEN`/`NOTIFY` registrations — can leak to the next client sharing that backend. pgagroal should reset connection state on return in transaction mode, controlled by a `server_reset_query` (as PgBouncer does).

## Proposal

- `server_reset_query` (string, default `DISCARD ALL`) — statement issued before a returned connection is reused.
- `server_reset_query_always` (bool, default off) — also run it in session mode (opt-in parity).

### Behaviour
- Transaction mode: run `server_reset_query` on return, while the connection is idle (never mid-transaction).
- Empty `server_reset_query` disables the reset (opt-out; document as unsafe for transaction mode).
- If the reset query errors, invalidate/close the connection and open a fresh one rather than returning dirty state to the pool.

### Invariant
No session state (`SET`, temp tables, prepared statements, advisory locks, `LISTEN`) survives across two clients on the same backend when reset is enabled.

### Design notes / caveats
- `DISCARD ALL` must run outside a transaction and deallocates prepared statements — so protocol-level prepared statements don't survive a reset (the same trade-off PgBouncer documents). Worth a docs note; a lighter reset set could be a follow-up.
- Reset adds ≤1 round-trip per transaction return; the A/B performance check can keep it within budget.

### Acceptance
- Default txn mode: `DISCARD ALL` issued on return; client B doesn't observe client A's `SET`/temp table.
- Empty reset query: no reset issued.
- Reset query fails: connection discarded, not reused.

Happy to open a PR for this.

fheikens:
Implements #941: a configurable reset statement run on a backend connection before it is reused, so session state (GUCs, temp tables, prepared statements, advisory locks, `LISTEN`) from one client can't leak to the next client sharing that backend.

## Options
- `server_reset_query` (string, default `DISCARD ALL`) - the statement to run; empty disables the reset.
- `server_reset_query_always` (bool, default off) - also run it in transaction pooling.

## Semantics (PgBouncer parity)
The reset runs **in session pooling by default** and **in transaction pooling only when `server_reset_query_always` is on** - matching PgBouncer, where transaction-pooling clients are not expected to rely on session state.

> Note: this corrects the `server_reset_query_always` description in the issue, which had it the other way round. pgagroal already ran a hardcoded `DISCARD ALL` on return in session mode (`pgagroal_return_connection`), so making the reset session-default keeps that behavior and just makes the statement configurable; forcing it into transaction mode is the opt-in. Following @fluca1978's PgBouncer-compatibility point, this is the least-surprising mapping.

## Implementation
- Session pooling: the previously hardcoded `DISCARD ALL` in `pgagroal_return_connection` now uses `server_reset_query` (empty = skip); failure still discards the connection.
- Transaction pooling: when `server_reset_query_always` is set, the reset runs on transaction return and **subsumes** the existing `DEALLOCATE ALL` step (`DISCARD ALL` ⊇ `DEALLOCATE ALL`). On reset failure the backend is invalidated (`pgagroal_kill_connection`) rather than returned dirty.
- `pgagroal_write_discard_all` is generalized to `pgagroal_write_reset_query(ssl, socket, query)`.

## Tests / coverage
- Unit tests (`test_configuration.c`): parsing of both options, boolean validation, and the `DISCARD ALL` default.
- The end-to-end session-state-isolation behavior is exercised by the reset mechanism, which reuses the proven `DISCARD ALL`/`DEALLOCATE ALL` return-path machinery. A dedicated MCTF integration test (two clients on one backend) would need a raw-SQL exec helper in `tsclient` - happy to add that as a follow-up if you'd like it in-tree.

## Follow-up
Per @fluca1978's suggestion, making the reset-failure behavior tunable (discard vs. warn-and-continue) is left as a separate change to keep this focused; I'll file an issue.

close #941


doc/man/pgagroal.conf.5.rst
```diff
@@ -210,6 +210,12 @@ tracker
 track_prepared_statements
   Track prepared statements (transaction pooling). Default is off
 
+server_reset_query
+  Statement run on a backend connection before it is reused. Runs in session pooling by default; an empty value disables it. Default is DISCARD ALL
```
fluca1978: wording: "before is is reused" or "as soon as it is released"?

doc/man/pgagroal.conf.5.rst
```diff
@@ -210,6 +210,12 @@ tracker
...
+
+server_reset_query_always
+  Also run server_reset_query in transaction pooling. Off by default, since transaction-pooling clients should not rely on session state. Default is off
```
fluca1978: "Run `server_reset_query` in any configured pipeline" ?

doc/manual/en/04-configuration.md
```diff
@@ -72,6 +72,8 @@ The available keys and their accepted values are reported in the table below.
 | hugepage | `try` | String | No | Huge page support (`off`, `try`, `on`) |
 | tracker | off | Bool | No | Track connection lifecycle |
 | track_prepared_statements | off | Bool | No | Track prepared statements (transaction pooling) |
+| server_reset_query | DISCARD ALL | String | No | Statement run on a backend connection before it is reused. Runs in session pooling by default; an empty value disables it |
+| server_reset_query_always | off | Bool | No | Also run server_reset_query in transaction pooling. Off by default, since transaction-pooling clients should not rely on session state |
```
fluca1978: same descriptive comments as above.

doc/CONFIGURATION.md
```diff
@@ -90,6 +90,8 @@ The available keys and their accepted values are reported in the table below.
 | hugepage | `try` | String | No | Huge page support (`off`, `try`, `on`) |
 | tracker | off | Bool | No | Track connection lifecycle |
 | track_prepared_statements | off | Bool | No | Track prepared statements (transaction pooling) |
+| server_reset_query | DISCARD ALL | String | No | Statement run on a backend connection before it is reused. Runs in session pooling by default; an empty value disables it |
+| server_reset_query_always | off | Bool | No | Also run server_reset_query in transaction pooling. Off by default, since transaction-pooling clients should not rely on session state |
```
fluca1978: ditto

src/libpgagroal/message.c
```diff
@@ -730,25 +730,25 @@ pgagroal_write_deallocate_all(SSL* ssl, int socket)
 }
 
 int
-pgagroal_write_discard_all(SSL* ssl, int socket)
+pgagroal_write_reset_query(SSL* ssl, int socket, char* query)
```
fluca1978: `query` is always `config->server_reset_query`, do we need it as an argument?

fluca1978 (commented):
Looks good to me, only a few comments for minor doubts.

@jesperpedersen ok?

fheikens:
Thanks @fluca1978 — addressed all of it in the latest push:

- **`message.c` — dropped the `query` argument.** You're right that it was always `config->server_reset_query`, so `pgagroal_write_reset_query(SSL*, int socket)` now reads it straight from the running configuration (same `shmem` access already used elsewhere in this file). Both callers in `pool.c` and `pipeline_transaction.c` are updated, and each still guards on a non-empty value before calling.
- **Docs wording (all three: `pgagroal.conf.5.rst`, `04-configuration.md`, `CONFIGURATION.md`).** Reworded `server_reset_query` to "run … as soon as it is released back to the pool" (which is when it actually fires in session pooling — on return, not just before next reuse), and `server_reset_query_always` to "run server_reset_query in the transaction pipeline too, not only in session pooling."

Rebuilt clean and re-ran clang-format on the changed files. @jesperpedersen it's ready for your look.

fluca1978:
merged, thanks @fheikens

Issue #923, opened 2026-05-29
Connection handler stalls under per-rule max_size oversubscription (performance pipeline)


fheikens:
### Summary

Under the **performance** pipeline, when a per-rule `max_size` cap is oversubscribed by concurrent clients, at least one connection handler appears to stall indefinitely: the client never receives a response and never disconnects, while the main daemon stays healthy. Observed in CI; not yet reproduced locally, so this is a report from log evidence rather than a confirmed root cause.

### How it surfaced

PR #907 adds `test_pgagroal_pool_max_size_cap`, which oversubscribes a rule with `2 * max_size` concurrent psql holds (`BEGIN; SELECT pg_sleep(3); COMMIT;`) against `test/conf/release-only/18` (`pipeline = performance`, `max_connections = 24`, rule `myuser/mydb max_size = 10`, `blocking_timeout = 10`). The Linux Release `run-configs` job hung for ~25 minutes until cancellation.

### Evidence (config-18 daemon log, CI run 26410433108)

- 16:38:56 — 20 clients connect; the cap correctly serializes them (two waves of 10).
- 16:38:56 → 16:39:03 — the test's metric sampling loop runs to completion; `pgagroal_limit{type="backend"}` never exceeds `max_size` (the cap itself works).
- ~16:39:03 — all client/backend/metrics activity stops.
- 16:43–17:03 — only the 5-minute `idle_timeout` maintenance fires; the main daemon is alive and idle the whole time.
- 17:04 — CI cancels the job.

The test then sat in `pthread_join`: the metric window had closed, but at least one psql client never exited. The main daemon shows no further work, consistent with a **forked connection handler** stalled while a client connection stays open — invisible in the main daemon log.

### Hypothesis

A connection handler can stall under `2 * max_size` oversubscription on the performance pipeline with a per-rule cap (interaction with the #848/#903 hard-cap reservation path is the most likely area), leaving the client hung in `recv()`. Needs local reproduction to confirm whether it is a lost wakeup, a reservation accounting edge, or a handler that never completes the response.

### Reproduction (proposed)

1. Start pgagroal with `pipeline = performance` and a per-rule `max_size = N` (no prefill).
2. Drive `2 * N` concurrent `BEGIN; SELECT pg_sleep(3); COMMIT;` sessions for that rule.
3. Watch for a client that never returns while the main daemon is idle.

### Notes

[...]

PR #952, opened 2026-07-02
923 event accept stall

fheikens:


jesperpedersen:
@fheikens Can you convert the shell script into a test case, or just remove it ?
@Userfrom1995 Please, look at the ev.c changes

fheikens:
Thanks @jesperpedersen — removed the `repro-923.sh` harness commit, so the PR is now just the `ev.c` fix. The script was a dev-only reproduction tool; the accept-stall it demonstrated is already covered by the drain logic here, and a proper MCTF test for the io_uring liveness path is better tracked separately under #923 rather than shipped as a shell script.

@Userfrom1995 the `ev.c` change is the only diff now — ready for your look.

Userfrom1995:
The changes look good, but there is one critical issue: the listener socket is blocking by default. Because of this, the while(true) accept loop will hang the main daemon forever as soon as the backlog is drained.
I actually verified this by running it on `kqueue`, and the process hangs immediately on the very first request.

To fix this, you must set the socket to non-blocking during initialization. You can do this by simply adding `pgagroal_socket_nonblocking(listen_fd)` inside `pgagroal_event_accept_init()`.

fheikens:
Good catch, thank you — you're right. The drain loop relies on `accept()` returning `EAGAIN`/`EWOULDBLOCK` to stop, but `pgagroal_bind()` leaves the listener blocking (only the systemd socket-activation path set it non-blocking), so the final `accept()` blocked the daemon once the backlog drained.

Fixed in `23f75ad`: set the listener non-blocking in `pgagroal_event_accept_init()`, the single point every accepting socket is registered through. That covers kqueue's drain loop, and it's also correct for epoll (already treats `EAGAIN` as "no more pending") and harmless to the io_uring multishot accept. Built clean on macOS/kqueue; would appreciate a re-check on your kqueue setup that the hang is gone.

fheikens:
Rebased onto master. The FreeBSD CI red was an infra flake in the pkg-repo update step (fixed on master by #964), not the `ev.c` change — the branch just predated that fix. Now green across all platforms including both FreeBSD variants.

@Userfrom1995 the non-blocking-listener fix from your review is in; ready for a re-check when you have a moment.

Userfrom1995:
@fluca1978 @jesperpedersen LGTM. If you're both okay with it, feel free to merge. Otherwise, let me know and I can merge it.

@fheikens Could you please squash those commits into a single commit?

jesperpedersen:
@fheikens Squash, and cross-port

fheikens:
Squashed into a single commit (`b74236f`) — the two changes (kqueue accept-backlog drain + non-blocking listener) are now one commit with a combined message; the tree is unchanged from the reviewed state. Thanks for the review @Userfrom1995.

@jesperpedersen on the cross-port: this fix lives in `ev.c`, which pgexporter and pgmoneta share. I'll open tracking issues + PRs against those repos for the same kqueue accept-drain / non-blocking-listener fix once this lands (keeping the port faithful to the merged version).

jesperpedersen:
@fheikens Yes, you will handle the cross-port for the projects that has ev.c - @Userfrom1995 will handle the rest

fheikens:
Gentle nudge on this one — it's been ready since the squash on 23 Jul:

- Single commit `b74236f` (kqueue accept-backlog drain + non-blocking listener), per @jesperpedersen's squash request.
- All 9 checks green (Linux gcc/clang, macOS, FreeBSD clang Debug/Release, rpm, format).
- @Userfrom1995's non-blocking-listener fix is in (`23f75ad`) and re-reviewed LGTM.
- I re-tested the merge locally against current master: the `ev.c` change applies cleanly, no conflict despite the branch being a few commits behind.

@Userfrom1995 / @jesperpedersen — happy for you to merge whenever it suits. Once it lands I'll open the cross-port issues + PRs for the same fix against pgexporter and pgmoneta.

Userfrom1995:
@fheikens Thank you for your contribution.

PR #957, opened 2026-07-04
Pool backend TLS 1.3 sessions

trxvorr:
After the OpenSSL handshake pgagroal takes over a serializable TLS 1.3 record layer, so a backend session is parked in shared memory on return and resumed on a later checkout without a new handshake. Scoped to the session and performance pipelines; enforces TLS 1.3 with an AEAD cipher whitelist and cleanses parked secrets when a slot is killed.


jesperpedersen:
Merged.

Issue #958, opened 2026-07-06
CI: run the integration test suite on macOS (kqueue backend)


fheikens:
### Problem

The integration test suite only runs on Linux. The `build-macos` and FreeBSD CI jobs build pgagroal but do not run `check.sh` / the testsuite:

- `build-linux` builds **and** runs `check.sh run-configs-ci-nonbuild` — but Linux only exercises the **epoll** and **io_uring** event backends.
- `build-macos` runs `cmake` + `make` only.
- `build-freebsd` ("Build and Test") also runs `cmake` + `make` only.

As a result the **kqueue** event backend (macOS/BSD) has no functional test coverage in CI — it is compiled but never exercised. Every integration test runs exclusively under epoll/io_uring.

This is not hypothetical: #923's accept-drain fix had a kqueue-only regression (a blocking listener hung the daemon once the backlog drained, fixed in #952). The existing `test_pool_saturation` test drives `max_connections + 2` concurrent sessions — exactly the accept-burst shape that triggers that path — but because it never runs under kqueue, CI could not catch it. A reviewer found it manually on their Mac.

### Proposal

Run the existing integration suite on a kqueue platform in CI. `macos-latest` can run PostgreSQL natively via Homebrew, so the lowest-effort path is to add the testsuite step to the existing `build-macos` job (install PostgreSQL, then `bash ./build/check.sh run-configs-ci-nonbuild`, mirroring the Linux job).

This gives every existing integration test kqueue coverage with no new test code.

### Scope

**In scope**
- Add a PostgreSQL install + testsuite run step to `build-macos` (or a dedicated macOS test job).
- Ensure the configs that set `ev_backend` still make sense on macOS (kqueue is the platform backend there; io_uring/epoll configs should skip cleanly rather than fail).

**Out of scope**
- New test cases (tracked separately).
- FreeBSD test execution (QEMU-based; can follow once the macOS path is proven).

### Acceptance criteria

- [ ] CI runs the integration suite on a kqueue platform (macOS).
- [ ] A failure in the accept/connection path under kqueue turns the job red.
- [ ] Backend-specific configs skip cleanly where a backend is unavailable on the platform, rather than failing the run.

Pranav-error:
I'd like to take this one. I develop on macOS, so the kqueue path is what I actually run, and I hit the gap this describes from the other direction recently — a macOS-only build break in a sibling project that Linux CI could not have seen.

[...]

PR #960, opened 2026-07-06
[#959] Add a per-test timeout to MCTF so hangs fail fast

### LINKED ISSUE #959: test: add a per-test timeout to MCTF so hangs fail fast
### Problem

MCTF has no per-test timeout. `MCTF_TEST` registers a test but there is no `SIGALRM`/watchdog bounding how long a single test may run. When a test hangs (for example a daemon that blocks on `accept()` and never returns — the failure mode behind #923/#952), the test does not fail with a useful signal; it stalls until the **job-level** timeout (30 minutes) fires and kills the whole run.

A 30-minute red with no indication of which test hung, or why, is a poor signal. A hang should fail fast and name the offending test.

### Proposal

Add an optional per-test timeout to MCTF. When a test exceeds its budget, abort it and record a clear failure (e.g. "test X exceeded Ns") rather than letting it stall the run.

Suggested shape:
- A default per-test timeout (generous, e.g. the largest `blocking_timeout` in the test configs plus a margin) enforced via `SIGALRM` around each `MCTF_TEST` body.
- A way for a test to override the budget when it legitimately needs longer (the saturation tests hold connections for several seconds).
- On expiry: mark the test failed with a timeout message and continue to the next test where possible.

### Scope

**In scope**
- Per-test timeout mechanism in `test/libpgagroaltest/mctf.c` / `test/include/mctf.h`.
- A default budget and a per-test override.
- A distinct timeout failure message in the report.

**Out of scope**
- Changing individual test logic beyond setting an override where needed.

### Acceptance criteria

- [ ] A test that hangs fails within its timeout with a message naming the test, instead of stalling until the job timeout.
- [ ] Tests with legitimately long holds (saturation) can set a larger budget and still pass.
- [ ] The timeout failure is distinguishable from an assertion failure in the report.

fheikens:
Closes #959.

### Problem

MCTF has no per-test time bound. When a test hangs — e.g. a daemon blocked in `accept()`, the failure mode behind #923/#952 — it does not fail with a useful signal; it stalls the run until the 30-minute CI job timeout fires, with no indication of which test hung.

### Change

Wrap each test invocation in a `SIGALRM`-based watchdog. On expiry the handler `siglongjmp`s back to the runner, which records a distinct `[TIMEOUT]` failure naming the test and its budget (separate from an assertion failure):

```
  test_timeout_probe_hangs (00:00:02,004) [TIMEOUT] (test_zzz_timeout_probe.c) - exceeded 2 seconds
...
Failed tests:
  - test_timeout_probe_hangs (...) - Timed out after 2 seconds
```

- Default budget `MCTF_DEFAULT_TIMEOUT_SECONDS = 300` — generous enough that no healthy test trips it (the slowest integration tests spawn a handful of psql sessions and finish well under a minute), while turning an infinite hang into a fast, named failure.
- Tests that legitimately run longer can override with a new `MCTF_TEST_TIMEOUT(name, seconds)` macro; `MCTF_TEST` is unchanged and uses the default.
- The test runner does not otherwise use `SIGALRM` (it installs only `SIGABRT`/`SIGSEGV` handlers), so there is nothing to clash with.

### Verification

- A probe test that `sleep(30)` under a 2-second budget is aborted at **2.004s** and reported `[TIMEOUT]`, exit code 1 — not run to completion.
- Existing unit modules (`deque`, `json`, ...) still pass with no false timeouts.

### Context

This is one of two test-hardening follow-ups from the #952 review, where a kqueue-only accept hang had to be found by hand. It pairs with #958 (run the integration suite on macOS/kqueue): the timeout ensures that if such a hang recurs under the new kqueue job, CI fails fast and names the test instead of stalling for 30 minutes.

### DIFF
--- a/test/include/mctf.h
+++ b/test/include/mctf.h
@@ -42,6 +42,15 @@ extern "C" {
 
 /* Special error codes */
 #define MCTF_CODE_SKIPPED (-1)
+#define MCTF_CODE_TIMEOUT (-2)
+
+/* Default per-test wall-clock budget, in seconds. Generous enough that no
+ * healthy test trips it (the slowest integration tests spawn a handful of
+ * psql sessions and finish well under a minute), while still turning a hung
+ * test into a fast, named failure instead of stalling until the CI job
+ * timeout. Individual tests can raise this with MCTF_TEST(name, secs) or
+ * opt out entirely with MCTF_TEST_NO_TIMEOUT(name). */
+#define MCTF_DEFAULT_TIMEOUT_SECONDS 300
 
 /* Filter types */
 typedef enum {
@@ -58,11 +67,12 @@ typedef int (*mctf_test_func_t)(void);
  */
 typedef struct mctf_test
 {
-   char* name;             /**< Human readable test name. */
-   char* module;           /**< Logical module/group this test belongs to. */
-   char* file;             /**< Source file where the test is defined. */
-   mctf_test_func_t func;  /**< Function implementing the test. */
-   struct mctf_test* next; /**< Pointer to the next test in the registration list. */
+   char* name;                   /**< Human readable test name. */
+   char* module;                 /**< Logical module/group this test belongs to. */
+   char* file;                   /**< Source file where the test is defined. */
+   mctf_test_func_t func;        /**< Function implementing the test. */
+   unsigned int timeout_seconds; /**< Per-test wall-clock budget in seconds (0 = no timeout). */
+   struct mctf_test* next;       /**< Pointer to the next test in the registration list. */
 } mctf_test_t;
 
 /**
@@ -75,6 +85,7 @@ typedef struct mctf_result
    int line;              /**< Line number of the failing assertion, or 0 on success. */
    bool passed;           /**< True if the test completed successfully. */
    bool skipped;          /**< True if the test was explicitly skipped. */
+   bool timed_out;        /**< True if the test was aborted by its timeout. */
    int error_code;        /**< Error code associated with a failure or skip. */
    char* error_message;   /**< Dynamically allocated, human readable error message. */
    long elapsed_ms;       /**< Execution time of the test in milliseconds. */
@@ -109,6 +120,11 @@ mctf_cleanup(void);
 /* Register a test (called automatically via constructor) */
 void
 mctf_register_test(const char* name, const char* module, const char* file, mctf_test_func_t func);
+
+/* Register a test with a non-default per-test timeout (seconds; 0 = no timeout) */
+void
+mctf_register_test_with_timeout(const char* name, const char* module, const char* file,
+                                mctf_test_func_t func, unsigned int timeout_seconds);
 
 /* Extract module name from file path */
 const char*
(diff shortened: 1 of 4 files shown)


jesperpedersen:
@shashidhar-bm You need to sign-off on this
@fheikens This need to be cross-ported once approved

test/include/mctf.h
```diff
@@ -62,6 +70,7 @@ typedef struct mctf_test
    char* module;           /**< Logical module/group this test belongs to. */
    char* file;             /**< Source file where the test is defined. */
    mctf_test_func_t func;  /**< Function implementing the test. */
+   int timeout_seconds;    /**< Per-test wall-clock budget in seconds. */
```
fluca1978: `unsigned int`?

test/libpgagroaltest/mctf.c
```diff
@@ -275,6 +293,14 @@ mctf_cleanup(void)
 /* Register a test */
 void
 mctf_register_test(const char* name, const char* module, const char* file, mctf_test_func_t func)
+{
+   mctf_register_test_with_timeout(name, module, file, func, MCTF_DEFAULT_TIMEOUT_SECONDS);
```
fluca1978, line 297: This means we are always using a timeout, is it worth providing a `mftf_register_test_without_timemout` function? Just asking.

test/libpgagroaltest/mctf.c
```diff
@@ -442,7 +470,33 @@ mctf_run_tests(mctf_filter_type_t filter_type, const char* filter)
...
+      sa_timeout.sa_handler = mctf_timeout_handler;
+      sigemptyset(&sa_timeout.sa_mask);
+      sa_timeout.sa_flags = 0; /* no SA_RESTART: let the alarm interrupt blocking calls */
+      sigaction(SIGALRM, &sa_timeout, &sa_prev);
+
+      if (sigsetjmp(g_timeout_jmp, 1) == 0)
+      {
+         alarm((unsigned int)test->timeout_seconds);
+         ret = test->func();
+         alarm(0);
```
fluca1978: `alarm(0)` out of the `if-else` block? It is executed in any case.

fluca1978 (changes requested):
I like the idea, minor changes but it is already done according to me.
Clearly this has to be cross-ported.

fheikens:
Both review changes pushed in b7c3da1: the timeout is now unsigned int end to end (it feeds alarm(2) directly, so the cast is gone too), and alarm(0) is hoisted below the if-else since it runs on both paths.

On mctf_register_test_without_timeout: I'd rather not add an opt-out — the point of the change is that no test can hang the runner, and an escape hatch reintroduces exactly that. A test that legitimately needs longer can pass a bigger budget via MCTF_TEST_TIMEOUT. Happy to add the variant if you see a case a larger budget doesn't cover.

@jesperpedersen ack on the cross-port once this is approved, and I'll coordinate with @shashidhar-bm on the sign-off.

shashidhar-bm:
@fheikens Yes, this is great!
and this is something I've been thinking about for a long time. We do have something similar in pgmoneta like `mctf_register_test_with_max_time` (it seems it was never cross-ported to pgagroal., at some point we should compare between frameworks and patch up the missing pieces). But that one is purely for catching performance regressions of a specific test case,  it doesn't stop the test in the middle, it just reports a failure if the test exceeded the threshold time after it completes. So it doesn't help with hung tests at all.

test/libpgagroaltest/mctf.c
shashidhar-bm: @fluca1978 ,This is something I was also wondering about, whether to make the timeout specific to individual test cases, or whether we can assume with a fair enough threshold such that any test exceeding it can only be a failure. 😅

shashidhar-bm: cc: @jesperpedersen

fheikens:
Thanks @shashidhar-bm! Agreed they're complementary: pgmoneta's mctf_register_test_with_max_time is a post-completion performance assertion, while this is a wall-clock kill switch — a hung test never reaches the max_time check at all, which is exactly the gap this fills. Having both eventually makes sense.

On per-test vs a single fair threshold (@fluca1978's question too): the PR does both. MCTF_DEFAULT_TIMEOUT_SECONDS (300s) is the global ceiling every test gets with no annotation, deliberately generous because this is a hang detector, not a perf budget — any test hitting it can only be a failure. MCTF_TEST_TIMEOUT exists only to raise it for the rare legitimately-long test, so the default stays tight-ish without an opt-out.

Happy to file a follow-up to diff the two MCTF implementations and cross-port the missing pieces in both directions — that would also cover the cross-port @jesperpedersen asked for once this lands.

test/libpgagroaltest/mctf.c
fluca1978: > @fluca1978 ,This is something I was also wondering about, whether to make the timeout specific to individual test cases, or whether we can assume with a fair enough threshold such that any test exceeding it can only be a failure. 😅

I believe having the test running with a timeout as default is good, then if we want a test to run without a timeout we need to explicitly set it.

test/include/mctf.h
```diff
@@ -159,6 +174,21 @@ mctf_format_error(const char* format, ...);
    }                                                                    \
    static int test_name##_impl(void)
 
+/* Test registration macro with an explicit per-test timeout (seconds). Use for
+ * tests that legitimately run longer than MCTF_DEFAULT_TIMEOUT_SECONDS. */
+#define MCTF_TEST_TIMEOUT(test_name, secs)                                \
```
shashidhar-bm: Since `MCTF_TEST` now always applies the default timeout, could we make the timeout an optional parameter, such as `MCTF_TEST(name, [secs])`, instead of introducing a separate `MCTF_TEST_TIMEOUT` macro? Just a suggestion.

That said, as @fluca1978 suggested, it would still be worthwhile to add a dedicated macro for cases where tests should run without any timeout.

shashidhar-bm (commented):
@fheikens , please update the docs as well.

fheikens:
Both suggestions are in 708846e, plus the docs update:

- `MCTF_TEST` now takes the timeout as an optional parameter: `MCTF_TEST(name)` uses the default, `MCTF_TEST(name, secs)` sets an explicit budget. The separate `MCTF_TEST_TIMEOUT` macro is gone (@shashidhar-bm).
- `MCTF_TEST_NO_TIMEOUT(name)` is the explicit opt-out @fluca1978 asked for. Internally a zero budget schedules no alarm, so the test runs untimed. Since the opt-out has to be spelled out at the registration site, the default stays timeout-on for everything else.
- doc/TEST.md and doc/manual/en/78-test.md still listed built-in timeouts as a framework limitation; both now document the three registration forms instead.

Also rebased onto current master for the FreeBSD 14.4 CI fix.

One data point from running the full suite on macOS while validating this: 8 of the 123 tests (the pgbench/psql-driven load tests) hit a pre-existing macOS-specific stall in the pooler and hung. Each one failed fast as a named `[TIMEOUT]` after 300s and the suite ran to completion with a clean summary — previously the first hang would have stalled the run indefinitely. The stall itself is unrelated to this PR and doesn't reproduce on Linux; those tests pass in CI.

fluca1978:
LGTM, @shashidhar-bm ok to merge or do you have any other suggestion?

shashidhar-bm:
> LGTM, @shashidhar-bm ok to merge or do you have any other suggestion?

@fluca1978 , LGTM  as well :)

fluca1978:
@fheikens squash with appropriate4 message and we are done

fheikens:
Squashed into a single commit as requested; the tree is unchanged from the reviewed state. Ready to merge when you are.

fluca1978:
merged, thnks @fheikens

PR #967, opened 2026-07-19
[#965] Import the aarch64 PGDG signing key in the test PostgreSQL image

### LINKED ISSUE #965: Test harness PostgreSQL image fails to build on aarch64: PGDG repo RPM lacks the aarch64 signing key
Building the test harness PostgreSQL image (`test/postgresql/src/postgresql17/Dockerfile`) from scratch currently fails: the PGDG yum repository no longer passes GPG verification with the key the repo RPM ships. Anyone without a cached Docker layer for that image cannot run `test/check.sh` in dev mode.

**Observed** (2026-07-11, Docker on macOS/Apple Silicon, so the aarch64 EL-10 repo):

`RUN dnf -y upgrade` fails on the repository metadata check:

```
Importing GPG key 0x08B40D20:
 Userid     : "PostgreSQL RPM Repository <pgsql-pkg-yum@lists.postgresql.org>"
 Fingerprint: D4BF 08AE 67A0 B4C7 A1DB CCD2 40BC A2B4 08B4 0D20
 From       : /etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-RHEL
Error: Failed to download metadata for repo 'pgdg-common': repomd.xml GPG signature verification error: Signing key not found
```

With `repo_gpgcheck` disabled to get past the metadata, the package check fails next, on the repo RPM itself:

```
Key imported successfully
Import of key(s) didn't help, wrong key(s)?
Public key for pgdg-redhat-repo-42.0-65.rhel10.2PGDG.noarch.rpm is not installed. Failing package is: pgdg-redhat-repo-42.0-65.rhel10.2PGDG.noarch
 GPG Keys are configured as: file:///etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-RHEL
Error: GPG check FAILED
```

So the repository contents are now signed with a key that is not the `PGDG-RPM-GPG-KEY-RHEL` installed by the current `pgdg-redhat-repo-latest.noarch.rpm` — this looks like a PGDG signing-key rotation, not anything in this repo. A `--no-cache` rebuild reproduces it; cached layers from before the rotation still work, which can hide the problem until a cache goes cold.

**CI exposure:** `build-linux` in `ci.yml` installs from the same repo on every run (`dnf install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-10-x86_64/pgdg-redhat-repo-latest.noarch.rpm` followed by `dnf install postgresql17 ...`), with no image cache to fall back on. CI is green today, so the x86_64 metadata apparently still verifies (or the affected path differs), but if the rotation reaches the x86_64 repo the same way, every CI run breaks at PostgreSQL install.

**Workaround I verified locally** (to keep running the suite): disable GPG checking for the pgdg repos after installing the repo RPM —

```dockerfile
RUN sed -i "s/^repo_gpgcheck.*/repo_gpgcheck=0/; s/^gpgcheck.*/gpgcheck=0/" /etc/yum.repos.d/pgdg-redhat-all.repo
```

That is not a fix I'd propose committing. The proper fix is importing the current PGDG signing key explicitly (e.g. `rpm --import` of the updated key from download.postgresql.org) in the Dockerfile — and the same in `ci.yml` if it starts failing — once PGDG publishes/settles the new key. Happy to open a PR for that if useful.

fheikens:
Fixes #965.

On aarch64 the PGDG repositories are signed with a separate per-architecture key (`PGDG-RPM-GPG-KEY-AARCH64-RHEL`, `177B343BB9738825`), but `pgdg-redhat-repo-latest` installs only the x86_64 RHEL key and references only it in `gpgkey=`. So on ARM the test image fails at `dnf makecache` and the `postgresql17` install with "Signing key not found" (x86_64 is unaffected, which is why CI has been green).

This imports the aarch64 key into `/etc/pki/rpm-gpg/`, adds it to `gpgkey=` in `pgdg-redhat-all.repo`, and `rpm --import`s it — keeping `repo_gpgcheck`/`gpgcheck` fully enabled on both architectures. On x86_64 it is just an extra trusted key, so no arch-conditional logic is needed.

Verified in a fresh Rocky 10 aarch64 container: `dnf makecache` and `dnf install postgresql17-libs` both pass with repo and package GPG checks on.

The upstream-side fix (the PGDG repo RPM shipping the aarch64 key) is tracked with @devrimgunduz in #965; this unblocks the test image independently.

### DIFF
--- a/test/postgresql/src/postgresql17/Dockerfile
+++ b/test/postgresql/src/postgresql17/Dockerfile
@@ -36,6 +36,14 @@ ENV PGROOT="/usr/pgsql-${PGVERSION}"
 
 RUN rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
 RUN dnf -y install --nogpgcheck --allowerasing https://download.postgresql.org/pub/repos/yum/reporpms/EL-10-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+# The PGDG aarch64 repositories are signed with a separate per-architecture key
+# (PGDG-RPM-GPG-KEY-AARCH64-RHEL) that the repo RPM does not install, so on
+# aarch64 `dnf makecache` and the postgresql17 install fail with "Signing key
+# not found". Import it and add it to gpgkey= so GPG checking stays enabled on
+# both architectures (a harmless extra trusted key on x86_64). See #965.
+RUN curl -so /etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-AARCH64-RHEL https://download.postgresql.org/pub/repos/yum/keys/PGDG-RPM-GPG-KEY-AARCH64-RHEL && \
+    sed -i 's|^gpgkey=.*|& file:///etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-AARCH64-RHEL|' /etc/yum.repos.d/pgdg-redhat-all.repo && \
+    rpm --import /etc/pki/rpm-gpg/PGDG-RPM-GPG-KEY-AARCH64-RHEL
 RUN dnf -qy --nogpgcheck module disable postgresql 2>/dev/null || true
 RUN dnf -y upgrade
 RUN dnf install -y dnf-plugins-core


fluca1978:
merged, thanks @fheikens

Issue #950, opened 2026-07-01
Make server_reset_query failure behavior tunable (discard vs. warn-and-continue)


fheikens:
Follow-up to #941 / #949.

When `server_reset_query` errors on a returned backend connection, #949 invalidates the connection (`pgagroal_kill_connection`) rather than returning dirty state to the pool. That is the safe default.

@fluca1978 suggested (in #941) that this be **tunable**, so an operator can choose between:

- **discard** (current behavior) — kill the connection and open a fresh one; safest, but a persistently failing reset query turns into repeated connection churn.
- **warn-and-continue** — log a warning and treat the reset as disabled going forward, reusing the connection anyway. This makes a broken/misconfigured `server_reset_query` visible without tearing down connections, at the cost of potential state leakage (so it must be clearly documented as unsafe).

### Proposal
Add a setting, e.g. `server_reset_query_on_failure = discard | ignore` (default `discard`), applied in both the session path (`pgagroal_return_connection`) and the transaction path (`pipeline_transaction`).

Split out of #949 to keep that PR focused. Happy to implement once there's agreement on the name and whether `ignore` is desirable at all.

fluca1978:
I would like to have:
- `discard` default behavior as described
- `try` do a try on the next round (and every round), so a lot of query failures and connections killed;
- `ignore` as described, ignore going forward.

Rationale: if the user notice the errors, it could fix `server_reset_query` and `pgagroal-cli conf reload` and then see the right one applied at runtime (scenario `try`).

@fheikens of course, you are free to choose the naming of tuanbles and values.

MohamedM216:
Can I start working on this?

jesperpedersen:
@MohamedM216 Go for it !

MohamedM216:
@jesperpedersen @fluca1978 Hi, I've raised a PR to add this feature as discussed here. Looking forward to your feedback. Thanks!

PR #970, opened 2026-07-30
[#950] Make server_reset_query failure behavior tunable (discard, try, ignore)

MohamedM216:
Reference issue: #950 
Follow-up to #941 / #949.

This PR adds a new configuration option, `server_reset_query_on_failure`, to control what happens when `server_reset_query` fails on a returned backend connection. 

**Options:**
*   **`discard` (default):** Invalidates and kills the connection. Safest behavior, prevents session-state leakage.
*   **`try`:** Kills the connection on failure, but will attempt the reset query again on the next connection. Creates visible connection churn if misconfigured, helping operators notice and fix the issue.
*   **`ignore`:** Logs a warning and reuses the connection anyway. Avoids connection churn but is explicitly documented as unsafe for transaction pooling due to potential session-state leakage.


fluca1978:
Looks good to me, but see inline comments.
I'm not comfortable with the name of this feature and tunable `server_reset_query_on_failure` suggests to my brain "if there is a failure then do server reset query", while this is not the aim.
Would it be clearer if renamed to `on_failure_server_reset_query` or `failure_of_server_reset_query` or `server_reset_query_behavior_on_failure` (probably too complex to digit)?

jesperpedersen:
I think latter (`server_reset_query_behavior_on_failure`)

MohamedM216:
Hi @fluca1978 , thanks for the feedback. I addressed the requested changes in https://github.com/pgagroal/pgagroal/compare/cc3748fb9c4eb2e51f2d145d7752e9c52965a07d..f045e7ab48da0f5a53143b8920da2ff26f9a0925. I settled on this name `server_reset_query_behavior_on_failure`

test/testcases/test_configuration.c
```diff
@@ -33,6 +33,8 @@
 #include <mctf.h>
 #include <utils.h>
 
+extern int pgagroal_as_server_reset_query_behavior_on_failure(char* str, int* val);
```
MohamedM216, line 36: to fix the compilation error in CI:
```
/__w/pgagroal/pgagroal/test/testcases/test_configuration.c:332:23: error: call to undeclared function 'pgagroal_as_server_reset_query_behavior_on_failure'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  332 |    MCTF_ASSERT_INT_EQ(pgagroal_as_server_reset_query_behavior_on_failure("bogus", &value), 1, cleanup,
      |        
```

test/testcases/test_configuration.c
```diff
@@ -265,6 +267,125 @@ MCTF_TEST(test_configuration_server_reset_query_default)
...
+   MCTF_ASSERT_INT_EQ(value, SERVER_RESET_QUERY_BEHAVIOR_ON_FAILURE_DISCARD, cleanup,
+                      "'discard' should map to SERVER_RESET_QUERY_BEHAVIOR_ON_FAILURE_DISCARD");
+
+   value = -1;
+   MCTF_ASSERT_INT_EQ(pgagroal_as_server_reset_query_behavior_on_failure("try", &value), 0, cleanup,
+                      "pgagroal_as_server_reset_query_behavior_on_failure('try') should return 0");
+   MCTF_ASSERT_INT_EQ(value, SERVER_RESET_QUERY_BEHAVIOR_ON_FAILURE_TRY, cleanup,
+                      "'try' should map to SERVER_RESET_QUERY_BEHAVIOR_ON_FAILURE_TRY");
+
+   value = -1;
```
MohamedM216, line 326: I also increase here the test coverage to ensure parser maps each string to the correct integer constant and configuration reload works correctly

fluca1978: looks good to me but there is an extra effort to do, or better, twice: since we are adding a new configuration setting, we need to expose this thru metrics and also place it in the configurator:
- see `prometheus.c` <https://github.com/pgagroal/pgagroal/blob/master/src/libpgagroal/prometheus.c> for the metrics
- see `config.c` <https://github.com/pgagroal/pgagroal/blob/master/src/config.c for the configurator.
At glance, the configurator will require `to_server:_reset:_query_behavior_on_failure` to become non-static.

MohamedM216: WIP...

fluca1978:
@MohamedM216 ping me when ready

MohamedM216:
@fluca1978  I've pushed a new patch. Thanks!

fluca1978:
LGTM, please rebase so we can merge.

MohamedM216:
> LGTM, please rebase so we can merge.

Done. Thanks!

fluca1978:
merged, thanks @MohamedM216 
Please keep commit message explanatory and in sync with the status of the patch. Commit messages **are** documentation (for us).

MohamedM216:
Thanks for the feedback! I'll make sure the future commit messages are more explanatory.

PR #977, opened 2026-08-10
[#692]Add standby WAL receiver status to CLI

### LINKED ISSUE #692: Add `status standbys` to `pgagroal-cli`

Zeyad-Daowd:
This PR resolves #692 by extending the pgagroal-cli status details command to report the streaming health of configured standby servers. It adds a new Standbys section to the output that details the WAL receiver status, replication slot, and the upstream primary host/port.

Implementation Details:

- Added new query for getting required info for standbys `SELECT status, slot_name, sender_host, sender_port FROM pg_stat_wal_receiver;`

- Generic Multi-Column Parser: Added pgagroal_read_query_multiple_columns_text to queries.c. This mimics the original pgagroal_read_query_first_column_text but is extended to retreive multiple column values

- Modified status_details to selectively execute the WAL receiver check only on servers identified as replicas (Primary: No). This prevents unnecessary querying.

Example Output:

image

Open Questions (Feedback Requested):

As discussed in the issue thread, there are a few follow-up items I would like your input on to see if they should be included in this PR or handled in a separate issue:

1. Split-Brain Detection: For the split-brain detection (multiple primaries), should we log an error in the FOREACH_VALID_SERVER loop in status_details, or add a specific split_brain flag to the output?

2. Primary Validation: Should we explicitly validate that the standbys are actually streaming from our recognized primary? (This might require IP resolving to match sender_host against our configured primary host/port).

closes #692

### DIFF
--- a/src/libpgagroal/server.c
+++ b/src/libpgagroal/server.c
@@ -727,6 +727,73 @@ done:
 }
 
 int
+pgagroal_server_get_wal_receiver_status(int server,
+                                        char* rep_status, size_t status_size,
+                                        char* slot_name, size_t slot_size,
+                                        char* sender_host, size_t host_size,
+                                        char* sender_port, size_t port_size)
+{
+   int fd = -1;
+   struct main_configuration* config = (struct main_configuration*)shmem;
+   char* cols[] = {rep_status, slot_name, sender_host, sender_port};
+   size_t sizes[] = {status_size, slot_size, host_size, port_size};
+
+   if (rep_status == NULL || slot_name == NULL || sender_host == NULL || sender_port == NULL)
+   {
+      return 1;
+   }
+
+   rep_status[0] = '\0';
+   slot_name[0] = '\0';
+   sender_host[0] = '\0';
+   sender_port[0] = '\0';
+
+   if (server < 0 || server >= config->number_of_servers)
+   {
+      return 1;
+   }
+
+   if (strlen(config->health_check_user) == 0)
+   {
+      pgagroal_log_debug("ping/status: health_check_user is not configured, cannot check WAL receiver status");
+      return 1;
+   }
+
+   if (pgagroal_server_query_execute(server,
+                                     config->health_check_user,
+                                     config->health_check_user,
+                                     (char*)pgagroal_queries_wal_receiver_status(),
+                                     MAX(1, (int)pgagroal_time_convert(config->health_check_timeout, FORMAT_TIME_S)),
+                                     NULL, &fd))
+   {
+      pgagroal_log_debug("Failed to execute WAL receiver query for server %d", server);
+      goto error;
+   }
+
+   if (pgagroal_read_query_multiple_columns_text(fd, 4, cols, sizes))
+   {
+      pgagroal_log_debug("Failed to read WAL receiver status columns for server %d", server);
+      goto error;
+   }
+
+   if (fd != -1)
+   {
+      (void)pgagroal_write_terminate(NULL, fd);
+      pgagroal_disconnect(fd);
+   }
+
+   return 0;
+error:
+   if (fd != -1)
+   {
+      (void)pgagroal_write_terminate(NULL, fd);
+      pgagroal_disconnect(fd);
+   }
+
+   return 1;
+}
+
+int
 pgagroal_get_primary(int* server)
 {
    int primary;
--- a/src/libpgagroal/status.c
+++ b/src/libpgagroal/status.c
@@ -157,6 +157,7 @@ status_details(bool details, struct json* response)
    int active = 0;
    int total = 0;
    struct json* servers = NULL;
+   struct json* standbys = NULL;
    struct main_configuration* config;
 
    config = (struct main_configuration*)shmem;
@@ -193,6 +194,7 @@ status_details(bool details, struct json* response)
    pgagroal_json_put(response, MANAGEMENT_ARGUMENT_NUMBER_OF_SERVERS, (uintptr_t)config->number_of_servers, ValueInt32);
 
    pgagroal_json_create(&servers);
+   pgagroal_json_create(&standbys);
 
    FOREACH_VALID_SERVER
    {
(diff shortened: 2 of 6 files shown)


src/libpgagroal/server.c
```diff
@@ -726,6 +726,64 @@ pgagroal_server_get_connectivity_info(int server, char** status, char** primary,
...
+
+   if (server < 0 || server >= config->number_of_servers)
+   {
+      return 1;
+   }
+
+   if (strlen(config->health_check_user) == 0)
+   {
+      pgagroal_log_debug("ping/status: health_check_user is not configured, cannot check WAL receiver status");
+      return 0;
```
fluca1978: should this case be an error?

src/libpgagroal/server.c
```diff
@@ -726,6 +726,64 @@ pgagroal_server_get_connectivity_info(int server, char** status, char** primary,
...
+                                     (char*)pgagroal_queries_wal_receiver_status(),
+                                     MAX(1, (int)pgagroal_time_convert(config->health_check_timeout, FORMAT_TIME_S)),
+                                     NULL, &fd))
+   {
+      goto done;
+   }
+
+   if (pgagroal_read_query_multiple_columns_text(fd, 4, cols, sizes))
+   {
+      pgagroal_log_debug("Failed to read WAL receiver status columns for server %d", server);
```
fluca1978, line 775: should this case return an error?

src/libpgagroal/status.c
```diff
@@ -230,9 +232,40 @@ status_details(bool details, struct json* response)
...
+                                                 sender_host, sizeof(sender_host),
+                                                 sender_port, sizeof(sender_port));
+
+         struct json* standby_js = NULL;
+         pgagroal_json_create(&standby_js);
+         bool has_receiver_data = (strlen(rep_status) > 0);
+         bool is_streaming = (strcmp(rep_status, "streaming") == 0);
+         pgagroal_json_put(standby_js, MANAGEMENT_ARGUMENT_STREAMING, (uintptr_t)is_streaming, ValueBool);
+
+         if (has_receiver_data)
```
fluca1978: should we set this boolean variable depending on the success of the get wal function return value?

fluca1978 (changes requested):
This is good work, but there are more check for error paths to include, as suggested.
Also we should provide the added information somehow into prometheus?

fluca1978:
With regard to the split brain detection, I believe a flag to output in `status` and `details`, and maybe also into `ping` (since this is supposed to be a very dangerous situation) is th right approach. If more than one primary is detected during the foreach loop, the flag has to be raise. However, I believe this change could be a pull request on its own, since it could implement a separated piece of logic, even if small, that has nothing to do with streaming (e.g., wrong configuration from the very beginning). Therefore, I would open a related issue and PR.

With regard to a replica streaming from a different primary, again, this is a separate issue/PR according to me.
Rather than checking from the IP properties, I would match the wal receiver and sender properties to see if a replica is not listed on the primary side.

src/libpgagroal/server.c
Zeyad-Daowd: yes, agreed. will change it to return 1.

src/libpgagroal/server.c
Zeyad-Daowd: yes, agreed. will change it to return 1 after optionally disconnecting fd.

src/libpgagroal/status.c
Zeyad-Daowd: Not sure if you meant is_streaming or has_receiver_data here 
On is_streaming: the reasoning for the match was (strcmp(rep_status, "streaming") == 0) that pg_stat_wal_receiver.status can also be starting, waiting, restarting, or stopping, not just streaming or empty
On has_receiver_data: if that's what you meant, I think you're right, It can probably be set from the function's return value instead (== 0 for success)

fluca1978: I was referring to `has_receiver_data`, so it should be true if the function returned no error, and hence you can protect all the JSON puts within a branch moving the extraction of real data, like `is_streaming` within the branch,

Zeyad-Daowd:
@fluca1978
regarding prometheus support, I am thinking of adding a second call in the periodic health check loop (health.c) that caches the WAL receiver status into a new streaming_state field on struct server, same pattern health_state already uses. this means the query runs on two schedules now health-check interval (new) plus CLI calls (existing)
If that is a reasonable approach, should that be another issue?

fluca1978:
> @fluca1978 regarding prometheus support, I am thinking of adding a second call in the periodic health check loop (health.c) that caches the WAL receiver status into a new streaming_state field on struct server, same pattern health_state already uses. this means the query runs on two schedules now health-check interval (new) plus CLI calls (existing) If that is a reasonable approach, should that be another issue?

Ok, that sounds fine, so just let the `pgagroal-cli` so far, and open a new issue/PR for updating prometheus.

Zeyad-Daowd:
Thanks, pushed the updates for the review comments above. Tested locally against a primary/replica setup, pgagroal-cli status details output looks correct for the streaming case.
Will open 3 follow-up issues for what we've discussed:
Prometheus support for WAL receiver status into struct server via the periodic health check loop, as discussed above.
Split-brain / multiple-primaries detection.
Validating a standby is actually streaming from its correct primary

src/libpgagroal/status.c
Zeyad-Daowd: Done, i changed it wal_result == 0 (the function's actual return code), and moved the is_streaming inside the branch to protect the JSON put

fluca1978:
merged, thanks @Zeyad-Daowd

Zeyad-Daowd:
Happy to help!

Issue #938, opened 2026-06-23
Introduce pgagroal-coordinator (v1)


Abdelrhmansersawy:
Introduce `pgagroal-coordinator`, a new optional daemon that fronts multiple `pgagroal` instances to provide manual primary/replica switchover.
One binary shipped in the same `.rpm` as `pgagroal`. Vanilla `pgagroal` keeps working standalone - coordinator is optional.

## Simple flow of v1

1. Configure pgagroal-coordinator with primary and replica
2. Start pgagroal on each node (primary, replica)
3. Start pgagroal-coordinator
4. Send query through coordinator -> hits primary
5. Manual Shutdown primary
6. Run: `pgagroal-cli coordinator-switch-to <replica>`
7. Coordinator triggers promotion on replica's pgagroal
8. Send query through coordinator -> hits replica (new primary)

## Config format

```ini
[pgagroal-coordinator]
host = 0.0.0.0
port = 6432
management = 6433

[node1]
host = 10.0.0.11
port = 2345
management = 2346
role = primary

[node2]
host = 10.0.0.12
port = 2345
management = 2346
role = replica
```

Notes:
1. `[nodeX].management` is mandatory - used to trigger promotion on nodeX.
2. `[nodeX].port` is the data-plane port - coordinator forwards client traffic here.

## For v2

1. Automatic failover / heartbeat
2. Load balancing across replicas
3. Star / cascading topologies — add `upstream = node` field
4. Runtime config sync across nodes

Abdelrhmansersawy:
@fluca1978 @jesperpedersen
Am I on the right direction?

Should I work on v1 as a single PR (with incremental progress),
or split it into sub-issues from this one?

jesperpedersen:
@Abdelrhmansersawy It is just `switch-to` and the Unix Domain Socket or Host/Port pair - there are no `coordinator-` commands

jesperpedersen:
@Abdelrhmansersawy One PR

jesperpedersen:
@Abdelrhmansersawy And this is a setup of 5 machines - 1 x Coordinator + 2 x pgagroal + 2 x PostgreSQL

Abdelrhmansersawy:
> [@Abdelrhmansersawy](https://github.com/Abdelrhmansersawy) And this is a setup of 5 machines - 1 x Coordinator + 2 x pgagroal + 2 x PostgreSQL

I have some doubts about running pgagroal and PostgreSQL on separate machines.
[...]

PR #983, opened 2026-08-11
[#938] Add pgagroal-coordinator

Abdelrhmansersawy:
## Changes

1. Added new optional daemon `pgagroal-coordinator` that fronts a number of `pgagroal` instances behind a single address.

2. Implemented node discovery over the management protocol, so the data plane port and the server role are pulled at runtime instead of configured.

3. Implemented data plane routing to the node that serves the primary, terminating the connections of the previous one on a switchover.

4. Added a management endpoint speaking the existing protocol, so `status` and `switch-to` are the same `pgagroal-cli` commands against a different endpoint.

5. Extended `pgagroal_remote_management_auth` into a shared implementation, so both daemons share the SCRAM path, and added the `-u` and `-A` credential files.

6. Extended the `status` response with the port and management port of the instance.

`switch-to` does not promote PostgreSQL, it only re-targets the routing, and it refuses a node it cannot reach. There is no health check and no automatic failover.


Abdelrhmansersawy:
@fluca1978 I know this is a pretty large PR  Let me know if there’s anything I can do to make the review easier, or if there’s anything I should clarify/provide.

jesperpedersen:
@fluca1978 Please, wait on me

doc/man/pgagroal-coordinator.1.rst
```diff
@@ -0,0 +1,74 @@
...
+pgagroal-coordinator [ -c CONFIG_FILE ] [ -u USERS_FILE ] [ -A ADMINS_FILE ]
+
+DESCRIPTION
+===========
+
+**pgagroal-coordinator** is an optional daemon that sits in front of a number of
+[**pgagroal**][pgagroal] instances, each of them serving one PostgreSQL server, and routes
+the client traffic to the instance that serves the primary.
+
+At startup the coordinator connects to the management port of each node and asks for its
```
fluca1978, line 23: so this means that each pgagroal has to be management configured, i think we should specify this.

Ops, it's below in the note section.

doc/man/pgagroal-coordinator.1.rst
```diff
@@ -0,0 +1,74 @@
...
+whose server reports being the primary is elected, and every client connection is proxied
+to it.
+
+The election is refreshed when a ``switch-to`` is performed. There is no health check and
+no automatic failover: the coordinator follows the decisions of the operator.
+
+**pgagroal-coordinator** is optional, and a [**pgagroal**][pgagroal] installation that does
+not use it is not affected by it.
+
+**Note:** the management port of each node must be open, the user of the coordinator must be
```
fluca1978, line 35: I would set this section as "requirements", rather than "note", since without these settings the whole coordinator stuff will not work.

doc/man/pgagroal-coordinator.conf.5.rst
```diff
@@ -0,0 +1,155 @@
...
+
+The options for the pgagroal-coordinator section are
+
+host
+  The bind address for pgagroal-coordinator. Mandatory
+
+port
+  The bind port for pgagroal-coordinator, where the clients connect. Mandatory
+
+management
```
fluca1978, line 36: is this really mandatory? I mean, could we set it to a default and apply a default to the `pgagroal` management port too?

doc/man/pgagroal-coordinator.conf.5.rst
```diff
@@ -0,0 +1,155 @@
...
+  be different from ``port``
+
+user
+  The entry of the users file that is presented to the management port of the nodes.
+  It is used by every node that does not define its own ``user``
+
+backlog
+  The backlog for listen(). Minimum ``16``. Default is 16
+
+ev_backend
```
fluca1978, line 47: from here going on we have almost the same configuration settings for `pgagroal`, would it be worth to simply redirect the user to the other part of the documnetation to keep it in sync?

doc/man/pgagroal-coordinator.conf.5.rst
```diff
@@ -0,0 +1,155 @@
...
+
+management
+  The remote management port of the pgagroal instance. Mandatory
+
+user
+  The entry of the users file that is presented to the management port of this node.
+  Overrides the ``user`` of the main section
+
+The port of the data plane and the role of a node are not configured. They are discovered
+through the management protocol when the node is probed, and they are reported by
```
fluca1978, line 112: "and they are reported by" -> "as they are reported by"?

doc/man/pgagroal-coordinator.conf.5.rst
```diff
@@ -0,0 +1,155 @@
...
+``pgagroal-cli status``.
+
+At most 8 node sections are supported.
```
fluca1978, line 115: Is there a particular reason for this or is simply a limitation you decided? Just curious.

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+
+> **Note:** All host addresses, ports, usernames, and passwords used in this chapter are examples only.
+> You are free to use any values that fit your environment.
+
+## Architecture
+
+```
+                                +--> pgagroal (node1) --> PostgreSQL (primary)
+                                |
+client --> pgagroal-coordinator -+
```
fluca1978, line 20: the `+` seems misaligned with the branches, move the branches on space to the right

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+                                |
+                                +--> pgagroal (node2) --> PostgreSQL (replica)
+```
+
+Each [**pgagroal**][pgagroal] instance serves one [PostgreSQL][postgresql] server, and the instances are normally installed on separate machines.
+
+The coordinator never connects to [PostgreSQL][postgresql] directly. It only talks to the management port of each [**pgagroal**][pgagroal] instance, and it forwards client traffic to the data plane of the instance that is currently the primary.
+
+The coordinator is a router, not a connection pool. Pooling is performed by the [**pgagroal**][pgagroal] instances, as usual.
```
fluca1978, line 29: I think this should be stressed more, so place in the beginning of the coordinator description and make it bold (or something like that).

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+
+All client traffic is routed to the primary. The instances that serve a replica are probed and reported, but they carry no client traffic until a switchover makes one of them the primary.
```
fluca1978, line 31: `until a switchover happens` ?

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+| PostgreSQL           | Replica | `10.0.0.12` | 5432 |            |
+| pgagroal             | node1   | `10.0.0.11` | 2345 | 2346       |
+| pgagroal             | node2   | `10.0.0.12` | 2345 | 2346       |
+| pgagroal-coordinator |         | `10.0.0.10` | 6432 | 6433       |
+
+Clients connect to `10.0.0.10:6432`, and `pgagroal-cli` administrates the coordinator on `10.0.0.10:6433`.
+
+## Discovery
+
+The topology is not written in the configuration file of the coordinator. Only the address and the management port of each node are configured, and everything else is discovered.
```
fluca1978, line 51: he topology will be discovered dynamically and therefore there is no need to declare it in the configuration ...` ?

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+
+When the coordinator starts, it connects to the management port of each node and asks for its status. The reply tells the coordinator
```
fluca1978, line 53: add a colon at the end of the line

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+> `primary` setting of the node configuration. Configure `health_check_user` on every node so that the elected
+> primary reflects the state of [PostgreSQL][postgresql] rather than the state of a configuration file.
+
+Discovery runs when the coordinator starts and when a switchover is performed. The coordinator does not poll the nodes, and it never changes the primary on its own.
+
+## Configuring the nodes
+
+Every [**pgagroal**][pgagroal] instance that is fronted by the coordinator needs four things.
+
+**Step 1 -- Enable the management port** in `pgagroal.conf` on each node, so that the coordinator can reach it:
```
fluca1978, line 72: I would remove "step" from numbering

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+**Step 3 -- Authorize that account** in `pgagroal_hba.conf` on each node. Remote management always uses the `admin` database, and the address is the one of the coordinator:
+
+```
+#
+# TYPE  DATABASE  USER   ADDRESS         METHOD
+#
+host    admin     admin  10.0.0.10/32    scram-sha-256
+```
+
+**Step 4 -- Restart the node** so that the new settings are applied, and verify that the management port answers:
```
fluca1978, line 99: verify is supposed to be run from the coordinator node, right? Better explain this.

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+[node1]
+host = 10.0.0.11
+management = 2346
+
+[node2]
+host = 10.0.0.12
+management = 2346
+```
+
+and press `Ctrl-d` (if running `cat`) to save the file.
```
fluca1978, line 135: I would remove the `cat` part, show the file and let the user decide.

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+
+| Action                                                | Performed by                                        |
+|-------------------------------------------------------|-----------------------------------------------------|
+| Promoting the [PostgreSQL][postgresql] replica         | The operator, or the failover tooling of the cluster |
+| Choosing the node that receives the client traffic     | The operator, through `switch-to`                    |
+| Routing the new client connections                     | The coordinator                                      |
+| Terminating the client connections of the old primary  | The coordinator                                      |
+| Stopping the old [PostgreSQL][postgresql] server       | The operator                                         |
+
+Promote first, then tell the coordinator to follow. The reason is that the coordinator routes traffic to whichever node it is told to use, without verifying that the server behind it accepts writes, so pointing it at a server that is still a replica sends the clients to a read-only server.
```
fluca1978, line 254: I've a doubt here: when we issue a `swotch-to` a couple of scripts could be run, so that we can reconfigure the nodes dynamically, could it be worth to mention this as "tooling"?
Second: promote first and then tell the coordinator, is this posing a possible split brain window?

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+
+**Step 1 -- Promote the [PostgreSQL][postgresql] server** of the node that is to become the primary, as described in the [Failover](16-failover.md) chapter. The command runs on the machine of that node:
```
fluca1978, line 256: I would show rather a `psql -c 'select pg_promote();'` so that it can be run from any node.

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+```
+psql -h 10.0.0.12 -p 5432 -U myuser -d mydb -c "SELECT pg_is_in_recovery()"
+```
+
+The query must return `f`.
+
+**Step 2 -- Point the coordinator at the new primary:**
+
+```
+pgagroal-cli -h 10.0.0.10 -p 6433 -U admin -P admin1234 switch-to node2
```
fluca1978, line 273: I don't remember exactly, could it be this is going to come later on, but I thought this was the only command to issue. The coordinator could propagate, in the proper way, the `seitch-to` to pgagroal backends where needed. We can postpone this development.

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+The coordinator performs the following, in order:
+
+1. It verifies that `node2` exists in its configuration, and probes it. If it does not exist, or cannot be
+   reached, the command fails and **nothing is changed**
+2. It records `node2` as the current primary. From this moment every **new** client connection is routed to `node2`
```
fluca1978, line 280: `is` -> `are`

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+3. It probes the remaining nodes, so that `status` reports the current role, health and port of each of them
+4. It terminates the client connections that were routed to the previous primary
```
fluca1978, line 282: shouldn't this be the first action after having verified the node is reachable?

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+
+Only the connections of the previous primary are terminated. Running `switch-to` for the node that is **already** the primary is reported as successful, and it does not terminate anything, so the command is safe to repeat:
+
+```
+INFO  pgagroal-coordinator: Node [node2] is already the primary
+```
+
+### The old primary
+
+The coordinator does not stop, promote or reconfigure the old node. It stops sending client traffic to it, and it reports it through `status`, nothing more.
```
fluca1978, line 321: `promote' does not make sense here.

And again, I think we could try to reconfigure it via the notify scripts.

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+Preventing the old [PostgreSQL][postgresql] server from accepting writes, and rebuilding it as a replica of the new primary, are part of the failover procedure of the cluster and remain the responsibility of the operator.
+
+> **Note:** If the old server is promoted, or was never demoted, two servers accept writes at the same time.
+> The coordinator reports this as `Split brain, 2 nodes report being the primary` and keeps using the node it
+> was told to use, because choosing between two primaries automatically would risk directing the clients to
+> the wrong one.
+
+### If the order is reversed
+
+If `switch-to` is run before the promotion, the command succeeds, because the coordinator routes to the node it is told to use. The clients then reach a server that is still in recovery, and every write fails:
```
fluca1978, line 332: I think this should be implemented better: `switch-to` should not allow routing to a replica node.
This implementation could be postponed.

src/libpgagroal/configuration.c
```diff
@@ -1230,6 +1231,452 @@ pgagroal_vault_validate_configuration(void* shm)
...
+      pgagroal_log_warn("pgagroal-coordinator: ev_backend configuration is empty. Default to 'auto'");
+      config->ev_backend = PGAGROAL_EVENT_BACKEND_AUTO;
+   }
+
+   if (config->ev_backend == PGAGROAL_EVENT_BACKEND_AUTO || !is_supported_backend(config->ev_backend))
+   {
+      config->ev_backend = DEFAULT_EVENT_BACKEND;
+   }
+
+#if HAVE_LINUX && HAVE_IO_URING
```
fluca1978, line 1634: I think we should refactor this piece into something like `pgagroal_check_io_uring_enabled()` and use it everywhere.
We could do this into another issue according to me.

src/libpgagroal/configuration.c
```diff
@@ -2156,7 +2603,273 @@ pgagroal_vault_read_users_configuration(void* shm, char* filename)
          // Strict UTF-8 validation after decryption
          if (!pgagroal_utf8_valid((const unsigned char*)password, strlen(password)))
          {
-            printf("pgagroal: Invalid VAULT USER entry: invalid UTF-8 password for user '%s'\n", username);
+            printf("pgagroal: Invalid VAULT USER entry: invalid UTF-8 password for user '%s'\n", username);
```
fluca1978, line 2606: why this vault related piece in here?

src/libpgagroal/configuration.c
```diff
@@ -2166,19 +2879,19 @@ pgagroal_vault_read_users_configuration(void* shm, char* filename)
          }
 
          if (strlen(username) < MAX_USERNAME_LENGTH &&
-             strlen(password) < MAX_PASSWORD_LENGTH &&
-             !strcmp(config->vault_server.user.username, username))
+             strlen(password) < MAX_PASSWORD_LENGTH)
```
fluca1978, line 2882: why vault realted stuff in here?

src/libpgagroal/coordinator.c
```diff
@@ -0,0 +1,1072 @@
...
+   config = (struct coordinator_configuration*)shmem;
+
+   username = config->users[config->nodes[node].user_index].username;
+   password = config->users[config->nodes[node].user_index].password;
+
+   if (connect_timeout(config->nodes[node].host, config->nodes[node].management, fd))
+   {
+      pgagroal_log_debug("pgagroal-coordinator: Could not connect to node [%s] %s:%d",
+                         config->nodes[node].name, config->nodes[node].host, config->nodes[node].management);
+      atomic_store(&config->nodes[node].reachable, SERVER_HEALTH_DOWN);
```
fluca1978, line 224: since we are using `SERVER_HEALTH_UP|DOWN` does it make sense to change `reachable` to `health`?

src/libpgagroal/coordinator.c
```diff
@@ -0,0 +1,1072 @@
...
+                     config->nodes[new].name);
+
+   pgagroal_coordinator_set_primary(new);
+
+   /* refresh the view of the remaining nodes */
+   pgagroal_coordinator_probe();
+
+   if (old >= 0)
+   {
+      pgagroal_coordinator_client_kill(old);
```
fluca1978, line 765: I think this should happen before we switch to a new primary. If the switch fails, then applications will reconnect to the old primary (no primary change in the coordinator), if the switch succeed we are avoiding split brain window.

src/coordinator.c
```diff
@@ -0,0 +1,994 @@
...
+         default:
+            usage();
+            exit(1);
+            break;
+      }
+   }
+
+   if (getuid() == 0)
+   {
+      errx(1, "pgagroal-coordinator: Using the root account is not allowed");
```
fluca1978, line 183: this is fine, but poses a doubt: we are testing all over the code (not only in this PR) that IP ports are withing 0 and 65536, but all our executables are non-root. This means our ports should be 1025-65536 bounadry limited, right?
@jesperpedersen is it worth adding a uniform utility "check port boundaries" function to use all over?
Clearly it does not belong to this PR.

fluca1978 (changes requested):
A lot of work, seems good to me even if I've not tested yet, requires more time.
I'm going to test the next week.

I've a few objections so far, mainly the fact that it seems to me there a split brain window since we promote a node and only **after** reconfigure the coordinator to the new primary, while old connections are still routed to the old primary (which in turn is write enabled).

@jesperpedersen let's wait for your comments and do both agree on the changes.

jesperpedersen:
We can't open port implicit - all ports have to be explicit set by users and visible in the configuration

doc/manual/en/22-coordinator.md
```diff
@@ -0,0 +1,389 @@
...
+All client traffic is routed to the primary. The instances that serve a replica are probed and reported, but they carry no client traffic until a switchover makes one of them the primary.
+
+## Prerequisites
+
+* [PostgreSQL][postgresql] streaming replication already configured and running
+* One [**pgagroal**][pgagroal] instance per [PostgreSQL][postgresql] server
+* The host/port layout used throughout this chapter:
+
+| Component            | Role    | Host        | Port | Management |
+|----------------------|---------|-------------|------|------------|
```
jesperpedersen, line 40: Use `:----`

jesperpedersen:
This is going in the right direction

jesperpedersen:
We shouldn't restrict port numbers - that is something `-i` / `--init` should hint about if < 1025

jesperpedersen:
@Sakshii-27 Where are we on that project ?

Abdelrhmansersawy:
I was traveling these past two days, so I started working on resolving the comments today and tomorrow.

doc/man/pgagroal-coordinator.conf.5.rst
Abdelrhmansersawy: Yes, Jesper suggest to be explicit instead of implicit here https://github.com/pgagroal/pgagroal/pull/983#issuecomment-5303187548

What do you think ?

doc/man/pgagroal-coordinator.conf.5.rst
Abdelrhmansersawy: No, It is just the array size. `nodes[]` is a fixed array in shared memory, the same as `servers[]` and `admins[]` in pgagroal, so 8 I choose it same like `admins[]. if you have a topology in mind that needs more, tell me the number and I will set it

doc/manual/en/22-coordinator.md
Abdelrhmansersawy: Right now it only handles one level of replicas, which is the goal of v1.

For v2 I was thinking of something per node like `stream_node = node_x`. That is basically Patroni's `replicatefrom` tag. The difference is that in Patroni the node publishes it itself, while here it would sit in the coordinator configuration, so the coordinator would hold the topology instead of discovering it. I am not sure yet which side it should live on.

What do you think?

I'm thinking there are lots of working here , especially in-case auto-failour so we need to adjust topology.

src/libpgagroal/configuration.c
Abdelrhmansersawy: it's identical I don't change on line, it seems to be a bug from code diff

fluca1978:
> We shouldn't restrict port numbers - that is something `-i` / `--init` should hint about if < 1025

I don't get this, but I'm fine with it: we are already restricting using `root`, so seems to me we cannot already run ports less than 1025, so I think checking boundaries is a natural path instead of a warning. Also this means `pgagroal-config` should use the very same logic and avoid the user to specify lower ports.
Note that I'm talking of ports on pgagroal side, so `port`, `management`, and so on.

Issue #52, opened 2020-02-26
Security module: SCRAM-SHA-256-PLUS


jesperpedersen:

MichaelDBA:
When will this be available?

jesperpedersen:
There is currently no roadmap for SCRAM-SHA-256-PLUS, which SCRAM-SHA-256 over TLS with channel binding.

Standard SCRAM-SHA-256 authentication is supported over UDS, TCP and TLS.

jesperpedersen:
https://www.postgresql.org/docs/current/sasl-authentication.html
https://datatracker.ietf.org/doc/html/rfc5802#page-14
https://www.postgresql.org/docs/devel/protocol-message-formats.html
AuthenticationSASL (B)

Userfrom1995:
> [@Userfrom1995](https://github.com/Userfrom1995) Can you take this ?
Sure, I can take this one.

PR #984, opened 2026-08-11
[#52] Add SCRAM-SHA-256-PLUS channel binding

trxvorr:
Dual-sided tls-server-end-point channel binding on the terminated path, gated by a per-server channel_binding option (disabled/prefer/require). Pass-through does not advertise it.


jesperpedersen:
@trxvorr Start cross-porting

jesperpedersen:
Merged.

PR #985, opened 2026-08-11
Fix SCRAM salted_password key reuse on OpenSSL 3.0.2

trxvorr:
EVP_MAC_init(ctx, NULL, 0, NULL) does not reuse the HMAC key after EVP_MAC_final() on OpenSSL 3.0.2, so PBKDF2 iterations 2..n use a wrong key and every SCRAM authentication fails. Re-pass the password each round.


jesperpedersen:
Rebased and merged.

PR #986, opened 2026-08-14
[#979] Add Prometheus metric for standby streaming status

### LINKED ISSUE #979: Prometheus support for standby streaming status
The status details output (see #977) now reports whether each standby is actively streaming from its primary, via a live query against pg_stat_wal_receiver. This isn't yet exposed via Prometheus.
Per discussion on #977 , the proposed approach: 
1. add a second call site in the periodic health check loop (health.c).
2. Prometheus would then read the cached value, consistent with how pgagroal_server_error and similar metrics work today.

Zeyad-Daowd:
Adding pgagroal_server_streaming gauge metric to prometheus (-1 = primary server, 0 = not streaming, 1 = streaming)

Description:
Adds the pgagroal_server_streaming gauge metric to Prometheus to monitor whether a standby server is actively streaming from its primary. The metric is evaluated during the periodic health check loop via pg_stat_wal_receiver.

config
image

results:
primary server correctly shows -1 both before initializing as it is (not_init_primary) and after initialization (connection to pgagroal)
image

metrics html
image

closes #979

### DIFF
--- a/src/libpgagroal/health.c
+++ b/src/libpgagroal/health.c
@@ -201,6 +201,48 @@ health_check_loop(void)
                atomic_store(&config->servers[i].health_state, SERVER_HEALTH_DOWN);
             }
          }
+
+         int current_state = atomic_load(&config->servers[i].state);
+         if (current_state == SERVER_PRIMARY || current_state == SERVER_NOTINIT_PRIMARY)
+         {
+            atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_PRIMARY);
+         }
+         else if (up)
+         {
+            char rep_status[64] = {0};
+            char slot_name[64] = {0};
+            char sender_host[MISC_LENGTH] = {0};
+            char sender_port[16] = {0};
+
+            int wal_result = pgagroal_server_get_wal_receiver_status(i,
+                                                                     rep_status, sizeof(rep_status),
+                                                                     slot_name, sizeof(slot_name),
+                                                                     sender_host, sizeof(sender_host),
+                                                                     sender_port, sizeof(sender_port));
+            bool is_streaming = pgagroal_compare_string(rep_status, "streaming");
+
+            if (wal_result == 0)
+            {
+               if (is_streaming)
+               {
+                  atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_YES);
+               }
+               else
+               {
+                  atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_NO);
+               }
+            }
+            else
+            {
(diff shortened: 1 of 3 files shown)


Zeyad-Daowd:
@fluca1978 @jesperpedersen I found an observation while testing this feature that i wanted to ask about,
In the shown configuration I put primary = on at the last line.
Initially when i was testing it i tested it as follows
image
I found that the feature wasn't working correctly and after investigation i found that when the primary = on is put first in the configuration the primary server state is NOTINIT not NOTINIT_PRIMARY
after investigating i think the reason is both "host" & "port" keys set the state to NOTINIT in parsing (configuration.c)
image


image

Is this expected?
If it is not I think we can replace the atomic store by a load first then store NOTINIT if it is not NOTINIT_PRIMARY
this won't be atomic but i think it will be fine since only configuration updates it initially?
Another thought: Since the state is already initialized to SERVER_NOTINIT at the very beginning of parsing section, what is the reason to re-apply it when parsing host or port?
image

fluca1978:
good catch, even if I don't think is going to hurt in any way.

I think we should set (atomically) `SERVER_NOTINIT` or `SERVER_NOTINIT_PRIMARY` only when parsing an host is compleeted, that is `host`, `port` and `primary` have been evaluated.
This poses also the problem that `primary` is not mandatory for standbys, so the solution I propose is that the `host` or `port` part be refined with something like:

```
atomic_store(&srv->state,  srv->primary ? SERVER_NOTINIT_PRIMARY :  SERVER_NOTINIT);
```

so this will keep consistent with `primary` wherever it is found in the configuration.

Now, since the configuration could be reloaded while pgagroal is running, we need to re-evaluate all the servers and set them to NOT_INIT (that will be quickly turned to a different status).

Thoughts?

src/libpgagroal/health.c
```diff
@@ -201,6 +201,48 @@ health_check_loop(void)
...
+                  atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_YES);
+               }
+               else
+               {
+                  atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_NO);
+               }
+            }
+            else
+            {
+               pgagroal_log_debug("server %d failed to receive wal_result, setting streaming state to NO", i);
```
fluca1978, line 237: should this be a warn message?

fluca1978 (approved):
Fine for me.

fluca1978:
merged, thanks @Zeyad-Daowd 

For the other question, let's open a new issue and restart discussion with my last comment.

Zeyad-Daowd:
Happy to help!
I'll open a new issue with the details and your proposed fix so we can continue the discussion there.

PR #988, opened 2026-08-16
Harden file handling

trxvorr:
add pgagroal_fopen_secure (O_NOFOLLOW on create/write/append, O_CLOEXEC always, 0600 perms via fchmod, O_EXCL via 'x' in mode) and route the credential/config file opens through it; use mkstemp for the pgagroal-admin user temp files instead of a predictable .tmp name.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -637,8 +637,7 @@ add_user(char* users_path, char* username, char* password, bool generate_pwd, in
       do_free = false;
    }
 
-   users_file = fopen(users_path, "a+");
-   if (users_file == NULL)
+   if (pgagroal_fopen_secure(users_path, "a+", &users_file))
    {
       warnx("Could not append to users file <%s>", users_path);
       goto error;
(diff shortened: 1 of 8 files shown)


jesperpedersen:
@trxvorr Make sure that this is rebased - orangu is going crazy on this pr

src/libpgagroal/utils.c
```diff
@@ -582,6 +583,142 @@ pgagroal_exists(const char* f)
...
+   if (strchr(mode, 'r'))
+   {
+      read = true;
+   }
+   if (strchr(mode, 'w'))
+   {
+      write = true;
+      create = true;
+   }
+   if (strchr(mode, 'a'))
```
fluca1978, line 623: since all these branches are `if` without `elses`, does it mean I can specify  `rwa` as mode?

src/admin.c
```diff
@@ -957,18 +956,24 @@ update_user(char* users_path, char* username, char* password, bool generate_pwd,
...
+   if (pgagroal_fopen_secure(users_path, "r", &users_file))
    {
       warnx("File <%s> not found", users_path);
       goto error;
    }
 
-   pgagroal_snprintf(tmpfilename, sizeof(tmpfilename), "%s.tmp", users_path);
-   users_file_tmp = fopen(tmpfilename, "w+");
+   pgagroal_snprintf(tmpfilename, sizeof(tmpfilename), "%s.XXXXXX", users_path);
+   int tmp_fd = mkstemp(tmpfilename);
```
fluca1978, line 966: does it make sense to wrap this into `pgagroal_mktemp` or alike function to create a temporary file?

fluca1978 (commented):
I like the idea, even if I don't understand why use `_secure` in the name of the function: since we are using it *everywhere*, does `pgagroal_fopen` suffice?
Moreover, at glance, it seems to me we can mix inconsistently mode chars.

PR #995, opened 2026-08-20
macOS: fix build, restore CI runtime verification, drop stale deps

Youssef-joe:
## Why

pgagroal already had real macOS support at the code level — a full `Darwin` branch in CMake, a complete `kqueue` event-loop backend used automatically on non-Linux platforms, and platform shims for `secure_getenv`/`setproctitle`. But the `build-macos` CI job only ran `cmake && make`. It never started `pgagroal` or checked that it actually works, so a macOS-specific runtime regression (e.g. in the kqueue backend) could ship without anyone noticing. This job used to install PostgreSQL and run `pgagroal-cli ping` as a smoke test (see cb47fbe, d9f58e2), but that step was dropped over time.

On top of that, docs/CI listed a dependency (`libev`) that isn't actually used anymore, and one dead macro reference was left over from an old code path.

While preparing the CI change and dry-running it locally on macOS (Darwin arm64), I found that `master` currently fails to build there at all: a static `_Atomic` array initializer is rejected by the installed AppleClang. This PR fixes that too, since otherwise the new CI smoke test can't even reach the point of running.

## What changed

- **`src/libpgagroal/ev.c`**: `signal_watchers`/`signal_callbacks` were declared as `static _Atomic(...)[...] = {0}`. Some Clang versions reject `= {0}` as a non-compile-time-constant initializer for an atomic array, which breaks the macOS build entirely. Fixed by dropping the initializer and zeroing both arrays at load time via a constructor function instead.
- **`src/libpgagroal/utils.c`**: `pgagroal_resolve_path()` checked `defined(HAVE_DARWIN) || defined(HAVE_OSX)`, but `HAVE_DARWIN` is never `#define`d anywhere in the CMake files — only `HAVE_OSX` is set for Darwin. Dropped the dead reference.
- **`.github/workflows/ci.yml`** (`build-macos` job):
  - After the existing build step, install and start PostgreSQL 17 via Homebrew, waiting for it to accept connections.
  - Start `pgagroal` against the sample `doc/etc` config and run `pgagroal-cli ping`, failing the job if it doesn't succeed — this is the actual regression fix, restoring real runtime verification instead of build-only.
  - Drop `libev` from the Homebrew install list — pgagroal has its own event loop (`ev.c`) and never links or looks for libev; confirmed by a clean local build with libev absent from CMake's dependency probing.
- **`doc/manual/en/79-distributions.md`**: drop the same stale `libev` mention from the macOS install instructions.
- **`README.md`**: add macOS to "Tested platforms", now that it's build+run verified rather than build-only.

## Validation

All of the above was validated locally end-to-end on macOS (Darwin arm64), not just written and pushed:
- Clean `cmake && make` from scratch succeeds (it did not before the `ev.c` fix).
- PostgreSQL 17 installs and starts via Homebrew, and accepts connections.
- `pgagroal` starts against the sample config, `pgagroal-cli ping` returns success, and `pgagroal-cli shutdown` cleanly stops it with no lingering processes.

## Out of scope (follow-ups, not included here)

- Making `test/check.sh` portable enough to run the full test suite natively on macOS (it currently relies on GNU-only `sed -i`, Linux-only `ipcs`/`ipcrm`, and `nproc`).
- Ubuntu support — the Linux CMake branch is already distro-agnostic, so this should mainly need a new CI job/apt path rather than code changes, but it's a separate piece of work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ See [Architecture](./doc/ARCHITECTURE.md) for the architecture of [**pgagroal**]
 
 * [FreeBSD](https://www.freebsd.org/)
 * [OpenBSD](http://www.openbsd.org/)
+* [macOS](https://www.apple.com/macos/)
 
 
 ## Compiling from sources
--- a/src/libpgagroal/ev.c
+++ b/src/libpgagroal/ev.c
@@ -141,8 +141,8 @@ static void init_watcher_message(struct io_watcher* watcher);
 /* context globals */
 
 static struct event_loop* loop = NULL;
-static _Atomic(struct signal_watcher*) signal_watchers[PGAGROAL_NSIG] = {0};
-static _Atomic(signal_cb) signal_callbacks[PGAGROAL_NSIG] = {0};
+static _Atomic(struct signal_watcher*) signal_watchers[PGAGROAL_NSIG];
+static _Atomic(signal_cb) signal_callbacks[PGAGROAL_NSIG];
 
 #if HAVE_LINUX
 
@@ -2032,5 +2032,15 @@ init_watcher_message(struct io_watcher* watcher)
       }
       watcher->msg->length = 0;
       watcher->msg->kind = 0;
+   }
+}
+
+static void __attribute__((constructor))
+pgagroal_ev_init(void)
+{
+   for (int i = 0; i < PGAGROAL_NSIG; i++)
+   {
+      atomic_store_explicit(&signal_watchers[i], NULL, memory_order_relaxed);
+      atomic_store_explicit(&signal_callbacks[i], NULL, memory_order_relaxed);
    }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,7 +198,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew update
-          brew install openssl libev zstd lz4 libssh bzip2 docutils check graphviz doxygen
+          brew install openssl zstd lz4 libssh bzip2 docutils check graphviz doxygen
 
       - name: Configure and build with clang
         run: |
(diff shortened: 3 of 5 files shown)


src/libpgagroal/ev.c
```diff
@@ -2052,3 +2034,13 @@ init_watcher_message(struct io_watcher* watcher)
...
+
+static void __attribute__((constructor))
+pgagroal_ev_init(void)
+{
+   for (int i = 0; i < PGAGROAL_NSIG; i++)
+   {
+      atomic_store_explicit(&signal_watchers[i], NULL, memory_order_relaxed);
+      atomic_store_explicit(&signal_callbacks[i], NULL, memory_order_relaxed);
+   }
+}
```
Userfrom1995, line 2046: Are you sure we need this constructor? 
Are you sure signal_watchers/signal_callbacks aren't already zero-initialized by the C runtime by default?

README.md
```diff
@@ -53,6 +53,7 @@ See [Architecture](./doc/ARCHITECTURE.md) for the architecture of [**pgagroal**]
 
 * [FreeBSD](https://www.freebsd.org/)
 * [OpenBSD](http://www.openbsd.org/)
+* [macOS](https://www.apple.com/macos/)
```
Userfrom1995, line 57: I don't think we should add macOS to Tested platforms. 
A ping smoke test doesn't qualify as tested, and I've never actually run pgagroal on macOS myself.

The same caveat arguably applies to FreeBSD/OpenBSD here too their CI only builds, no suite run either but I've tested FreeBSD manually many times, so that one is genuinely exercised. macOS has no such track record yet, so let's hold off until the real test suite runs there (see #958).

Userfrom1995:
@Youssef-joe, which Clang version on macOS does it fail to build on? CI was green on macOS, so it looks like it might be failing on an older version.

Also, it would be great if we could eventually get the test script working on macOS. It would also help with resolving https://github.com/pgagroal/pgagroal/issues/958. But obviously, not something we need to address here, we can look at that in a different PR.

jesperpedersen:
@Userfrom1995 Mac is community effort. Core Committers have critical tasks to work on

fluca1978:
I think this PR is apparently sane, at least the CI part, but I've doubts about the initializers, so I would like to clarify the clang version too and see if we can reproduce on other platforms.

Youssef-joe:
@jesperpedersen @fluca1978 @Userfrom1995 
Sorry for being late I got stuck for some issues in my company, but here I am again still very interested in contributing with that lovely community and passionate to learn from you all guys..

first of all, thanks for the reviews and for pointing these out.

I removed the constructor. `signal_watchers` and `signal_callbacks` have static storage duration, so they are guaranteed to be zero-initialized before any code runs under C11/C17 §6.7.9p10, including `_Atomic` objects. The constructor was therefore redundant and would also have introduced an unnecessary compiler-specific extension.

I also confirmed the local compiler involved in the original failure:

--> Apple clang 15.0.0 (`clang-1500.3.9.4`)
--> Xcode 15.0 Command Line Tools

I was able to isolate the issue to this specific form: AppleClang 15 rejects `= {0}` for a static aggregate of `_Atomic(T*)`, reporting that `0` is not a compile-time constant. The equivalent non-atomic declaration, and an explicit `(T *)0` initializer, compile successfully on the same compiler. A newer Clang (22.1.2) also accepts the original form, so this appears to be an AppleClang issue that has since been fixed upstream.

This also explains why the current macOS CI is green: the runner is using a newer toolchain than the one where I reproduced the failure. I haven't assumed a specific CI compiler version, but I can add `clang --version` to the macOS job if useful.

The current fix now simply uses the standard static zero-initialization:

`static _Atomic(...) x[N];`

I also reverted the README change adding macOS to Tested platforms. I agree that a build + ping smoke test isn't equivalent to running the full test suite. I'll leave the broader macOS test-suite support for the separate #958 work.

Thanks again.. I think this leaves the PR with a simpler and more portable fix.

Youssef-joe:
Another thing I wanted to flag about adding MacOS to README, now I have removed it.. but when gonna be the sign that we can add it ?
if it requires a specific type of test suits, so I am willing to learn and add them and do all needed verifications 
I can take the ownership of supporting MacOS here and anywhere else as well with knowing the needed steps to do so and learn it

Userfrom1995:
@Youssef-joe Please squash the commits into a single commit and rebase. Then we can merge.

> Another thing I wanted to flag about adding MacOS to README, now I have removed it.. but when gonna be the sign that we can add it ? if it requires a specific type of test suits, so I am willing to learn and add them and do all needed verifications I can take the ownership of supporting MacOS here and anywhere else as well with knowing the needed steps to do so and learn it

I think once we have `check.sh` running successfully for macOS builds in CI, or #958 is resolved in that PR, we can add macOS to the README.

Youssef-joe:
> Please squash the commits into a single commit and rebase. Then we can merge.

Squashed and rebased..
okayy, will consider that and review #958

Userfrom1995:
@Youssef-joe Thank you for your contribution!

PR #996, opened 2026-08-20
[#973] Fix reading split PostgreSQL message headers

### LINKED ISSUE #973: Transaction pipeline treats a TCP chunk's first byte as `ReadyForQuery`
**PgAgroal version**

`2.2.0` (commit `9f06693`)

**Description**

With `pipeline = transaction`, a query returning a large text result whose payload contains long runs of repeated `Z` bytes returns normally through direct PostgreSQL but stalls or loses protocol synchronization through PgAgroal. Replacing the repeated `Z` payload with another character returns normally through both endpoints.

`read_message()` stores the first byte of each TCP read in `m->kind`, although the read may begin inside a PostgreSQL message [`message.c`](https://github.com/pgagroal/pgagroal/blob/9f0669331cf79ee251ede5600430515048291c94/src/libpgagroal/message.c#L1380-L1385):

```c
m->kind = (signed char)(*((char*)m->data));
m->length = numbytes;
```

`transaction_server()` parses the protocol byte at its current message boundary into local `kind` and updates `in_tx` when `kind == 'Z'` [`pipeline_transaction.c#L420-L435`](https://github.com/pgagroal/pgagroal/blob/9f0669331cf79ee251ede5600430515048291c94/src/libpgagroal/pipeline_transaction.c#L420-L435), but later tests the chunk-level `msg->kind` to stop the watcher and return the backend [`pipeline_transaction.c#L472-L512`](https://github.com/pgagroal/pgagroal/blob/9f0669331cf79ee251ede5600430515048291c94/src/libpgagroal/pipeline_transaction.c#L472-L512):

```c
if (msg->kind == 'Z' && !in_tx && slot != -1)
{
   if (io_watcher_active)
   {
      pgagroal_io_stop(&server_io.io);
      io_watcher_active = false;
   }

   if (!fatal)
   {
      /* <snip> */
      if (pgagroal_return_connection(slot, wi->server_ssl, true))
      {
         goto return_error;
      }

      slot = -1;
   }
   else
   {
      exit_code = WORKER_SERVER_FATAL;
      pgagroal_event_loop_break();
   }
}
```

When a TCP read starts with `Z` while `in_tx` is false, this check can stop the backend watcher and return the connection even though that byte is not a message-boundary `ReadyForQuery`. The response then stalls or becomes misaligned.

**Reproduce:**

1. Start PostgreSQL 16 and PgAgroal with `pipeline = transaction`.
2. Run the following with the direct and PgAgroal DSNs supplied by the environment:

```python
import psycopg


SQL = "SELECT i, repeat('Z', 2048) FROM generate_series(1, 256) AS g(i)"


def run(label: str, dsn: str) -> None:
    print(f"{label}: running", flush=True)
    try:
        with psycopg.connect(dsn, autocommit=True) as conn:
            with conn.cursor() as cur:
                rows = list(cur.stream(SQL, size=1))
        print(f"{label}: {len(rows)} rows", flush=True)
    except Exception as exc:
        print(f"{label}: {type(exc).__name__}: {exc}", flush=True)


run("direct", DIRECT_DSN)
run("pgagroal", PGAGROAL_DSN)
```

3. Replace `Z` with `A` and run it again.

**Expected behavior:** Both queries should return 256 rows through PgAgroal.

**Actual behavior:** Direct PostgreSQL returns 256 rows for both values. PgAgroal returns 256 rows for `A`, but with `Z` it either remains blocked after `pgagroal: running` or prints `DatabaseError: lost synchronization with server: got message type "Z", length 1515870810`.

The reported length `1515870810` is `0x5a5a5a5a`, four `Z` bytes parsed as a protocol length.

fluca1978:
Introduces a new structure `postgresql_message_state` that is shared across several continuos invocations of `pgagroal_parse_message` to inspect the header and the content arrived so far.

`pgagroal_parse_message` gets the bytes of an incoming message and look to see if the header is complete or not, and unless the header is complete and a first payload byte has arrived it does nothing but keeping the status of incoming data.
Once the first byte of payload arrives, a callback function is invoked to parse the header and possibly the first byte.

Close #973

### DIFF
--- a/src/libpgagroal/pipeline_session.c
+++ b/src/libpgagroal/pipeline_session.c
@@ -60,8 +60,8 @@ static void session_destroy(void*, size_t);
 static void session_periodic(void);
 
 static bool in_tx;
-static int next_client_message;
-static int next_server_message;
+static struct postgresql_message_state client_message_state;
+static struct postgresql_message_state server_message_state;
 static bool saw_x = false;
 
 #define CLIENT_INIT   0
@@ -140,8 +140,8 @@ session_start(struct event_loop* loop __attribute__((unused)), struct worker_io*
    config = (struct main_configuration*)shmem;
 
    in_tx = false;
-   next_client_message = 0;
-   next_server_message = 0;
+   memset(&client_message_state, 0, sizeof(client_message_state));
+   memset(&server_message_state, 0, sizeof(server_message_state));
 
    for (int i = 0; i < config->max_connections; i++)
    {
@@ -283,6 +283,40 @@ session_periodic(void)
 }
 
 static void
+session_client_message(char kind, char* msg, int msglen __attribute__((unused)), void* arg)
+{
+   struct worker_io* wi = (struct worker_io*)arg;
+
+   /* The Q and E message tell us the execute of the simple query and the prepared statement */
+   if (kind == 'Q' || kind == 'E')
+   {
+      pgagroal_prometheus_query_count_add();
+      if (wi != NULL)
+         pgagroal_prometheus_query_count_specified_add(wi->slot);
+   }
+}
+
+static void
+session_server_message(char kind, char* msg, int msglen __attribute__((unused)), void* arg __attribute__((unused)))
+{
+   /*
+   * 'Z' needs the first payload byte to get the
+   * transaction status
+   */
+   if (kind == 'Z')
+   {
+      char tx_state = pgagroal_read_byte(msg + 5);
+
+      if (tx_state != 'I' && !in_tx)
+      {
+         pgagroal_prometheus_tx_count_add();
+      }
+
+      in_tx = tx_state != 'I';
+   }
+}
+
+static void
 session_client(struct io_watcher* watcher)
 {
    int status = MESSAGE_STATUS_ERROR;
@@ -303,40 +337,11 @@ session_client(struct io_watcher* watcher)
 
       if (likely(msg->kind != 'X'))
       {
-         int offset = 0;
-
-         while (offset < msg->length)
-         {
-            if (next_client_message == 0)
-            {
-               char kind = pgagroal_read_byte(msg->data + offset);
-               int length = pgagroal_read_int32(msg->data + offset + 1);
-
-               /* The Q and E message tell us the execute of the simple query and the prepared statement */
-               if (kind == 'Q' || kind == 'E')
-               {
-                  pgagroal_prometheus_query_count_add();
-                  pgagroal_prometheus_query_count_specified_add(wi->slot);
-               }
-
-               /* Calculate the offset to the next message */
-               if (offset + length + 1 <= msg->length)
-               {
-                  next_client_message = 0;
-                  offset += length + 1;
-               }
-               else
-               {
-                  next_client_message = length + 1 - (msg->length - offset);
-                  offset = msg->length;
-               }
-            }
-            else
-            {
-               offset = MIN(next_client_message, msg->length);
-               next_client_message -= offset;
-            }
-         }
+         pgagroal_parse_message(&client_message_state,
+                                msg->data,
+                                msg->length,
+                                session_client_message,
+                                wi);
 
          status = pgagroal_send_message(watcher, msg);
 
@@ -452,46 +457,11 @@ session_server(struct io_watcher* watcher)
    {
       pgagroal_prometheus_network_received_add(msg->length);
 
-      int offset = 0;
-
-      while (offset < msg->length)
-      {
-         if (next_server_message == 0)
-         {
-            char kind = pgagroal_read_byte(msg->data + offset);
-            int length = pgagroal_read_int32(msg->data + offset + 1);
-
-            /* The Z message tell us the transaction state */
-            if (kind == 'Z')
-            {
-               char tx_state = pgagroal_read_byte(msg->data + offset + 5);
-
-               if (tx_state != 'I' && !in_tx)
-               {
-                  pgagroal_prometheus_tx_count_add();
-               }
-
-               in_tx = tx_state != 'I';
-            }
-
-            /* Calculate the offset to the next message */
-            if (offset + length + 1 <= msg->length)
-            {
-               next_server_message = 0;
-               offset += length + 1;
-            }
-            else
-            {
-               next_server_message = length + 1 - (msg->length - offset);
-               offset = msg->length;
-            }
-         }
-         else
-         {
-            offset = MIN(next_server_message, msg->length);
-            next_server_message -= offset;
-         }
-      }
+      pgagroal_parse_message(&server_message_state,
+                             msg->data,
+                             msg->length,
+                             session_server_message,
+                             NULL);
 
       status = pgagroal_send_message(watcher, msg);
 
--- a/src/libpgagroal/pipeline_transaction.c
+++ b/src/libpgagroal/pipeline_transaction.c
@@ -68,8 +68,8 @@ static char username[MAX_USERNAME_LENGTH];
 static char database[MAX_DATABASE_LENGTH];
 static char appname[MAX_APPLICATION_NAME];
 static bool in_tx;
-static int next_client_message;
-static int next_server_message;
+static struct postgresql_message_state client_message_state;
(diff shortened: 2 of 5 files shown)


fluca1978:
My attempt at fixing this.
Please note that there is something that makes clang (even version 21) unhappy with comments.
Besides this, I'm not happy with the naming of `postgresql_message_state`, anything more appropriate?
Also seems there is some common shared between `transaction_server_message` and `session_server_message`, is it worth merging the two?

@MohamedM216 can you have a look?
@jesperpedersen thoughts?

MohamedM216:
Regarding the clang problem, I use 21 as well and it's been working fine. I use this version
```
Ubuntu clang-format version 21.1.8 (++20251221032842+2078da43e25a-1~exp1~20251221153008.77)
```
May be it's a problem in the configuration. I attached my configuration, I hope it can help
[my-format.txt](https://github.com/user-attachments/files/31274985/my-format.txt)

test/testcases/test_message.c
```diff
@@ -0,0 +1,317 @@
...
+
+  length = build_server_result(buffer);
+  memset(&state, 0, sizeof(state));
+  memset(&result, 0, sizeof(result));
+
+  pgagroal_parse_message(&state, buffer, length - 1, walk_cb, &result);
+  pgagroal_parse_message(&state, buffer + length - 1, 1, walk_cb, &result);
+
+  MCTF_ASSERT(result_ok(3, "DCZ", 1, &result, 'I', true), cleanup,
+              a               "a Z split at 5/1 must report its real state byte");
```
MohamedM216, line 313: The stray `a`, it's the cause of clang warning that CI complains about.

MohamedM216:
> Also seems there is some common shared between transaction_server_message and session_server_message, is it worth merging the two?

They're identical. I think it would be better to have a single shared callback or an inline helper, maybe `server_rfq_message` or `pgagroal_parse_message_server_rfq` or whatever you prefer,  since the two functions will inevitably drift again otherwise.

MohamedM216:
> Besides this, I'm not happy with the naming of `postgresql_message_state`, anything more appropriate?

what about `postgresql_message_parse_state`, `message_parse_state`, `pgagroal_message_parse_state` or maybe shorter like `pg_msg_parse_state`?

src/libpgagroal/pipeline_transaction.c
```diff
@@ -482,10 +455,10 @@ transaction_server(struct io_watcher* watcher)
...
          {
             /* In transaction pooling the reset query runs only when
-             * server_reset_query_always is enabled (PgBouncer parity). A
-             * non-empty reset (e.g. DISCARD ALL) subsumes DEALLOCATE ALL. If it
-             * fails the backend is discarded rather than returned dirty. */
-            if (config->server_reset_query_always && config->server_reset_query[0] != '\0' && !config->connections[slot].reset_query_failed)
+               * server_reset_query_always is enabled (PgBouncer parity). A
+               * non-empty reset (e.g. DISCARD ALL) subsumes DEALLOCATE ALL. If it
+               * fails the backend is discarded rather than returned dirty. */
+            if (config->server_reset_query_always && config->server_reset_query[0] != '\0')
```
MohamedM216, line 461: IMHO, it's unrelated change, a separate PR would be better?

You're right about removing it. Once a reset query fails on a slot, it skips trying it again on future returns, which is wrong in transaction pipeline. If the reset query failed on a previous client's use of that slot, the next client gets back a dirty connection with no reset attempted and no indication anything is wrong. The guard was suppressing the reset permanently per-slot across clients, not just within one session. Thanks for noting that!

fluca1978:
> > Besides this, I'm not happy with the naming of `postgresql_message_state`, anything more appropriate?
> 
> what about `postgresql_message_parse_state`, `message_parse_state`, `pgagroal_message_parse_state` or maybe shorter like `pg_msg_parse_state`?

I like the idea of uniforming the naming with `_message_state` everywhere, but the `pg` prefix could be confugsed with `pgagroal` at fingertips, `pgsql` maybe shorter than `postgresql`.

src/libpgagroal/pipeline_transaction.c
fluca1978: Looks like a wrong rebase, sorry.

jesperpedersen:
We are not doing `pg_` - its `pgagroal_` for functions in .h, and `static` function can have their own name that is fitting

jesperpedersen:
And, this is a cleanroom implementation of functionality, so we don't `postgresql` in the function names, we need a name that describe what it does

src/libpgagroal/pipeline_session.c
```diff
@@ -282,6 +282,40 @@ session_periodic(void)
    exit(0);
 }
 
+static void
+session_client_message(char kind, char* msg, int msglen __attribute__((unused)), void* arg)
```
MohamedM216, line 286: It'll raise `error: unused parameter ‘msg’ [-Werror=unused-parameter]` when compiled.

MohamedM216:
I tested the changes locally with python file mentioned in https://github.com/pgagroal/pgagroal/issues/973
and updated pgagroal.conf to:
```
[pgagroal]
host = localhost
port = 2345

log_type = console
log_level = debug5
log_path = 

max_connections = 8
idle_timeout = 0
blocking_timeout = 0
validation = off
unix_socket_dir = /tmp/
ev_backend = auto
pipeline = transaction
allow_unknown_users = false

[primary]
host = localhost
port = 5432

```
The error is still there when passing `Z` in the script, it hangs then with multiple `ctrl+c`, it stops and when sending a connection with `A` or `E`, even multiple times, it handles it correctly, but it throws error: `pgagroal: DatabaseError: lost synchronization with server: got message type "Z", length 1515870810` and I get this in the logs of pgagroal server:
```
2026-08-21 11:14:00 WARN  pipeline_transaction.c:526 [S] Client error (slot 0 database postgres user mohammad): Success (socket 8 status 2)
```

src/libpgagroal/pipeline_transaction.c
MohamedM216, line 447: The bug is that `transaction_server()` still uses `msg->kind == 'Z'` and `msg->kind == 'E'` for the connection-return and fatal decisions. `msg->kind` is the first byte of the raw TCP chunk, not a message boundary, so the original bug still triggers.

The fix is to move both decisions into `transaction_server_message()` itself, which already fires at correct boundaries. Add a static `saw_rfq` flag, set it in the callback when a genuine idle `Z` is seen, **reset** it before `pgagroal_parse_message` call inside `transaction_server()` and at the beginning of `transaction_start()`, and use it instead of `msg->kind == 'Z'`
```
if (saw_rfq && !in_tx && slot != -1)
```
Also move the `'E'/FATAL` check into the callback the same way; `msg + 6` is safe there because the callback only fires once the first payload byte has arrived.

I tested this locally and it fixes the hang and error `pgagroal: DatabaseError: lost synchronization with server: got message type "Z", length 1515870810` with `repeat('Z', 2048)`.

fluca1978: @MohamedM216 would you jump in?

MohamedM216: @fluca1978 Happy to jump in! Do you want to add me as a collaborator on your fork so I can push directly, or would you prefer I paste the exact diff here and you apply it?

fluca1978: > @fluca1978 Happy to jump in! Do you want to add me as a collaborator on your fork so I can push directly, or would you prefer I paste the exact diff here and you apply it?

OPen your own patch, then we can test and merge. Thanks for jumping in, I'm not very much available to code right now.

MohamedM216: OK, I'll raise a PR shortly. But could you merge this PR first https://github.com/pgagroal/pgagroal/pull/1010, so I can raise the wanted PR?

MohamedM216: > OK, I'll raise a PR shortly.

@fluca1978 when you have time, please have a look at https://github.com/pgagroal/pgagroal/pull/1011. thanks!

PR #999, opened 2026-08-21
Check star_fds allocations in network.c

Pranav-error:
Cross-port of pgmoneta/pgmoneta#1246 (issue pgmoneta/pgmoneta#1245), as requested by @jesperpedersen.

`network.c` contains the same code in all four projects. `star_fds` is overwritten with the result of `malloc` and `realloc` and then passed straight to `memcpy` with no check. `realloc` returns NULL *without* freeing the original, so on failure the original buffer is leaked and then written through a NULL pointer. The `malloc` branch one block above has the same problem.

Allocate into a temporary, check it, and on failure free `new_fds` and `continue` — which is what the loop already does when `bind_host` fails.

## Verification

- `cppcheck --enable=warning` reported `memleakOnRealloc` on this line before the change and reports nothing after.
- `clang-format` reports no changes for the file.
- I have **not** built this project locally, so the change is verified by static analysis and by being character-identical to the pgmoneta fix rather than by a compile here. CI will be the real check.

Added myself to `AUTHORS` and `doc/manual/en/97-acknowledgement.md` per DEVELOPERS.md, since this is my first PR here.

These are OOM-only paths, so practical severity is low.

### DIFF
--- a/src/libpgagroal/network.c
+++ b/src/libpgagroal/network.c
@@ -87,6 +87,7 @@ pgagroal_bind(const char* hostname, int port, int** fds, int* length, bool no_de
              (ifa->ifa_flags & IFF_UP))
          {
             int* new_fds = NULL;
+            int* tmp_fds = NULL;
             int new_length = 0;
 
             memset(addr, 0, sizeof(addr));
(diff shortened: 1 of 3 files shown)


src/libpgagroal/network.c
```diff
@@ -110,13 +111,25 @@ pgagroal_bind(const char* hostname, int port, int** fds, int* length, bool no_de
 
             if (star_fds == NULL)
             {
-               star_fds = malloc(new_length * sizeof(int));
+               tmp_fds = malloc(new_length * sizeof(int));
```
fluca1978: here `star_fds` is already `NULL`, so why using `tmp_fds`? Cannot we just use `star_fds = malloc` and then check `if ( star_fds == NULL )` again?

fluca1978 (approved):
Since it has already been ported, this is fine for me to merge, but I've a question on the first `if` branch, is my comment correct?

Pranav-error:
You're right, thanks — in the malloc branch `star_fds` is still NULL at that point, so there's nothing to lose on failure. Simplified to assign directly and check `star_fds == NULL`, and kept the temporary only in the realloc branch, where `star_fds` still holds the only pointer to the original buffer. Pushed.

fluca1978:
merged, thanks @Pranav-error

PR #1001, opened 2026-08-21
[#1000] Clear freed pointers in add_user and update_user

### LINKED ISSUE #1000: Double free in pgagroal-admin add_user and update_user on the error path
Cross-porting from pgmoneta, where @jesperpedersen asked me to carry fixes across. This one is the same bug as pgmoneta/pgmoneta#1247.

Found with the clang static analyzer (`clang --analyze` over a compile database). pgagroal's CI runs a build and tests but no static analysis, so this class has not been looked for here.

## add_user (`src/admin.c`)

The success path frees six pointers without clearing them:

```c
   fputs(entry, users_file);

   free(entry);
   free(master_key);
   free(encrypted);
   free(encoded);
   if (do_free)
   {
      free(password);
   }
   free(verify);

   fclose(users_file);
   users_file = NULL;

   end_t = time(NULL);

   if (pgagroal_management_create_outcome_success(j, start_t, end_t, &outcome))
   {
      goto error;
   }

   if (create_response(users_path, j, &response))
   {
      goto error;
   }
```

The `error:` label then frees the same six again. So a failure in either `pgagroal_management_create_outcome_success` or `create_response` — both of which run after the frees — double frees each of them.

## update_user (`src/admin.c`)

Same shape: `master_key`, `encrypted`, `encoded`, `password` and `verify` are freed before `fclose(users_file)`, two `goto error` statements follow, and the `error:` label frees them again.

## Scope

The analyzer also reports `Attempt to free released memory` at `src/admin.c:573` and `src/libpgagroal/management.c:1331`. I have **not** read through those two, so I am not claiming them — the analyzer is noisy on `goto error` patterns and I would rather verify than assert.

For what it is worth, I ran the same analysis across the family: **pgexporter and pgvictoria do not have this bug** (no `free released` findings in either), so this cross-port is pgagroal only.

Double free is undefined behaviour and corrupts the allocator; it needs the error path to be taken, so it is not hit in normal operation.

PR to follow.

Pranav-error:
Fixes #1000. Cross-port of pgmoneta/pgmoneta#1248, as requested by @jesperpedersen on pgmoneta/pgmoneta#1242.

`add_user` and `update_user` free `entry`, `master_key`, `encrypted`, `encoded`, `password` and `verify` on the success path, then run `pgagroal_management_create_outcome_success` and `create_response`, both of which `goto error`. The `error:` label frees the same pointers again, so a failure in either call frees each of them twice.

Setting the pointers to NULL after the free makes the error label's `free()` a no-op — matching what the surrounding code already does for the file handles (`users_file = NULL;` right after `fclose`).

## Verification

- Full `cmake --build` succeeds; `clang-format` reports no changes.
- `clang --analyze` reported `Attempt to free released memory` three times in `admin.c` before the change and reports one after — the remaining one is `admin.c:573`, in a different function that I deliberately did not touch because I have not traced its flow.
- `src/libpgagroal/management.c:1331` has the same analyzer finding and is likewise not claimed here.
- Verified by reading the code and with the analyzer, not by forcing `create_response` to fail at runtime.

I ran the same analysis across the family: pgexporter and pgvictoria show no `free released` findings, so this particular fix is pgagroal-only.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -839,14 +839,20 @@ password:
    fputs(entry, users_file);
 
    free(entry);
+   entry = NULL;
    free(master_key);
+   master_key = NULL;
    free(encrypted);
+   encrypted = NULL;
    free(encoded);
+   encoded = NULL;
    if (do_free)
    {
       free(password);
+      password = NULL;
    }
    free(verify);
+   verify = NULL;
 
    fclose(users_file);
    users_file = NULL;
@@ -1167,13 +1173,18 @@ password:
    }
 
    free(master_key);
+   master_key = NULL;
    free(encrypted);
+   encrypted = NULL;
    free(encoded);
+   encoded = NULL;
    if (do_free)
    {
       free(password);
+      password = NULL;
    }
    free(verify);
+   verify = NULL;
 
    fclose(users_file);
    users_file = NULL;


fluca1978:
ok for me, but remove the commit from another patch, and keep the authors changes, so it does matter which one we merge first.

Pranav-error:
Done — rebuilt on current master with just this fix, and its own AUTHORS/acknowledgement entry. The star_fds commit lives only in #999 now.

doc/manual/en/97-acknowledgement.md
```diff
@@ -59,6 +59,7 @@ Luca Ferrari <fluca1978@gmail.com>
 Haoran Zhang <andrewzhr9911@gmail.com>
 Bassam Adnan <mailbassam@gmail.com>
 Tejas Tyagi <tejastyagi.tt@gmail.com>
+R Sai Pranav <rajasaipranav0@gmail.com>
```
Userfrom1995: Not here. Append your name to the above list of Authors (contributors). This list is for committers.

You have made the same mistake in all of your PRs. Please fix them and append your name in the right place.

Also, there should be only one commit per PR, so squash them.

fluca1978: > Not here. Append your name to the above list of Authors (contributors). This list is for committers.

My fault while merging one the PRs, sorry about that, now fixed.
Thanks @Userfrom1995

fluca1978:
@Pranav-error please rebase, squash and force push

Pranav-error:
Done — rebased onto current master, single commit (`3176f41`), force-pushed. The diff is unchanged.

I dropped the AUTHORS/acknowledgement commit rather than squashing it in: both entries are already on master from #999, so re-adding them would have duplicated the lines (and would have conflicted with your "Fix contributor name position" commit).

Did the same on #1002, #1003 and #1005 so you don't have to ask four times — each is now one commit on top of current master.

fluca1978:
merged, thanks @Pranav-error

PR #1007, opened 2026-08-24
Guard the auth_query_get_password error path against an unset tmsg

### LINKED ISSUE #1004: auth_query_get_password: NULL dereference and infinite loop in the error path
Noticed while adding an allocation check in this function for #1003. I have not touched any of the below — reporting rather than bundling it into that PR.

All line numbers are from `src/libpgagroal/security.c` on `master` as of today.

## 1. `tmsg` is dereferenced in the error label while it can still be NULL

```c
   status = pgagroal_write_message(server_ssl, socket, &qmsg);
   if (status != MESSAGE_STATUS_OK)
   {
      goto error;                       /* tmsg is still NULL here */
   }

   status = pgagroal_read_block_message(server_ssl, socket, &tmsg);
   ...
error:
   pgagroal_log_trace(...);

   if (tmsg->kind == 'E')              /* dereference */
```

`tmsg` is initialised to NULL and only filled by `pgagroal_read_block_message`. The `goto error` on a failed **write** runs before that, so a write failure — a dropped connection to the server, for instance — dereferences NULL in the error handler.

## 2. The error label can jump to itself

```c
error:
   ...
   if (tmsg->kind == 'E')
   {
      char* error = NULL;

      if (pgagroal_extract_error_message(tmsg, &error))
      {
         goto error;                    /* back to the same label */
      }
```

If `pgagroal_extract_error_message` fails, this loops forever with no state change between iterations.

## 3. `result` is not checked before `memcpy`

```c
   result_size = dmsg->length - 11 + 1;
   result = (char*)calloc(1, result_size);
   memcpy(result, dmsg->data + 11, dmsg->length - 11);
```

## 4. `dmsg->length` is not checked against the 11 bytes that are skipped

`length` is `ssize_t`, so if a `'D'` message arrives with fewer than 11 bytes:

- `result_size` becomes zero or negative;
- `calloc(1, result_size)` converts a negative value to `size_t`, so the request is enormous and almost certainly fails, leaving `result` NULL;
- `memcpy(result, dmsg->data + 11, dmsg->length - 11)` then gets a negative length, also converted to a huge `size_t`.

This is the response to the auth query, so the length comes from the server. I have **not** checked whether `pgagroal_extract_message` already enforces a minimum length beyond the 4-byte header it documents — if it guarantees at least 11 bytes for `'D'`, then item 4 is not reachable and only items 1 to 3 stand. Worth confirming either way, since the arithmetic reads as unguarded at this call site.

## How this surfaced

I first wrote the #1003 allocation check as `goto error`, matching the rest of the file. It failed to build:

```
error: variable 'status' is used uninitialized whenever 'if' condition is true
       [-Werror,-Wsometimes-uninitialized]
```

`status` itself is fine on every path that exists today — it is assigned before all three current `goto error` statements. The warning appeared only because my new check sat above those assignments. Chasing it is what turned up items 1 and 2, so the check in #1003 returns directly instead of jumping to the label.

Happy to send a PR for any or all of these if you tell me how you would like the error path restructured — item 1 in particular could be fixed by guarding the dereference or by not routing pre-read failures through that label, and those are different calls.

Pranav-error:
Fixes #1004.

`error:` dereferences `tmsg->kind` unconditionally, but `tmsg` is still NULL when the label is reached from the write failure above the read, so that path was a NULL dereference rather than the trace-and-return every other failure gets. Guarded it — `tmsg is still NULL if the write above failed` is the actual reason, not a defensive habit.

The same block did `goto error` back to itself when `pgagroal_extract_error_message` failed. That loops forever with no state change between iterations. It now just skips logging the server's error text in that case and falls through to the return.

`result_size = dmsg->length - 11 + 1` was unchecked. A `'D'` message shorter than 11 bytes (the 1 byte kind + 4 byte length + the fixed header `pgagroal_extract_message` already strips) makes that negative, and both `result_size` and the `memcpy()` length are `size_t`, so it becomes an enormous value. `pgagroal_extract_message` does not enforce a minimum length — it trusts the length field from the wire — so this is reachable from a malformed response, not just a hypothetical. Rejected before the arithmetic. The `calloc()` result is also checked before the `memcpy()` writes through it.

## What I have not done

Restructuring so the `'E'` check is not reachable from a pre-read failure at all, rather than guarding it — that was closer to what you described in the issue thread. I went with the guard because it is the smaller change and keeps the one call site that legitimately needs the check (read succeeded, message was not `'D'`) exactly as it was. Happy to restructure it the other way if you would rather.

## Verification

`cppcheck` reports no new findings in the changed lines (the `nullPointerArithmeticOutOfMemory` warnings on `aq` above are pre-existing and #1003's territory). `clang-format` reports no changes. I have not been able to configure a full build locally, so this has not run — verified by reading and by the analyzer.

I used an AI assistant while preparing this. I understand the change and can explain it.

### DIFF
--- a/src/libpgagroal/security.c
+++ b/src/libpgagroal/security.c
@@ -5255,8 +5255,20 @@ auth_query_get_password(int socket, SSL* server_ssl, char* username, char* datab
       goto error;
    }
 
+   if (dmsg->length < 11)
+   {
+      /* Below the 1 byte kind + 4 byte length + 6 byte fixed header that
+       * pgagroal_extract_message() already copied. dmsg->length - 11 would
+       * be negative and, once converted to size_t below, an enormous value. */
+      goto error;
+   }
+
    result_size = dmsg->length - 11 + 1;
    result = (char*)calloc(1, result_size);
+   if (result == NULL)
+   {
+      goto error;
+   }
    memcpy(result, dmsg->data + 11, dmsg->length - 11);
 
    *password = result;
@@ -5270,17 +5282,17 @@ auth_query_get_password(int socket, SSL* server_ssl, char* username, char* datab
 error:
    pgagroal_log_trace("auth_query_get_password: socket (%d) status (%d)", socket, status);
 
-   if (tmsg->kind == 'E')
+   /* tmsg is still NULL if the write above failed, so this has to be
+    * guarded rather than assumed set just because the label was reached. */
+   if (tmsg != NULL && tmsg->kind == 'E')
    {
       char* error = NULL;
 
-      if (pgagroal_extract_error_message(tmsg, &error))
+      if (!pgagroal_extract_error_message(tmsg, &error))
       {
-         goto error;
+         pgagroal_log_error("%s in %s", error, database);
+         free(error);
       }
-
-      pgagroal_log_error("%s in %s", error, database);
-      free(error);
    }
 
    free(aq);


fluca1978:
merged, thanks @Pranav-error

PR #1008, opened 2026-08-24
[#997] Reject reserved words for databases and users

### LINKED ISSUE #997: Should we reject special names for user `all` and `replication` database?
While working on #993 I found that apparently user `all` in configuration was already blocked, good.
The problem is that `pgagroal-admin` allows the addition of a user `all`:

```
% pgagroal-admin  user add    
User name: all
Password : 
Verify   : 
Header:
  ClientVersion: 2.2.0
  Command: 20
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20260821092740
Outcome:
  Status: true
  Time: 00:00:08
Response:
  Users:
    - pgbench
    - all

```

and the configuration is ok if the user `all` is specified in the limit file and is found in the system (https://github.com/pgagroal/pgagroal/blob/master/src/libpgagroal/configuration.c#L1638).
My point is: does it make sense to materialize a `all` user (with a password) that cannot be used to connect to a database (since PostgreSQL rejects it, right?).

Should we force `pgagroal-admin` to avoid specific names like `all` and `replication` since they cannot be used in PostgreSQL?
Similarly, should we force configuration to reject a limit with a `replication` database?

fluca1978:
Add two new utility functions:
- pgagroal_is_username_reserved
- pgagroal_is_database_reserved

that accept a username or database name and return true if the name belongs (case insensitive) to the list of reserved words. This for example prevents the usage of 'all` as a valid entry for `pgagroal-admin` or to use `all` as an alias in entry like

      pgbench=all pgbench 10 5 3

since it would not make any sense.

`pool.c` has been modified to prevent prefil when either a username or a database is set to a reserved word. This is more strict than just avoiding the prefil for `all` as database/username, but it does not make sense at all to prefill against a dataabse/username that PostgreSQL will reject.

Added a specific test for the two new functions.

Close #997

### DIFF
--- a/src/libpgagroal/pool.c
+++ b/src/libpgagroal/pool.c
@@ -1292,6 +1292,19 @@ pgagroal_prefill(bool initial)
 
       if (size > 0)
       {
+         // do not perform prefil if
+         // - the database is a reserved word
+         // - the username is reserved
+         if (pgagroal_is_database_reserved(config->limits[i].database) || pgagroal_is_username_reserved(config->limits[i].username))
+         {
+            pgagroal_log_warn("Cannot perform prefill for database '%s' user '%s' at limit entry %d (row %d)",
+                              config->limits[i].database,
+                              config->limits[i].username,
+                              i + 1,
+                              config->limits[i].lineno);
+            continue;
+         }
+
          if (strcmp("all", config->limits[i].database) && strcmp("all", config->limits[i].username))
          {
             int user = -1;
@@ -1361,13 +1374,6 @@ pgagroal_prefill(bool initial)
                pgagroal_log_warn("Unknown user '%s' for limit entry (%d)", config->limits[i].username, i + 1);
             }
          }
-         else if (!strcmp("all", config->limits[i].database))
-         {
-            pgagroal_log_warn("Cannot perform prefill for database '%s' at limit entry %d (row %d)", config->limits[i].database,
-                              i + 1,
-                              config->limits[i].lineno);
-         }
-
          else
          {
             pgagroal_log_warn("Limit entry %d with invalid definition at row %d",
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -2267,3 +2267,53 @@ pgagroal_cleanse(void* data, size_t size)
       OPENSSL_cleanse(data, size);
    }
 }
+
+bool
+pgagroal_is_username_reserved(char* username)
+{
+   char* restricted_usernames[] = {"all", "sameuser"};
+   int i;
+   size_t count;
+
+   if (username == NULL || strlen(username) <= 0)
+   {
+      return false;
+   }
+
+   count = sizeof restricted_usernames / sizeof restricted_usernames[0];
+
+   for (i = 0; i < count; i++)
+   {
+      if (pgagroal_compare_string(username, restricted_usernames[i]))
+      {
+         return true;
+      }
+   }
+
+   return false;
+}
+
+bool
+pgagroal_is_database_reserved(char* database)
+{
+   char* restricted_databases[] = {"all", "replication"};
+   int i;
+   size_t count;
+
+   if (database == NULL || strlen(database) <= 0)
+   {
+      return false;
+   }
+
+   count = sizeof restricted_databases / sizeof restricted_databases[0];
+
+   for (i = 0; i < count; i++)
+   {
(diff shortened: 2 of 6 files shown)


fluca1978:
@jesperpedersen if the overall structure is fine I can merge or see if there are other parts to extend this (vault?)
I've used a couple of arrays to keep reserved words because:
- it gives us room for automatically add a new term (I don't think PostgreSQL will add a new one, but we could)
- it handles also `sameuser` and `replication` other than just `all`, since it does not make sense to configure limits for such entries too

src/libpgagroal/pool.c
```diff
@@ -1292,6 +1292,21 @@ pgagroal_prefill(bool initial)
 
       if (size > 0)
       {
+         /* 
+	 * do not perform prefil if
```
jesperpedersen, line 1296: Formatting

src/libpgagroal/utils.c
```diff
@@ -2262,3 +2262,45 @@ pgagroal_cleanse(void* data, size_t size)
...
+
+bool
+pgagroal_is_username_reserved(char* username)
+{
+   char* restricted_usernames[] = {"all", "sameuser"};
+   int i;
+   size_t count;
+
+   if (username == NULL || strlen(username) <= 0)
+      return false;
```
jesperpedersen, line 2274: Missing `{` and `}`

src/libpgagroal/utils.c
```diff
@@ -2262,3 +2262,45 @@ pgagroal_cleanse(void* data, size_t size)
...
+
+   count = sizeof restricted_usernames / sizeof restricted_usernames[0];
+
+   for (i = 0; i < count; i++)
+   {
+      if (!strcasecmp(username, restricted_usernames[i]))
+         return true;
```
jesperpedersen, line 2281: Ditto

src/libpgagroal/utils.c
```diff
@@ -2262,3 +2262,45 @@ pgagroal_cleanse(void* data, size_t size)
...
+
+bool
+pgagroal_is_database_reserved(char* database)
+{
+   char* restricted_databases[] = {"all", "replication"};
+   int i;
+   size_t count;
+
+   if (database == NULL || strlen(database) <= 0)
+      return false;
```
jesperpedersen, line 2295: Same

src/libpgagroal/utils.c
```diff
@@ -2262,3 +2262,45 @@ pgagroal_cleanse(void* data, size_t size)
...
+
+   count = sizeof restricted_databases / sizeof restricted_databases[0];
+
+   for (i = 0; i < count; i++)
+   {
+      if (!strcasecmp(database, restricted_databases[i]))
```
jesperpedersen, line 2301: pgagroal_compare_string

jesperpedersen:
@fluca1978 Other than that - LGTM

src/libpgagroal/pool.c
fluca1978: shame on me, definitely my `clang-format` is not working as expected...

src/libpgagroal/utils.c
fluca1978: why isn't `clag-format` catching this?

src/libpgagroal/utils.c
fluca1978: ok, but please note that `pgagroal_compare_string` uses `strcmp` instead of `strcasecmp`. It should not be a problem for rejected words here, but it means `Replication` will be a valid username.

jesperpedersen:
@fluca1978 You can change the signature

fluca1978:
> @fluca1978 You can change the signature

Uhm, I don't like this, better adding an explicit function name like `pgagroal_string_icompare` or something alike.
In this particular case, leave `strcmp` since `all` should be managed differently than `AlL`.

jesperpedersen:
@fluca1978 Yes, you can do that too

fluca1978:
Uhm...

```
postgres=# create role all;
ERROR:  syntax error at or near "all"
LINE 1: create role all;
                    ^
postgres=# create role "all";
CREATE ROLE
postgres=# create role "All";
CREATE ROLE
```

and even worst with databases (remember, `replication` is a "should not use word"):

```
postgres=# create database replication;
CREATE DATABASE
postgres=# create database "Replication";
CREATE DATABASE

```

So I guess the right way to right now is with case-sensitive comparison.

PR #1009, opened 2026-08-24
[#981] Verify standby replication against primary (slot-based)

### LINKED ISSUE #981: Validate standby is streaming from its configured primary
Currently (#977), a standby's streaming status is checked independently, pgagroal confirms a standby is streaming, but doesn't verify it's streaming from the primary pgagroal itself has configured for it. It's possible for a user to misconfigure things such that a "standby" entry is actually streaming from a different, unrelated primary.

See [Luca's comment on 977](https://github.com/pgagroal/pgagroal/pull/977#issuecomment-5250093350) for a proposed direction

Zeyad-Daowd:
**Description**
Once per health check period, pgagroal queries the primary's pg_stat_replication/pg_replication_slots and compares it against what each standby reports about itself (reusing the existing wal receiver call data so just the one extra query to the primary). Matching is done in C, slot name only. Address is stored for future extension (will be compared versus addresses of replicas in case of no slots)

Scope is slot-only for now, per the discussion in #981 - non-slot/address-based standbys aren't checked yet (see [luca's comment](https://github.com/pgagroal/pgagroal/issues/981#issuecomment-5341226748))

Two cases:
A configured standby reports a slot name that has no matching active connection on the primary = error (this is the actual misconfiguration that means replica streaming from the wrong primary).
The primary has an active slot connection that doesn't match any configured standby = warning (could be something else entirely, so kept lower severity).

Didn't want errors/warnings spamming the log every period once something's wrong, so I added transition tracking for both cases, logs once when it goes bad and once when it recovers.

**Trial** 

- the "primary" server has 2 replicas "replica" has no slot name & "slotted_replica" has a slot
- the "wrong_primary" server has no replicas

**Example for primary having a slotted replica that pgagroal doesn't see**
image
**Example for a slotted replica that is streaming from wrong primary**
image

closes #981


Zeyad-Daowd:
Note: I also refactored pgagroal_read_query_multiple_columns_text() to use the new multi-row reader instead of duplicating the same logic in two places. Behavior for all existing callers is unchanged, since the expect exactly one row.

src/libpgagroal/health.c
```diff
@@ -42,8 +42,14 @@
 #include <signal.h>
 #include <stdatomic.h>
 #include <sys/wait.h>
+#include <string.h>
 
 static void health_check_loop(void);
+static bool verify_standby_replication(char standby_slot_names[][MISC_LENGTH],
+                                       bool standby_has_slot[],
+                                       bool previous_ok[],
```
fluca1978: I'm wondering if it does make sense to add a field to `struct server` like `bool replication_ok` or alike, instead of passing around an array of previous states.

src/libpgagroal/health.c
```diff
@@ -231,6 +244,11 @@ health_check_loop(void)
                {
                   atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_NO);
                }
+               if (strlen(slot_name) > 0)
+               {
+                  standby_has_slot[i] = true;
+                  memcpy(standby_slot_names[i], slot_name, strlen(slot_name) + 1);
```
fluca1978: same thinking here, could it be worth addind a `slot` field to the `struct server` to handle the current slot replication name (for a standby) so to have a direct check?
One problem that could arise is if the standby changes the slot name (that requires some effort on the primary too), in such case the `struct server` could be out of sync and raise an error where it is a reconfiguration, but I think this is acceptable.

src/libpgagroal/queries.c
```diff
@@ -56,20 +57,27 @@ pgagroal_queries_wal_receiver_status(void)
    return "SELECT status, slot_name, sender_host, sender_port FROM pg_stat_wal_receiver;";
 }
 
+const char*
+pgagroal_queries_replication_slots_status(void)
+{
+   return "SELECT s.slot_name, r.client_addr FROM pg_stat_replication r LEFT JOIN pg_replication_slots s ON r.pid = s.active_pid;";
```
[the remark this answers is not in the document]
fluca1978: automatic review clearly pointed out we need at least `pg_monitor` privilege, so this has to be documented. I would also comment this on the query function.

fluca1978 (commented):
Seems fine as idea, I'm wondering if it is worth adding information to `struct server` for replication monitoring instead of passing around arrays to keep "previous" status.
In such case, also `pgagroall-cli status` and firends has to be adapted.

src/libpgagroal/queries.c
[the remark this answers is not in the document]
Zeyad-Daowd: Agreed. I'll add a comment above the query function noting the requirement, and I will update the documentation

src/libpgagroal/health.c
Zeyad-Daowd: Agreed. I will drop the arrays and add a bool directly to struct server to track it

src/libpgagroal/health.c
Zeyad-Daowd: Makes sense. I'll add char slot_name[MISC_LENGTH] to struct server instead of passing arrays.
I'm a little unsure about the edge case of the standby changing its slot name, though. Since we currently query the replica's WAL receiver to get the active slot name on each health check, we just update it dynamically if it changes. Unless the change happens in the exact middle of a single health probe, shouldn't it stay perfectly in sync?

Zeyad-Daowd:
@fluca1978 Hi Luca,
I've just pushed an update that addresses the feedback.

Here is a summary of the changes:

1. Moved state to struct server: Added slot_name and replication_ok (atomic_bool) directly to struct server. The health check loop no longer passes arrays around.
2. slot_name is conditionally overwritten via memcpy only if the standby actually changes its slot
3. Documentation & Comments: Added the requirement for GRANT pg_monitor in 19-health_check.md and added explanatory comments above the query functions.
4. CLI status details: Added a "Replication valid": true/false field for standbys. It is dynamically shown only if there is a slot name for this standby


example when using the wrong primary(the one with no replicas):
image

Let me know what you think

fluca1978:
Looks good to me, I would prefer to see names like `ReplicationStatus` instead of `ReplicationValid` and `ReplicationSlotName` as the output of CLI.

@jesperpedersen thoughts?
@jesperpedersen note that I've placed several comments for variable declared in scope of iteration (e.g., `for( int i = 0; `, is this still a thing?

Zeyad-Daowd:
Hi @fluca1978, I went ahead and renamed them to ReplicationStatus/ReplicationSlotName locally per your suggestion, happy to adjust if @jesperpedersen prefers something different.

Also a quick question on the declaration placement, should locals be at the top of their immediate block, or the top of the whole function? I assumed immediate block, since I saw that pattern in existing code I didn't write (ex: pgagroal_read_query_first_column_text), but is_replication_ok (in status.c) was already the first line inside its if block when you flagged it, so I want to make sure I'm applying the right rule before going through the rest.

fluca1978:
> Hi @fluca1978, I went ahead and renamed them to ReplicationStatus/ReplicationSlotName locally per your suggestion, happy to adjust if @jesperpedersen prefers something different.
> 

Ok

> Also a quick question on the declaration placement, should locals be at the top of their immediate block, or the top of the whole function? I assumed immediate block, since I saw that pattern in existing code I didn't write (ex: pgagroal_read_query_first_column_text), but is_replication_ok (in status.c) was already the first line inside its if block when you flagged it, so I want to make sure I'm applying the right rule before going through the rest.

We need a policy here. I prefer keeping variables inside their block to avoid one of them escaping the context, but in the past we used to have all to the beginning of the **function** (i.e., not within the scope block).
It could be that code reviews did not all applied this rule, we should stick to one and use wherever possible.
@jesperpedersen I suspect you want all declarations at the begin of the function, right?

src/libpgagroal/health.c
```diff
@@ -348,3 +370,139 @@ server_probe(int server_idx, bool* up, int* auth_type)
...
+   }
+
+   if (pgagroal_server_get_replication_slots_status(primary, primary_slot_names, primary_client_addrs,
+                                                    NUMBER_OF_SERVERS, &primary_rows))
+   {
+      pgagroal_log_debug("Failed to get replication slots status from primary '%s'", config->servers[primary].name);
+      return true;
+   }
+
+   FOREACH_VALID_SERVER
```
jesperpedersen, line 398: No

src/libpgagroal/health.c
```diff
@@ -348,3 +370,139 @@ server_probe(int server_idx, bool* up, int* auth_type)
...
+   {
+      if (i == primary || config->servers[i].slot_name[0] == '\0')
```
jesperpedersen, line 400: No

jesperpedersen:
`ReplicationStatus` is good.

Using macros for iteration, and hacks for testing isn't

Zeyad-Daowd:
I'll replace the FOREACH_VALID_SERVER macro with explicit for loops.
Could you clarify what you mean by "hacks for testing"? I don't think I added anything test-related in this PR, want to make sure I'm not missing something you're seeing.

Zeyad-Daowd:
@jesperpedersen @fluca1978 I’ve replaced FOREACH_VALID_SERVER with explicit for loops. I also moved the local declarations to their immediate blocks to keep them scoped to where they’re used. If you prefer declarations at the beginning of the function, I can adjust them accordingly.

fluca1978:
> @jesperpedersen @fluca1978 I’ve replaced FOREACH_VALID_SERVER with explicit for loops. I also moved the local declarations to their immediate blocks to keep them scoped to where they’re used. If you prefer declarations at the beginning of the function, I can adjust them accordingly.

Ok, I still don't get why we need explicit loops to skip in any case not valid servers.
@jesperpedersen what do you thjink about variable declaration position?

src/include/management.h
```diff
@@ -141,9 +141,10 @@ extern "C" {
 #define MANAGEMENT_ARGUMENT_STANDBYS            "Standbys"
 #define MANAGEMENT_ARGUMENT_STREAMING           "Streaming"
 #define MANAGEMENT_ARGUMENT_WAL_RECEIVER_STATUS "WALReceiverStatus"
-#define MANAGEMENT_ARGUMENT_SLOT_NAME           "SlotName"
+#define MANAGEMENT_ARGUMENT_SLOT_NAME           "ReplicationSlotName"
```
jesperpedersen: Wrong constant name

src/include/pgagroal.h
```diff
@@ -411,6 +411,8 @@ struct server
    atomic_schar state;            /**< The state of the server */
    atomic_schar health_state;     /**< The health state of the server */
    atomic_int streaming_state;    /**< The streaming state of the server SERVER_STREAMING_PRIMARY/NO/YES */
+   char slot_name[MISC_LENGTH];   /**< The replication slot name for this standby */
```
jesperpedersen: Missing `replication_`

src/include/queries.h
```diff
@@ -63,6 +63,32 @@ pgagroal_queries_replication_lag_bytes(void);
...
 
+/*
+ * SQL for replication slots status (multiple columns, multiple rows).
+ * Note: Executing this query requires the database user to have at least 
+ * the 'pg_monitor' privilege, as it reads from 
+ * the pg_stat_replication view.
+ * @return Static query string (do not free)
+*/
+
+const char*
```
jesperpedersen: We don't use `const`

src/libpgagroal/health.c
```diff
@@ -142,6 +147,8 @@ health_check_loop(void)
    for (int i = 0; i < NUMBER_OF_SERVERS; i++)
    {
       previous_state[i] = -2; /* Initial value representing 'never checked' */
+      atomic_store(&config->servers[i].replication_ok, true);
```
jesperpedersen, line 150: Where is the `_init` ?

src/libpgagroal/health.c
```diff
@@ -231,6 +238,20 @@ health_check_loop(void)
                {
                   atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_NO);
                }
+               if (strlen(slot_name) > 0)
```
jesperpedersen: We need to rewrite this - either pgagroal_compare_string true or false.

We don't want code blocks like this all over

src/libpgagroal/health.c
```diff
@@ -348,3 +370,149 @@ server_probe(int server_idx, bool* up, int* auth_type)
...
+      if (strlen(primary_slot_names[row_idx]) == 0)
+      {
+         continue;
+      }
+
+      for (int i = 0; i < config->number_of_servers; i++)
+      {
+         if (config->servers[i].valid)
+         {
+            if (i == primary || config->servers[i].slot_name[0] == '\0')
```
jesperpedersen: Fold into pgagroal_compare_string

src/libpgagroal/queries.c
```diff
@@ -31,11 +31,12 @@
 #include <queries.h>
 #include <utils.h>
 #include <string.h>
+#include <logging.h>
```
jesperpedersen: Sorting

src/libpgagroal/queries.c
```diff
@@ -56,31 +57,43 @@ pgagroal_queries_wal_receiver_status(void)
    return "SELECT status, slot_name, sender_host, sender_port FROM pg_stat_wal_receiver;";
 }
 
+/* Note: Requires pg_monitor privilege to read from pg_stat_replication */
+
+const char*
```
jesperpedersen: const

src/libpgagroal/queries.c
```diff
@@ -56,31 +57,43 @@ pgagroal_queries_wal_receiver_status(void)
...
    struct message* msg = NULL;
 
-   if (values == NULL || value_sizes == NULL || expected_cols <= 0)
+   if (values == NULL || value_sizes == NULL || num_rows == NULL || expected_cols <= 0 || max_rows <= 0)
    {
       return 1;
    }
 
-   for (int i = 0; i < expected_cols; i++)
+   for (cell_idx = 0; cell_idx < max_rows * expected_cols; cell_idx++)
```
jesperpedersen: int cell_idx - init in the for loop

src/libpgagroal/queries.c
```diff
@@ -104,39 +117,41 @@ pgagroal_read_query_multiple_columns_text(int fd, int expected_cols, char** valu
...
-                     {
-                        memcpy(values[i], msg->data + dr_offset, (size_t)col_len);
-                        values[i][col_len] = '\0';
-                     }
-                     else
-                     {
-                        goto error;
-                     }
+                     memcpy(dest, msg->data + dr_offset, (size_t)col_len);
+                     dest[col_len] = '\0';
```
jesperpedersen: We use memset

src/libpgagroal/server.c
```diff
@@ -730,6 +726,69 @@ pgagroal_server_get_connectivity_info(int server, char** status, char** primary,
...
+      return 1;
+   }
+
+   if (strlen(config->health_check_user) == 0)
+   {
+      pgagroal_log_debug("ping/status: health_check_user is not configured, cannot check replication slots");
+      return 1;
+   }
+
+   for (row_idx = 0; row_idx < max_rows; row_idx++)
```
jesperpedersen: Same

jesperpedersen (changes requested):

src/include/pgagroal.h
fluca1978: > Missing `replication_`

we should also convert `streaming_state` to `replication_streaming_state` then? Better using `repl_` as prefix to get shorter names.

Zeyad-Daowd:
> Ok, I still don't get why we need explicit loops to skip in any case not valid servers.

From my understanding invalid servers are kind of disabled and are not used, I understand that valid flags is set to false in case of duplicate system identifiers.
I thought it would be unnecessary to query them again
@fluca1978 What do you think?

src/libpgagroal/server.c
Zeyad-Daowd: will do

src/libpgagroal/queries.c
Zeyad-Daowd: Noted, thanks for letting me know

src/libpgagroal/queries.c
Zeyad-Daowd: will do

src/include/management.h
Zeyad-Daowd: I will fix it

src/include/pgagroal.h
Zeyad-Daowd: will add it

src/include/queries.h
Zeyad-Daowd: noted, will remove it
should I remove it from the wal query as well?

src/libpgagroal/health.c
Zeyad-Daowd: I'll add the atomic_init for replication_ok to pgagroal_init_configuration() and move the initialization of replication_slot_name there as well. I'll then remove both from the health check loop.

src/libpgagroal/health.c
Zeyad-Daowd: understood, will change it and other strlen

src/libpgagroal/health.c
Zeyad-Daowd: will do

src/libpgagroal/queries.c
Zeyad-Daowd: I moved logging along with pgagroal headers and added a /* pgagroal */ & /* system */ tag similar to health.c file

src/libpgagroal/queries.c
Zeyad-Daowd: done

Zeyad-Daowd:
@fluca1978 @jesperpedersen I found something that i needed your input on
I have added `atomic_init(&config->servers[i].replication_ok, true)` to the configuration.c init configuration function and I found out it correctly initializes all servers to `1`. However, later servers replication_ok server becomes become `0`. It looks like this is because `struct server` is copied with `memcpy()` in `configuration.c`, and `srv` is zeroed with `memset()` before the copy. Since `replication_ok` is now part of `struct server`, the copy also overwrites it. 
I thought about adding an atomic_store to health.c in the beginning of the loop but i am not sure if this is a good handling

How would you prefer me to handle this?

I also added the following debug lines at read_configuration
image
and found the following
image

fluca1978:
> @fluca1978 @jesperpedersen I found something that i needed your input on I have added `atomic_init(&config->servers[i].replication_ok, true)` to the configuration.c init configuration function and I found out it correctly initializes all servers to `1`. However, later servers replication_ok server becomes become `0`. It looks like this is because `struct server` is copied with `memcpy()` in `configuration.c`, and `srv` is zeroed with `memset()` before the copy. Since `replication_ok` is now part of `struct server`, the copy also overwrites it. I thought about adding an atomic_store to health.c in the beginning of the loop but i am not sure if this is a good handling

AFAIK ìmemcpy` is not safe for atomic fields, so there is the need to initialize again after the copy all the atomic fields.

fluca1978:
> > Ok, I still don't get why we need explicit loops to skip in any case not valid servers.
> 
> From my understanding invalid servers are kind of disabled and are not used, I understand that valid flags is set to false in case of duplicate system identifiers. I thought it would be unnecessary to query them again @fluca1978 What do you think?

My same point, it is not clear to me why @jesperpedersen didn't like it.

Zeyad-Daowd:
> AFAIK ìmemcpy` is not safe for atomic fields, so there is the need to initialize again after the copy all the atomic fields.

@fluca1978 I didn't know that thanks for letting me know.
 I think I can have an atomic_store at the beginning of the health loop to account for the copy (this feels to me like the simplest approach). What do you think?

Zeyad-Daowd:
I went ahead with the approach I mentioned above and added an atomic_store at the beginning of the health loop to account for the value being overwritten by the memcpy().
I think this is the simplest approach because I think that struct server is copied in multiple places, but I'm happy to revert/change it if you think handling this differently would be preferable.

fluca1978:
looks good to me, @jesperpedersen WDYT?

src/libpgagroal/health.c
```diff
@@ -42,8 +42,11 @@
 #include <signal.h>
 #include <stdatomic.h>
 #include <sys/wait.h>
+#include <string.h>
```
jesperpedersen, line 43: Sorting

src/libpgagroal/health.c
```diff
@@ -244,11 +264,12 @@ health_check_loop(void)
             atomic_store(&config->servers[i].streaming_state, SERVER_STREAMING_NO);
          }
       }
+
+      verify_standby_replication(previous_warned_slots, &previous_warned_count);
    }
 
    pgagroal_log_info("Health check stopped");
 }
-
```
jesperpedersen, line 251: Keep the linefeed

src/libpgagroal/health.c
```diff
@@ -348,3 +369,149 @@ server_probe(int server_idx, bool* up, int* auth_type)
    }
    return 1;
 }
+
+static bool
```
jesperpedersen: I think need an `int` here - and constants

src/libpgagroal/health.c
```diff
@@ -348,3 +369,149 @@ server_probe(int server_idx, bool* up, int* auth_type)
...
+   {
+      pgagroal_log_debug("Could not determine primary server, skipping replication verification");
+      return true;
+   }
+
+   if (pgagroal_server_get_replication_slots_status(primary, primary_slot_names, primary_client_addrs,
+                                                    NUMBER_OF_SERVERS, &primary_rows))
+   {
+      pgagroal_log_debug("Failed to get replication slots status from primary '%s'", config->servers[primary].name);
+      return true;
```
jesperpedersen: I find this confusing

src/libpgagroal/health.c
```diff
@@ -348,3 +369,149 @@ server_probe(int server_idx, bool* up, int* auth_type)
...
+   int new_warned_count = 0;
+   int row_idx = 0;
+   bool already_warned = false;
+   int warned_idx = 0;
+   int new_warned_idx = 0;
+
+   if (pgagroal_get_primary(&primary))
+   {
+      pgagroal_log_debug("Could not determine primary server, skipping replication verification");
+      return true;
```
jesperpedersen: I find this confusing

src/libpgagroal/queries.c
```diff
@@ -26,16 +26,20 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+/* pgagroal */
```
jesperpedersen, line 29: Why ?

src/libpgagroal/queries.c
```diff
@@ -26,16 +26,20 @@
...
 #include <pgagroal.h>
 #include <message.h>
 #include <queries.h>
 #include <utils.h>
+#include <logging.h>
```
jesperpedersen: Sorting

jesperpedersen (changes requested):

src/libpgagroal/queries.c
Zeyad-Daowd: I was mimicing health.c 
```
/* pgagroal */
#include <health.h>
#include <logging.h>
#include <message.h>
#include <network.h>
#include <pgagroal.h>
#include <server.h>
#include <shmem.h>
#include <utils.h>

/* system */
#include <signal.h>
#include <stdatomic.h>
#include <stdlib.h>
#include <string.h>
#include <sys/wait.h>
#include <unistd.h>
```

should I remove them (/* pgagroal \*/ and /* system \*/)?

src/libpgagroal/queries.c
Zeyad-Daowd: I will sort includes alphebitically

src/libpgagroal/health.c
Zeyad-Daowd: will do

src/libpgagroal/health.c
Zeyad-Daowd: I will sort includes alphebitically

src/libpgagroal/health.c
Zeyad-Daowd: makes sense, will change it to be integer based with tristate (success/ failure/ check skipped)

src/libpgagroal/health.c
Zeyad-Daowd: changed to be the HEALTH_CHECK_REPLICATION_VERIFY_SKIPPED constant

src/libpgagroal/health.c
Zeyad-Daowd: changed return to be the HEALTH_CHECK_REPLICATION_VERIFY_SKIPPED constant

src/libpgagroal/health.c
Zeyad-Daowd: I changed verify_standby_replication to return int with named constants in pgagroal.h: HEALTH_CHECK_REPLICATION_VERIFY_OK (0), HEALTH_CHECK_REPLICATION_VERIFY_FAILED (1), HEALTH_CHECK_REPLICATION_VERIFY_SKIPPED (2) - grouped with the existing HEALTH_CHECK_* constants. Both early-return paths (couldn't determine primary / couldn't fetch replication slot status) now return SKIPPED instead of the old true, so they're distinguishable from an actual passing check. The final return now maps all_ok to OK/FAILED. 
One open question: health_check_loop() currently calls this and ignores the return value, should that be logged or something in case of skipping, or is it fine for now

Zeyad-Daowd:
Hi @jesperpedersen, I just Pushed changes addressing the review comments, reasoning is on the respective threads .(return type/constants on the replication verification function, and the include comment convention). 
Let me know if anything needs another look.

fluca1978:
@Zeyad-Daowd please rebase.
Also do not put images if not necessary, use text please

Zeyad-Daowd:
@fluca1978 Rebased.Noted on images too, I'll stick to text going forward. Please let me know if you'd also like me to swap the existing screenshots in the description for text.

fluca1978:
LGTM, @jesperpedersen thoguts?
@Zeyad-Daowd there is no need to switch the altedy posted images

PR #1010, opened 2026-08-24
Remove invalid initializers from atomic signal arrays

MohamedM216:
Clang fails to compile `ev.c` because `{0}` is not a valid constant initializer for the` _Atomic` pointer/function-pointer arrays `signal_watchers` and `signal_callbacks`. Since these variables have static storage duration, the explicit initializers can be removed and they will still be zero-initialized automatically.


fluca1978:
merged, thanks @MohamedM216

PR #1011, opened 2026-08-24
[#973] Fix reading split PostgreSQL message headers

### LINKED ISSUE #973: Transaction pipeline treats a TCP chunk's first byte as `ReadyForQuery`

MohamedM216:
Introduces a new structure `postgresql_message_state` that is shared across several continuos invocations of `pgagroal_parse_message` to inspect the header and the content arrived so far.

`pgagroal_parse_message` gets the bytes of an incoming message and look to see if the header is complete or not, and unless the header is complete and a first payload byte has arrived it does nothing but keeping the status of incoming data.
Once the first byte of payload arrives, a callback function is invoked to parse the header and possibly the first byte. `pgagroal_message_state` is used as per-connection client/server parser message state declared in `io_watcher` instead of being global variables.

The critical fix: transaction_server and session_server were using msg->kind (the first byte of the raw TCP chunk) to decide when to return a connection to the pool. A large result whose payload contains 'Z' bytes (e.g. repeat('Z', 2048)) could land a TCP read starting on one of those bytes, making msg->kind == 'Z' while in_tx is false, and triggering a premature connection return mid-stream. The same class of bug affected the 'E'/fatal detection.

Both decisions are now driven by a saw_rfq flag and the fatal flag set inside the callback, which fires only at true protocol message boundaries. The 'E' check is also moved into the callback where msg + 6 is guaranteed safe.

The two per-pipeline server callbacks (transaction_server_message and session_server_message) are merged into a single shared pgagroal_pipeline_server_rfq callback in message.c, alongside pgagroal_parse_message where the callback type is defined.

Close #973

### DIFF
--- a/src/include/message.h
+++ b/src/include/message.h
@@ -134,6 +134,85 @@ void
 pgagroal_free_message(struct message* msg);
 
 /**
+ * Callback invoked by pgagroal_parse_message once per complete message boundary.
+ * For messages with a payload it fires when both the header and the first payload
+ * byte have arrived, so the callback can safely read msg[5] (e.g. the transaction
+ * state byte of a ReadyForQuery 'Z' message). A NULL arg is valid; the callback
+ * is responsible for guarding against it if needed.
+ * @param kind The message type byte (e.g. 'Z', 'E', 'D')
+ * @param msg Pointer to the message start (kind byte included); at least 6 bytes
+ *            are valid when msglen > 5
+ * @param msglen Declared total length of the message (kind byte + 4-byte length field + payload)
+ * @param arg Caller-supplied context passed through unchanged from pgagroal_parse_message
+ */
+typedef void (*pgagroal_message_callback)(char kind, char* msg, int msglen, void* arg);
+
+/**
+ * Parser state for pgagroal_parse_message. Zero-initialize before the first
+ * call and pass the same instance on every subsequent call for the same
+ * connection so the parser can resume exactly where the previous read stopped.
+ * All fields are private to the parser; callers must not modify them directly.
+ */
+struct pgagroal_message_state
+{
+   char header[5];          /**< Partial header buffer (kind byte + 4-byte length field) */
+   char first_payload_byte; /**< First byte of the payload, e.g. transaction state for 'Z' */
+   int header_len;          /**< Bytes of the header received so far (0-5); 5 means complete */
+   int payload_remaining;   /**< Payload bytes not yet consumed from the current message */
+};
+
+/**
+ * Read a buffer holding one or more concatenated PostgreSQL protocol messages
+ * that may be split across multiple reads. The caller keeps the parser state
+ * in `state` (zero it before the first call); it is updated so a subsequent
+ * call can continue exactly where this one stopped.
+ *
+ * The callback is invoked once per message. For a message without a payload
+ * it fires as soon as the header (kind byte + length field) is complete; for
+ * a message with a payload it fires when the header and the first payload
+ * byte are present, so the callback can read past the header (e.g. the
+ * transaction state byte of a Z message). A header split across reads is
+ * buffered in the state until it can be parsed, so a split never
+ * desynchronizes the parser.
+ * @param state Parser state, zero-initialized before the first call
+ * @param data The buffer holding the received bytes
+ * @param length The number of received bytes
+ * @param callback The callback to invoke per message
+ * @param arg An argument passed through to the callback
+ */
+void
+pgagroal_parse_message(struct pgagroal_message_state* state,
+                       char* data,
+                       int length,
+                       pgagroal_message_callback callback, void* arg);
+
+/**
+ * Caller-supplied context passed to pgagroal_pipeline_server_rfq via the
+ * pgagroal_parse_message arg parameter. All pointer fields are optional;
+ * pass NULL for any field that does not need tracking.
+ */
+struct pipeline_server_state
+{
+   bool* in_tx;   /**< Updated with the transaction state from 'Z' messages */
+   bool* saw_rfq; /**< Set to true when a genuine idle ReadyForQuery is seen */
+   bool* fatal;   /**< Set to true when an 'E' message carries FATAL or PANIC */
+};
+
+/**
+ * Shared server-side message callback for the session and transaction
+ * pipelines. Updates the pipeline transaction state and prometheus counters
+ * when a genuine ReadyForQuery ('Z') boundary is seen, and sets the fatal
+ * flag when a genuine ErrorResponse ('E') carries FATAL or PANIC.
+ * Safe to call with a NULL arg; returns immediately in that case.
+ * @param kind The message type byte
+ * @param msg Pointer to the message start (kind byte included)
+ * @param msglen Declared total length of the message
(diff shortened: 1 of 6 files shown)


fluca1978:
According to @jesperpedersen comment we don't name it `postgresql_xxx`, am I right? <https://github.com/pgagroal/pgagroal/pull/996#issuecomment-5366127238>

So it can become `pipeline_` whatever instead of `postgresql_`?

MohamedM216:
@fluca1978 I've pushed a new patch to fix the comments formatting and rename `postgresql` entities to follow the naming conventions.

src/include/message.h
```diff
@@ -133,6 +133,71 @@ pgagroal_copy_message(struct message* msg);
...
+ * @param callback The callback to invoke per message
+ * @param arg An argument passed through to the callback
+ */
+void
+pgagroal_parse_message(struct pgagroal_message_state* state,
+                       char* data,
+                       int length,
+                       pgagroal_message_callback callback, void* arg);
+
+struct pipeline_server_state
```
jesperpedersen, line 194: No docs

src/include/message.h
```diff
@@ -133,6 +133,71 @@ pgagroal_copy_message(struct message* msg);
...
+ * \param msg is a  pointer to the message start (ATTENTION: kind byte included)
+ * \param msglen the declared total length of the message (kind byte + payload)
+ * \param arg any caller supplied argument (e.g., io related structs)
+ */
+typedef void (*pgagroal_message_callback)(char kind, char* msg, int msglen, void* arg);
+
+/**
+ * A struct that stores the current state of a PostgreSQL message parsing.
+ * It is used by pgagroal_parse_message across all the chunks.
+ */
```
jesperpedersen, line 155: Docs

src/include/message.h
```diff
@@ -133,6 +133,71 @@ pgagroal_copy_message(struct message* msg);
...
+   bool* saw_rfq;
+   bool* fatal;
+};
+
+/**
+ * Shared server-side message callback for the session and transaction
+ * pipelines. Updates the pipeline in_tx flag and prometheus counters
+ * when a genuine ReadyForQuery ('Z') boundary is seen, and sets fatal
+ * on a genuine ErrorResponse ('E') carrying FATAL or PANIC.
+ */
```
jesperpedersen, line 211: Docs

jesperpedersen (changes requested):

jesperpedersen:
@MohamedM216 You need to have the full documentation chain installed and working

MohamedM216:
Failing tests are unrelated to this PR. They pass locally.

MohamedM216:
Hi @jesperpedersen , I've added more documentation in the previous patch, is it ready to be merged now?

Userfrom1995:
> Failing tests are unrelated to this PR. They pass locally.

They do seem to fail locally on my end as well, at least with the changes in this PR. Could you please take a look?

MohamedM216:
@Userfrom1995 which configuration you used? I'm using the default .conf files. I reran the tests again, and no test fails locally. I'll try to investigate more. But for the first glance, may be they are flaky tests? Also what's more important is they aren't related to this PR changes. @jesperpedersen @fluca1978 Any thoughts?

Userfrom1995:
> @Userfrom1995 which configuration you used? I'm using the default .conf files. I reran the tests again, and no test fails locally. I'll try to investigate more. But for the first glance, may be they are flaky tests? Also what's more important is they aren't related to this PR changes.

I ran:

```sh
./test/check.sh run-configs
```

and I'm still seeing the same failure locally as in CI. On master, the same command passes both locally and in CI, so it seems the failure may be related to the changes in this PR.

It could still be a flaky test, but I think it would be worth investigating further since I can reproduce the failure locally with the PR changes.

MohamedM216:
@Userfrom1995 I ran the same command you used `./test/check.sh run-configs` and it passes locally :) you can have a look at https://github.com/pgagroal/pgagroal/pull/1037 and see that some tests that fail here already pass there or even timeout instead, although it's a clean version.

PR #1028, opened 2026-09-06
Size the number buffers with sizeof

### LINKED ISSUE #1027: append_int, append_ulong and append_ullong truncate the largest value
The three number-formatting helpers in `utils.c` pass a size one smaller than the buffer they write into, so the largest value of each type is truncated.

```c
char number[12];
pgagroal_snprintf(&number[0], 11, "%d", i);

char number[21];
pgagroal_snprintf(&number[0], 20, "%lu", l);

char number[21];
pgagroal_snprintf(&number[0], 20, "%llu", l);
```

The buffers are the right size. `INT_MIN` is `-2147483648`, 11 characters plus a terminator, and `ULONG_MAX` is 20 digits plus a terminator — but the size argument is the digit count rather than `sizeof(number)`, so `snprintf` has one byte too few and drops the last digit:

```
append_int(INT_MIN)        size 11 -> "-214748364"            correct: -2147483648
append_ulong(ULONG_MAX)    size 20 -> "1844674407370955161"   correct: 18446744073709551615
append_ullong(ULLONG_MAX)  size 20 -> "1844674407370955161"   correct: 18446744073709551615
```

There is no overflow — `snprintf` truncates safely. The result is a wrong number, not a crash.

**How reachable this is, honestly:** low. The main callers are the Prometheus exporters (89 calls in `prometheus.c`), and a query or transaction counter reaching 10^19 is not going to happen. `append_int` is the one that is actually reachable, since `INT_MIN` is a single specific value rather than an extreme — though the current callers pass ports, versions and pipeline identifiers, none of which reach it.

So this is a latent correctness bug rather than something biting today. I am filing it because the fix is to write `sizeof(number)`, which is the idiom used everywhere else in the tree (`admin.c` does `pgagroal_snprintf(&buf[0], sizeof(buf), ...)`), and because it removes a trap for any future caller that does pass a large value.

The same three functions with the same off-by-one are in pgmoneta, pgexporter and pgvictoria. Happy to cross-port once you have looked at this one.

Pranav-error:
Fixes #1027.

`pgagroal_append_int()`, `pgagroal_append_ulong()` and `pgagroal_append_ullong()` pass the digit count where `snprintf` wants the buffer size, so each one is a byte short and drops the last digit of the largest value of its type:

```
append_int(INT_MIN)        size 11 -> "-214748364"            correct: -2147483648
append_ulong(ULONG_MAX)    size 20 -> "1844674407370955161"   correct: 18446744073709551615
append_ullong(ULLONG_MAX)  size 20 -> "1844674407370955161"   correct: 18446744073709551615
```

The buffers themselves are already the right size, so this just uses `sizeof(number)` — which is what `admin.c` and the rest of the tree do.

There is no overflow here; `snprintf` truncates safely. It produces a wrong number, not a crash, and as I said in the issue the reachable impact today is small — the Prometheus counters that use these will not reach 10^19. It is a trap for a future caller more than a live bug.

## Verification

The table above is from running both forms side by side against `INT_MIN`, `ULONG_MAX` and `ULLONG_MAX`.

`clang -fsyntax-only -I src/include` is clean and `clang-format` reports no changes. `utils.c` needs `-DPGAGROAL_MAJOR_VERSION` and friends plus a `secure_getenv`/`setproctitle` on macOS; unmodified `master` reports the same 6 errors without them, so they are not from this change.

### DIFF
--- a/src/libpgagroal/utils.c
+++ b/src/libpgagroal/utils.c
@@ -919,17 +919,30 @@ char*
 pgagroal_format_and_append(char* buf, char* format, ...)
 {
    va_list args;
-   va_start(args, format);
+   int len;
+   char* formatted_str = NULL;
 
    // Determine the required buffer size
-   int size_needed = vsnprintf(NULL, 0, format, args) + 1;
+   va_start(args, format);
+   len = vsnprintf(NULL, 0, format, args);
    va_end(args);
 
+   // Leave buf as it is on failure, like pgagroal_append() does
+   if (len < 0)
+   {
+      return buf;
+   }
+
    // Allocate buffer to hold the formatted string
-   char* formatted_str = malloc(size_needed);
+   formatted_str = malloc((size_t)len + 1);
+
+   if (formatted_str == NULL)
+   {
+      return buf;
+   }
 
    va_start(args, format);
-   vsnprintf(formatted_str, size_needed, format, args);
+   vsnprintf(formatted_str, (size_t)len + 1, format, args);
    va_end(args);
 
    buf = pgagroal_append(buf, formatted_str);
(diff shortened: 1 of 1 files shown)


src/libpgagroal/utils.c
```diff
@@ -945,7 +945,7 @@ pgagroal_append_int(char* orig, int i)
    char number[12];
 
    memset(&number[0], 0, sizeof(number));
-   pgagroal_snprintf(&number[0], 11, "%d", i);
+   pgagroal_snprintf(&number[0], sizeof(number), "%d", i);
```
fluca1978: should we check that the `sizeof(number)` is less then size of the buffer?

fluca1978 (changes requested):
I think this is ok but we need to check the buffer is longer than the `sizeof(number)`, this will result in zero-ed strings all the time.

src/libpgagroal/utils.c
jesperpedersen: or we can use the _append_ functions

PR #1044, opened 2026-09-09
Fix fd for anonymous mmap fallback

### LINKED ISSUE #1043: Use -1 for the file descriptor in anonymous mmap() in shmem creation
In pgagroal_create_shared_memory() in `src/libpgagroal/shmem.c`, the fallback mmap() uses 0 as the file descriptor:

```c
s = mmap(NULL, size, protection, visibility, 0, 0);
```

Since the mapping uses `MAP_ANONYMOUS`, the file descriptor is not used on Linux, so this works there.

However, `mmap()` implementations on FreeBSD and OpenBSD require the file descriptor to be `-1` for anonymous mappings. pgagroal supports both platforms.

The initial `mmap()` in the same function already uses `-1`:

```c
s = mmap(NULL, size, protection, visibility, -1, 0);
```

The fallback should therefore use the same value:

```c
s = mmap(NULL, size, protection, visibility, -1, 0);
```

This would make the anonymous mapping portable and consistent across the two calls.

References:

* Linux `mmap(2)`: https://man7.org/linux/man-pages/man2/mmap.2.html
* FreeBSD `mmap(2)`: https://man.freebsd.org/cgi/man.cgi?query=mmap
* OpenBSD `mmap(2)`: https://man.openbsd.org/mmap.2
* pgagroal supported platforms: https://github.com/pgagroal/pgagroal#pgagroal

AbdulrahmanXSO25:
Use -1 instead of 0 as fd for the fallback anonymous mmap() in pgagroal_create_shared_memory() to match the initial call and fix portability on FreeBSD and OpenBSD which require -1 for MAP_ANONYMOUS. Update AUTHORS.

Close #1043

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,3 +45,4 @@ R Sai Pranav <rajasaipranav0@gmail.com>
 Esraa Yehia <yehiae996@gmail.com>
 Yousuf Mahmoud <yusufmahmoud.dev@gmail.com>
 Youssef Sherief <youssefsherief2718@gmail.com>
+Abdulrahman Nader <a0xnader.oss@outlook.com>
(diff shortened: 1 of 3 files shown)


AbdulrahmanXSO25:
@jesperpedersen

PR #1061, opened 2026-09-17
Use portable sed and cpu count in check.sh

### LINKED ISSUE #1060: check.sh uses GNU-only sed -i and nproc, so it cannot run on macOS
`test/check.sh` uses two GNU-only constructs, so it cannot run on a BSD userland such as macOS. Found while looking at #958, but they are independent of it.

**`sed -i`.** GNU sed takes `-i` with no argument; BSD sed requires one, and treats the next word as the suffix. So

```
sed -i "s|port = 5432|port = $PORT|g" pgagroal.conf
```

on macOS reads the expression as the backup suffix and then tries to use the filename as the script:

```
sed: 1: "pgagroal.conf": unterminated substitute pattern
```

The two spellings cannot be shared, since `sed -i ''` is an error on GNU. Two call sites, both in `run_config_test()`.

**`nproc`.** Part of coreutils, which macOS does not ship, so `make -j$(nproc)` expands to `make -j` and the build either serialises or errors depending on the make.

Neither is reachable on Linux today, so this changes nothing there. I have a patch adding a `sed_in_place` helper that picks the spelling from `sed --version`, and a `cpu_count` helper falling back to `sysctl -n hw.ncpu`.

Worth noting a third for anyone scripting around this: `timeout` is coreutils too and is likewise absent on macOS.

Pranav-error:
Fixes #1060.

`test/check.sh` uses two GNU-only constructs, so it cannot run on a BSD userland such as macOS.

**`sed -i`** — GNU takes `-i` with no argument, BSD requires one and reads the next word as the backup suffix:

```
sed: 1: "pgagroal.conf": unterminated substitute pattern
```

`sed -i ''` is in turn an error on GNU, so the two spellings cannot be shared. Added `sed_in_place`, which picks based on `sed --version`. Two call sites, both in `run_config_test()`.

**`nproc`** — coreutils, absent on macOS, so `make -j$(nproc)` expanded to `make -j`. Added `cpu_count`, falling back to `sysctl -n hw.ncpu`.

Neither fallback is reachable on Linux, so CI behaviour there is unchanged.

## Verification

Exercised both helpers on macOS: `sed_in_place` rewrites in place and leaves no `.bak` behind (which is what confirms the BSD form is right rather than silently writing a backup), and `cpu_count` returns 10 here.

Found while working on #958. This does not make `check.sh` run on macOS by itself — `start_postgresql_local()` still assumes a `postgres` system user and passwordless sudo, which I have written up on that issue. These two are independent of that and worth having either way.


Userfrom1995:
@Pranav-error I think it would be better to include this change in the macOS PR that resolves #958. I don't see a point in merging it separately under a new issue, since it doesn't really solve anything on its own and the macOS PR will contain this change as well.

Let's leave it here for now. Once we have the macOS CI PR, we can close this as superseded.

Pranav-error:
Makes sense — I'll fold these into the #958 PR when I get to the macOS setup path, and this one can be closed as superseded then.

